PR #6, opened 2015-07-02
fix typo

callahanchris:


leebyron:
Crowdsourced spellcheck is wildly inefficient, but it sure is fun! Thanks!

PR #19, opened 2015-07-03
Update Section 2 -- Consistent use of comma

clentfort:
Some queries are missing a `,` to separate fields in a query.


leebyron:
Thanks

This was intentional as commas are optional in GraphQL (they're whitespace), but I agree that consistency is important.

PR #30, opened 2015-07-06
Fix a validation example

crpdm:
I think this example is valid because `nickname` is defined in `Pet` interface too. So I changed it to `barkVolume` which is the field of a concrete implementor `Dog`.

### DIFF
--- a/Section 5 -- Validation.md
+++ b/Section 5 -- Validation.md
@@ -31,7 +31,7 @@ type Dog : Pet {
 }
 
 interface Sentient { name: String! }
-interface Pet { name: String!, nickname: String }
+interface Pet { name: String! }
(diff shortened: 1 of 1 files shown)


dschafer:
Good find!

Based on https://github.com/graphql/graphql-js/blob/master/src/validator/__tests__/harness.js#L38-43, I think we should probably remove `nickname` from `Pet`, which will make this example invalid again, and demonstrate the case it was trying to establish (even if all of the implementations have a field, it is still not valid to query it if the interface doesn't).

crpdm:
I didn't know about tests in graphqh-js. I replace my commit.

dschafer:
Thanks!

PR #31, opened 2015-07-06
Change possibleTypes wording for Interface

ckimes89:
Most likely a copy/paste error from possibleTypes in Union right above.


dschafer:
Thanks!

PR #32, opened 2015-07-06
Directives improvements

leebyron:
Here is spec language mirroring the changes in https://github.com/graphql/graphql-js/pull/31.

Changed:

* Directive grammar changed to `@ Name Arguments?`
* "Field Arguments" changed to "Arguments" almost everywhere to be more generic.
* Argument validation rules made more generic to handle directives cases.
* Directive validation simplified.
* Directive explaination in Language now matches current state.
* @include and @skip documented in Type System

### DIFF
--- a/Section 5 -- Validation.md
+++ b/Section 5 -- Validation.md
@@ -119,8 +119,8 @@ fragment definedOnImplementorsQueriedOnUnion on CatOrDog {
   * Let {setForKey} be the set of selections with a given response key in {set}
   * All members of {setForKey} must:
     * Have identical target fields
-    * Have identical sets of arguments name-value pairs.
-    * Have identical sets of directive name-value pairs.
+    * Have identical sets of arguments.
+    * Have identical sets of directives.
 
 ** Explanatory Text **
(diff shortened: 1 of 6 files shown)


Section 5 -- Validation.md
```diff
@@ -314,18 +310,17 @@ fragment argOnRequiredArg on Dog {
 }
 
 fragment argOnOptional on Dog {
-  isHousetrained(atOtherHomes: true)
+  isHousetrained(atOtherHomes: true) @include(if: true)
 }
 ```
 
-and the following is invalid since command is not defined on DogCommand:
+and the following is invalid since `command` is not defined on `DogCommand`,
```
dschafer: I would split this out into a separate example, since it shows that two things are invalid:

```
fragment invalidFragmentArgName {
  doesKnowCommand(dogCommand: SIT) @include(unless: false)
}
```

and

```
fragment invalidFieldArgName {
}
```

leebyron:
Thanks Dan, I broke that out into two examples

PR #35, opened 2015-07-07
Fix a few typos

pushrax:
Thanks for your work so far! Looking forward to the final spec.

### DIFF
--- a/Section 5 -- Validation.md
+++ b/Section 5 -- Validation.md
@@ -425,7 +425,7 @@ fragment goodNonNullArg on Arguments {
 
 ```
 
-On a field with a a nullable arg, that argument can be omitted.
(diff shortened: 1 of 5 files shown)


Section 5 -- Validation.md
```diff
@@ -425,7 +425,7 @@ fragment goodNonNullArg on Arguments {
 
 ```
 
-On a field with a a nullable arg, that argument can be omitted.
+A field with a nullable argument can be omitted.
```
dschafer: This seems to change the meaning, suggesting that the field can be omitted (rather than that the argument can be omitted)?

pushrax: You're right. I'll change it back to the original meaning.

dschafer:
Thanks!

PR #38, opened 2015-07-08
Correct a few typos and grammatical issues.

simenbrekken:
* client -> clients
* with -> width
* missing "a"
* "should returns errors" -> "should return errors"


leebyron:
Thanks!

PR #45, opened 2015-07-13
Add prev/next links in documentation

jesstelford:
Makes navigating the documentation on Github a little easier :)


leebyron:
These markdown files are input into our spec generator which produces facebook.github.io/graphql. The changes here aren't compatible with that tool which is why the build has failed. 

The markdown files are written to be easy to read on their own and to render reasonably in github (as opposed to writing directly in HTML or LaTeX) but aren't the final deliverable.

jesstelford:
Oh, in that case, it might be worthwhile linking to that page from the top of the README - I didn't see the link in the description at the top of the page until you mentioned it here. Thanks :)

leebyron:
Great suggestion!

PR #49, opened 2015-07-14
Describe how non-null and list types are written earlier.

gabrielf:
The `Type!` syntax for the non-null wrapping type is used in section 4.1.4 - Type Name Introspection, but it is not mentioned anywhere before that which I found confusing (probably because I'm not used to develop in a language that has this construct). I've added it to the section where Non-null is defined and added a similar explanation for the List wrapping type for symmetry.


leebyron:
Thanks for this improvement

PR #57, opened 2015-07-16
Update README with clarified hero root field

dschafer:
Updates the README with a more clear `hero` field, that takes an argument. This matches the changes in https://github.com/graphql/graphql-js/pull/70.


leebyron:
:+1:

PR #58, opened 2015-07-17
make __type name arg a non-null

martinandert:
According to the JS implementation, this is a `NonNull` argument (which makes more sense).


leebyron:
Thanks for the catch

PR #60, opened 2015-07-19
Clarify operation name

tristanz:
Jumping around the spec and README, I was confused by the meaning and purpose of operation names.  I tried to clarify things as I understand them and make things more precise.


leebyron:
These are great and well written suggestions, thank you!

PR #61, opened 2015-07-20
Add section on comments

tristanz:
Comments are not defined except deep inside the Ignore Source section.  I added them before they are used in examples.

### DIFF
--- a/Section 2 -- Language.md
+++ b/Section 2 -- Language.md
@@ -34,6 +34,11 @@ If a document contains only one query operation, that operation may be
 represented in the shorthand form, which omits the query keyword and
 query name.
 
+## Comments
+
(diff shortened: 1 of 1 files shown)


clentfort:
Now line terminator is not defined... How about keeping it simpler? 
> GraphQL supports single line comments: On a line, anything after a `#` is considered a comment and will be ignored.

leebyron:
I'll expand on this in the spec, thanks for the suggestion and this initial wording.

leebyron:
Merged in 140abf378bc03c63e92a320d5d1b8b119ff75510 with some amendments.

tristanz:
Much better.  :+1:  Take my other pull requests as suggestions too.  I basically just tried to fix things I was searching for definitions on.

PR #62, opened 2015-07-20
clarify meaning of selection sets

### LINKED ISSUE #41: Selection set formal definition
It would be great to have some formal definition of selection set. It is introduced in the [fields section](http://facebook.github.io/graphql/#sec-Language.Fields) without a peep.

tristanz:
Selection sets are used before being clearly defined and fields generally described only in the context of global fields, which is mildly confusing.

I attempted to fix both issues to make fields and selection sets easier to understand when reading the spec from top to bottom.

This fixes #41.

### DIFF
--- a/Section 2 -- Language.md
+++ b/Section 2 -- Language.md
@@ -52,31 +52,35 @@ Note: many examples below will use the query shorthand syntax.
 
 ## Fields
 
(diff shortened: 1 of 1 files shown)


leebyron:
Thanks for these suggestions!

PR #68, opened 2015-07-28
Move source files into spec

dschafer:
Moves the source files into their own directory, to clarify to someone just looking at the top level directory that they are source files and not designed to be read.


leebyron:
:+1:

PR #75, opened 2015-08-10
Improve Grammar Summary Readability and Consistency

enaqx:


leebyron:
I would prefer not to do this. While it helps clarity when viewed in Github, it makes the text files themselves a bit harder to read.

enaqx:
Ok, got it. :dog:

leebyron:
Thanks for the exploration though!

PR #78, opened 2015-08-12
Clarify execution behavior of abstract types in Spec.

leebyron:
`CompleteValue` is missing reference to abstract type resolution.

Thanks to @seanchas for the careful read and highlighting this issue.


dschafer:
Looks good.

PR #83, opened 2015-08-19
[RFC] Null value

leebyron:
This proposal adds a null literal to the GraphQL language and allows it to be provided to nullable typed arguments and input object fields.

This presents an opportunity for a GraphQL service to interpret the explicitly provided null differently from the implicitly not-provided value, which may be especially useful when performing mutations using a per-field set API.

For example, this query may represent the removal/clearing of the "bar" field from thing with id: 4.

```
mutation editThing {
  editThing(id: 4, edits: { foo: "added", bar: null }) {
    # ...
  }
}
```

In addition to allowing `null` as a literal value, this also proposes interpretting the variables JSON to distinguish between explicit null and implicit not provided:

```
mutation editThing($edits: EditObj) {
  editThing(id: 4, edits: $edits) {
    # ...
  }
}
```

This variables results in the unsetting of `bar`

```
{ "edits": { "foo": "added", "bar": null } }
```

Finally, this proposes following the not-provided-ness of variables to their positions in arguments and input-obj fields

```
mutation editThing($editBaz: String) {
  editThing(id: 4, edits: { foo: "added", bar: null, baz: $editBaz }) {
    # ...
  }
}
```

Such that the three variables are semantically different:

 * `{}` The "baz" input field is "not provided"
 * `{"editBaz": null}` The "baz" input field is `null`
 * `{"editBaz": "added"}` The "baz" input field is `"added"`

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -115,6 +115,9 @@ and floating-point values, they are interpreted as an integer input value if
 they have an empty fractional part (ex. `1.0`) and otherwise as floating-point
 input value.
 
+For all types below, with the exception of Non-Null, if the explicit value
+{null} is provided, then then the result of input coercion is {null}.
+
 **Built-in Scalars**
 
 GraphQL provides a basic set of well-defined Scalar types. A GraphQL server
@@ -809,15 +812,20 @@ input ExampleInputObject {
 
 Original Value          | Variables       | Coerced Value
 -------------------------------------------------------------------------------
-`{ a: "abc", b: 123 }`  | `{}`            | `{ a: "abc", b: 123 }`
-`{ a: 123, b: "123" }`  | `{}`            | `{ a: "123", b: 123 }`
-`{ a: "abc" }`          | `{}`            | Error: Missing required field {b}
+`{ a: "abc", b: 123 }`  |                 | `{ a: "abc", b: 123 }`
+`{ a: 123, b: "123" }`  |                 | `{ a: "123", b: 123 }`
+`{ a: "abc" }`          |                 | Error: Missing required field {b}
+`{ a: "abc", b: null }` |                 | Error: {b} must be non-null.
+`{ a: null, b: 1 }`     |                 | `{ a: null, b: 1 }`
 `{ b: $var }`           | `{ var: 123 }`  | `{ b: 123 }`
+`{ b: $var }`           | `{}`            | Error: Missing required field {b}.
 `{ b: $var }`           | `{ var: null }` | Error: {b} must be non-null.
-`{ b: $var }`           | `{}`            | Error: {b} must be non-null.
-`{ b: $var }`           | `{}`            | Error: {b} must be non-null.
 `{ a: $var, b: 1 }`     | `{ var: null }` | `{ a: null, b: 1 }`
 `{ a: $var, b: 1 }`     | `{}`            | `{ b: 1 }`
+
+Note: there is a semantic difference between the input value
+explicitly declaring an input field's value as the value {null} vs having not
+declared the input field at all.
 
 #### Input Object type validation
 
@@ -867,9 +875,20 @@ By default, all types in GraphQL are nullable; the {null} value is a valid
 response for all of the above types. To declare a type that disallows null,
 the GraphQL Non-Null type can be used. This type wraps an underlying type,
 and this type acts identically to that wrapped type, with the exception
-that `null` is not a valid response for the wrapping type. A trailing
+that {null} is not a valid response for the wrapping type. A trailing
 exclamation mark is used to denote a field that uses a Non-Null type like this:
 `name: String!`.
+
+**Nullable vs. Optional**
+
+Fields are *always* optional within the context of a query, a field may be
+omitted and the query is still valid. However fields that return Non-Null types
+will never return the value {null} if queried.
+
+Inputs (such as field arguments), are always optional by default. However a
+non-null input type is required. In addition to not accepting the value {null},
+it also does not accept omission. For the sake of simplicity nullable types
+are always optional and non-null types are always required.
 
 **Result Coercion**
 
@@ -881,42 +900,40 @@ must be raised.
 
 **Input Coercion**
 
-If the argument of a Non-Null type is not provided, a query error must
-be raised.
+If an argument or input-object field of a Non-Null type is not provided, is
+provided with the literal value {null}, or is provided with a variable that was
+either not provided a value at runtime, or was provided the value {null}, then
+a query error must be raised.
 
-If an argument of a Non-Null type is provided with a literal value, it is
-coerced using the input coercion for the wrapped type.
+If the value provided to the Non-Null type is provided with a literal value
+other than {null}, or a Non-Null variable value, it is coerced using the input coercion for the wrapped type.
 
-If the argument of a Non-Null is provided with a variable, a query error must be
-raised if the runtime provided value is not provided or is {null} in the
-provided representation (usually JSON). Otherwise, the coerced value is the
-result of using the input coercion for the wrapped type.
-
-Note that `null` is not a value in GraphQL, so a query cannot look like:
+Example: A non-null argument cannot be omitted.
 
 ```!graphql
 {
-  field(arg: null)
+  fieldWithNonNullArg
 }
 ```
 
-to indicate that the argument is {null}. Instead, an argument would be {null}
-only if it is omitted:
+Example: The value {null} cannot be provided to a non-null argument.
 
-```graphql
+```!graphql
 {
-  field
+  fieldWithNonNullArg(nonNullArg: null)
 }
 ```
 
-Or if passed a variable of a nullable type that at runtime was not provided
-a value:
+Example: A variable of a nullable type cannot be provided to a non-null argument.
 
 ```graphql
 query withNullableVariable($var: String) {
-  field(arg: $var)
+  fieldWithNonNullArg(nonNullArg: $var)
 }
 ```
+
+Note: The Validation section defines providing a nullable variable type to
+a non-null input type as invalid.
 
 **Non-Null type validation**
 
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -127,11 +127,14 @@ Value[Const] :
   - FloatValue
   - StringValue
   - BooleanValue
+  - NullValue
   - EnumValue
   - ListValue[?Const]
   - ObjectValue[?Const]
 
 BooleanValue : one of `true` `false`
+
+NullValue : `null`
 
 EnumValue : Name but not `true`, `false` or `null`
 
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -634,6 +634,7 @@ Value[Const] :
   - FloatValue
   - StringValue
   - BooleanValue
+  - NullValue
   - EnumValue
   - ListValue[?Const]
   - ObjectValue[?Const]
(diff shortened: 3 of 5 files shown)


leebyron:
cc @dschafer @schrockn

dschafer:
This looks good to me!

dylanahsmith:
:+1: this will definitely simplify the transition from REST to GraphQL for mutations, otherwise, a boolean field would need to be provided to distinguish between setting a field to be `null` and not changing the field.

oguzbilgic:
:+1: 
Seems like it has been some time. Is this still on roadmap?

urbanappetite:
I created a fork a short while ago that allows for this. I've not merged in the most recent commits yet but it's working in production.

https://www.npmjs.com/package/graphql-nullable

OlegIlyenko:
:+1: Looks like this PR was open for quite a while now. Just wanted to give my option on this change.

I'm coming from scala background. I don't see an addition of `null` as a problem. It opens up quite a few interesting possibilities. In general statically-typed languages still make it possible to distinguish between `undefined` (property is not defined on object) and `null` (for example all scala JSON libraries have `JNull` or `JsNull` AST nodes). I tried to implement this proposal in the [sangria](https://github.com/sangria-graphql/sangria) and it looks good so far. If one does not care about distinction between `null` and `undefined` then they can be just treated the same (it happens pretty often). But on the other hand, if you need to distinguish between them, then this addition is very helpful.

> Note: there is a potential semantic difference between the input value
> explicitly declaring an input field's value as the value `null` vs having not
> declared the input field at all. GraphQL services may optionally interpret these
> two cases differently if it is advantageous.

I like this part and would keep it like this. So that it's up to the server implementation to decide whether to treat them differently or not. This would also mean that standard protocols (like introspection API or relay connection) can't relay on this mechanism, if I understand it correctly.

joewood:
Any reason why this RFC is restricted to mutations? Could it be applied to explicit input field arguments? E.g. a query being interpreted to a SQL clause `where table.field is NULL`.

dylanahsmith:
It doesn't look like this RFC is specific to mutations.  Mutations are just the prominent example.  The only change to the spec that mentions mutations is this added line:

> For example, to represent deleting a field vs not altering a field during a mutation, respectively.

dbackeus:
LGTM :+1: 

Coming from a ruby background not being able to send `null` to set an attribute to a `nil` value is certainly surprising.

taion:
I'd be very happy to see this change land. We have a number of places where our GraphQL layer bridges to REST services that have special handling for explicit nulls in e.g. `PATCH` operations, and it would be much easier to bridge to those services if we have the option of using a `null` literal.

akre54:
We often have NULLs in our database represent default values (i.e. a localizer with a NULL locale is the default translation). We've had to hack around this, but would be great for GraphQL to support natively.

skosch:
Not to be a bother, but what's the status on this?  Are there still open questions?

leebyron:
There are still open questions on how to resolve "nullability" with "optionality" once a value can be explicitly provided and be the value `null`. That also has implications on whether the value `null` should be replaced with a default value or not.

This is still open, but it's been lower priority relative to some of the other proposals we've been working the kinks out of.

skosch:
Thanks for the quick response @leebyron, that's good to hear.

Default values mean well, but my current practical problem, unless I'm doing it wrong, is that database updates happen based on changesets, and so right now it's impossible to know whether the client wanted a field nulled or simply left unchanged. (I'm probably beating one rotten carcass of a horse here, but since this is marked RFC I wanted to throw another vote in the hat for bona fide nulls.)

taion:
BTW, I think in most other contexts that I've seen, "value is required" (may not be `undefined`) and "`null` is permitted" are treated as orthogonal concepts.

One place this comes up the most strongly is for e.g. a `PATCH` update (for a partial update of an object). In such a case, certain non-nullable fields may still not be required (i.e. to have their values not be changed).

leebyron:
Yeah, to be clear here the question isn't "is it a good idea to introduce a `null` value" the question is "is it worth the additional introduced complexity"

Splitting the concepts of optionality and nullability actually adds a lot of cascading changes to GraphQL, so while in isolation it's perfectly easy to make a case for why having them separate is fine, what we actually want to be doing here is striving to maintain GraphQL's simplicity whenever possible.

The costs of maintaining simplicity is that there are often corner-cases where GraphQL feels awkward, but the value is that the primary cases feel simple. I understand that may feel like a slight - it certainly doesn't feel like a corner-case when you're the one experiencing it, but it is the reality that most mutations in GraphQL use the action pattern rather than the CRUD pattern, so "setting a field to null" is actually a case we haven't had to deal with in production yet.

And also to be clear, I do think having a `null` literal is a good idea - I do want to make enabling CRUD style operations more reasonable, I would just like to do so with as little collateral damage to simplicity as possible.

skosch:
That was thoughtfully written. Thanks. :)

Not sure if it's been discussed elsewhere – could you sidestep those issues if you defined some `Relay.NULL` constant for us to use, instead of messing with the JS `null` literal?

taion:
I don't think `Relay.NULL` can be the answer, since input types can't be union types.

The CRUD stuff just comes up when migrating existing systems to GraphQL, I guess.

skosch:
Ah, forgot about that. Never mind then.

danielkwinsor:
I might be the first person to say that I'd like to see more work on this RFC before it is part of the spec.  Don't get me wrong, I do want to see implicit / explicit nulls.

2 things:

It is my opinion that there ought not to be multiple interpretations of what happens for implicit or explicit null, depending on which GraphQL service is being used.  But that's what 2 different sections say: 
>A GraphQL service may interpret the two forms differently, if it is advantageous.
____
>Note: there is a potential semantic difference between the input value explicitly declaring an input field's value as the value `null` vs having not declared the input field at all. GraphQL services may optionally interpret these two cases differently if it is advantageous.

In practical terms, there's a high chance of my company going from graphql-java to graphql-go, and I don't want to get bitten.  It's pretty easy to let seemingly minor things make major impact down the line, like changing the "ordering" of a contractually unordered collection (I refer to [Java 8 HashMaps](https://examples.javacodegeeks.com/core-java/util/hashmap/hashmap-changes-in-java-8) and the 75% of PRD issues attributable to this, even though I love how Java 8 HashMap is now).
In theoretical terms, the first comment says there are 3 distinct semantical interpretations of variables in a query, so why should the semantics of the GraphQL service be a black box to the querier?




Besides that, I'd also like some more clarity on null + lists.
Consider this, with the value after coercion being left as an exercise for the reader.

Spec | imp/exp Null | Input Type | Coerced Value
------------ | ------------- | ------------ | -------------
Current | N/A | String | null
Current | N/A | [String] | ?
Current | N/A | [String!] | ?
Current | N/A | [[[[String]]]] | ?
Proposed | Implicit | String | null
Proposed | Implicit| [String] | ?
Proposed | Implicit | [String!] | ?
Proposed | Implicit | [[[[String]]]] | ?
Proposed | Explicit | String | null
Proposed | Explicit | [String] | ?
Proposed | Explicit | [String!] | ?
Proposed | Explicit | [[[[String]]]] | ?

Take the case of [[String] from graphql-java](https://github.com/graphql-java/graphql-java/issues/74).
An implicit null was input to [String], yielding the (unexpected) value of a list object with 1 null element.
So in the case of `null` input coercion, we might do something different for implicit or explicit.
My first thoughts are to word things in such as way that explicit nulls are always coerced into the input [thus potentially yielding list of (list of (null))], while implicit nulls are never coerced but substituted directly for the `null` value [thus yielding null with the chance of NPE].


Last and least, typo on spec/Section 3 -- Type System.md line 117, it has 2 "then"s

freiksenet:
Is there any update on this branch? Our CRUD API is a bit inconvenient, because you can't set values to null without using 'replace'. I am considering adding some other way of nullifying fields, but if null literal is about to land, I'd rather wait for it to happen.

I can help landing this in graphql-js, if some work there is needed.

Freezystem:
If I may put my two cents in, based on the SQL design we have to distinguish creation mutation from update mutation. 

1. On `INSERT` when you don't specify a column value it's considered as `NULL`.
2. On `UPDATE` when you don't specify a column value it's considered unchanged and you have to explicitly set the column to `NULL` if you want to reset the corresponding value.

IMO GraphQL have to keep this policy. Indeed when an update action is done on a table row, user isn't always aware of the whole table schema despite of introspection and therefor should be able to omit column. 

There are two ways to reset a column in that situation:
- Allow the use of `null` or any dedicated keyword in the request as a column value: 
```
mutation { 
  editThing(foo : "stuff", bar : null) { ... }
}
```
- Explicitly define a keyword attribute to reset columns:
```
mutation { 
  editThing(change : { foo : "stuff" }, reset : { bar }) { ... }
}
```

First solution seems more standard and is pretty straightforward to every readers. 
Second is also a valuable design but can lead to a lot of different implementation. If preferred it may be a good thing to define common keywords in the spec and stick to it to keep code understandable to everyone at first sight.

Second solution is for now the only one to really work in production, but I personally find that the code written is more complex to maintain compared to the first one.

thorstadt:
I certainly understand the argument around complexity that missing vs. specified null semantics introduces. I think a counter-argument could also be made that enabling the CRUD model and being able to push `null` values is simpler for some scenarios since it presents a more direct migration path from REST-style web services.

We've actually added support to our own GraphQL library for a `null` literal since supplying separate mutations for clearing values was not a good option in our case.

mcheshkov:
I'd like to add to @danielkwinsor table more kind of types: `[[Int]!]`, `[[Int!]]` etc.
How should input coercion work in this cases?
`[[Int]!]` can be `null`or `[[null]]`.
`[[Int!]]` can be `null`, `[null]` or invalid.
`[[[Int!]]]` can be `null`, `[null]`, `[[null]]` or invalid.
`[[[Int]]!]` can be `null`, `[[null]]` or `[[[null]]]`.

IMO it's very surprising that explicit `null` will be proper "no list", but implicit "no value" will be list(s) of single `null`, or even invalid input.
So I think it's a bad idea to coerce both explicit and implicit null to list. Same goes with nested nulls: `[null]` won't silently coerce to `[[null]]`, for `[[Int]]` it will stay `[null]`, and for `[[Int]!]` it will get rejected as invalid value.

leebyron:
Thanks for the discussion! I'm still interested in furthering this and really happy to have some additional concerns tossed in - that will help define clear behavior in the spec.

I think the next step is prototyping this so the coercion rules can be toyed with until we have something that is both easy to explain (and predict) but are also valuable to solve the problems at hand

theorygeek:
Per @danielkwinsor's comment, I do think it'd be good to eliminate the odd behavior that lists have when `null` is the input type (even though this is probably tangential).

Currently, [per the spec](http://facebook.github.io/graphql/#sec-Lists):
>If the value passed as an input to a list type is not as list, it should be coerced as though the input was a list of size one, where the value passed is the only item in the list. This is to allow inputs that accept a “var args” to declare their input type as a list; if only one argument is passed (a common case), the client can just pass that value rather than constructing the list.

So if the input type is `[String!]`, and the provided value is `null`, it seems that this would require that input to be interpreted as `[null]`, and then subsequently reject it. I think this behavior would surprise most people, and in fact I don't even think most implementations behave like this. As a side-effect, it also means that `[String!]` is not really distinct from `[String!]!`.

Perhaps we could add to the spec some language like this:
>If the value passed as an input to a list type is not a list **and is not `null`**, it should be coerced as though the input was a list of size one...

leebyron:
>  I think this behavior would surprise most people, and in fact I don't even think most implementations behave like this. 

Thanks for this feedback - I agree that it's easy to misinterpret the spec in this way - the way implementations behave is spec compliant so your suggestion is really just to clarify how things are supposed to work rather than being a change. I'll make it happen.

langpavel:
`null` as explicitly passed is not ambiguous. Coercion is. So if list may be coerced from single value,I expect error instead. 
We are solving there same problem that been solved in SQL... 
Required list of required value is clear error for null
Required list should be coerced to empty list. Backend then handle it as desired..
Not required list should be treated as null.

jbinto:
Is the change in https://github.com/facebook/graphql/commit/98187c86963708b524380ffe6f0fd22959c26faa (i.e.   clarifying the spec re: the non-nullable input list) just a proposal? Can we expect to see this in `graphql-js` any time soon?

> [String!] is not really distinct from [String!]!

This is the problem I'm hitting. I would like to have lists as optional input arguments (`[String!]`), but now clients must provide an empty list, or I must remove the non-null requirement on the Strings themselves.

langpavel:
[String!] should be coerced to null,
[String!]! should be coerced to empty list

leebyron:
@jbinto - graphql-js should be compliant and if not is a bug. Please feel free to file a bug there or open a PR with a failing test case if you've found one!

leebyron:
Rebased this just now. For reviewers, here's what this looks like https://ns-ykdwgdukwo.now.sh/

leebyron:
Just got a careful review from @dschafer offline - so merging this now.

freiksenet:
Party today, null is in :)

spec/Section 3 -- Type System.md
```diff
@@ -115,6 +115,9 @@ and floating-point values, they are interpreted as an integer input value if
 they have an empty fractional part (ex. `1.0`) and otherwise as floating-point
 input value.
 
+For all types below, with the exception of Non-Null, if the explicit value
+{null} is provided, then then the result of input coercion is {null}.
```
taion: "then then"

langpavel:
Work in progress in https://github.com/graphql/graphql-js/pull/544

spec/Section 3 -- Type System.md
leebyron: Thanks for the careful read, fixed in 3ce8b790da33f52f9258929258877a0a7768c620

PR #87, opened 2015-08-25
Fix typo

ning-github:
corrected 'out' to 'our', since we have been referring to it as 'our type system' like in line 833.


dschafer:
Thanks!

PR #89, opened 2015-08-25
Update Section 1 -- Overview.md

lyip1992:
I'm not sure if the semicolons were deliberate. But the way I understand semicolons is that they are used to separate complete sentences. Let me know your thoughts on this. Thanks.


leebyron:
Thanks

PR #90, opened 2015-09-02
[RFC] GraphQL Schema Definition Language (SDL)

leebyron:
This adds the type definition syntax to the GraphQL specification.

There are some remaining issues to solve:
- [x] Include reference in Execution chapter about non-executable GraphQL
- [x] Description is not yet represented
- [x] Deprecation is not yet represented
- [x] Directives are not yet represented
- [x] Top level Schema is not yet represented
- [x] Investigate impact on validation rules
- [x] Decide on _any_ or _many_ rules for type fields / enum values
- [x] Combine into type system chapter
- [x] Consider separating into two top level grammars
- [x] Solve ambiguity for extending interfaces (https://github.com/graphql/graphql-js/pull/1166)


leebyron:
Proposal for schema:

```
schema {
  query: QueryType
  mutation: MutationType
}
```

```
SchemaDefinition : { OperationTypeDefinition+ }
OperationTypeDefinition : OperationType : NamedType
```

spec/Appendix B -- Grammar Summary.md
```diff
@@ -171,3 +172,38 @@ NonNullType :
...
+
+FieldDefinition : Name ArgumentsDefinition? : Type
+
+ArgumentsDefinition : ( InputValueDefinition+ )
+
+InputValueDefinition : Name : Type DefaultValue?
+
+InterfaceTypeDefinition : interface Name { FieldDefinition+ }
+
+UnionTypeDefinition : union Name = UnionMembers
```
dittos: syntax for `UnionMembers` is missing

leebyron: Good catch, fixed!

spec/Appendix B -- Grammar Summary.md
```diff
@@ -171,3 +172,42 @@ NonNullType :
...
+  - ScalarTypeDefinition
+  - EnumTypeDefinition
+  - InputObjectTypeDefinition
+  - TypeExtensionDefinition
+
+ObjectTypeDefinition : type Name ImplementsInterfaces? { FieldDefinition+ }
+
+ImplementsInterfaces : implements NamedType+
+
+FieldDefinition : Name ArgumentsDefinition? : Type
```
nuance: should this be NamedType?

nuance:
Should this also include directives? Something like:

```
DirectiveDefinition : directive Name ArgumentsDefinition?
```

This doesn't capture scope / validation rules, so perhaps it should include an 'on' clause?

```
DirectiveDefinition : directive Name ArgumentsDefinition? DirectiveScopes?
DirectiveScopes : on DirectiveScope+
DirectiveScope : field | fragment | operation
```

So include would be defined like:

```
directive include(if: Boolean!) on field, fragment
```

joshprice:
Is there any movement on this? This is a pretty compelling piece of the GraphQL puzzle and opens up the possibility of tooling around schema design and also allows the creation of a cross language test suite.

nuance:
I ended up implementing most of this grammar in a project I'm working on. Here's the relevant lines from a PEG grammar (just the type parsing stuff):

```
TypeDefinition <- ObjectTypeDefinition / InterfaceTypeDefinition / UnionTypeDefinition / ScalarTypeDefinition / EnumTypeDefinition / InputObjectTypeDefinition / TypeExtensionDefinition / DirectiveDefinition
ObjectTypeDefinition <- "type" _ Name _ ImplementsInterfaces? _ "{" __ (FieldDefinition __)* "}"
ImplementsInterfaces <- "implements" _ (Name _)+
FieldDefinition <- Name _ ArgumentsDefinition? _ ':' _ InputType
ArgumentsDefinition <- '(' _ ( InputValueDefinition _ )* ')'
InputValueDefinition <- Name _ ':' _ InputType _ DefaultValue?
InterfaceTypeDefinition <- "interface" _ Name _ '{' __ (FieldDefinition __)+ '}'
UnionTypeDefinition <- ("union" / "Union") _ Name _ "=" _ UnionMembers
UnionMembers <- NamedType _ ("|" _ NamedType _)*
ScalarTypeDefinition <- "scalar" _ Name
EnumTypeDefinition <- "enum" _ Name _ "{" __ (EnumValue __)+ __ "}"
InputTypeDefinition <- Name _ ':' _ InputType _ DefaultValue?
InputObjectTypeDefinition <- "input" _ Name _ "{" __ (InputTypeDefinition __)+ "}"
DirectiveDefinition <- "directive" _ Name _ "on" _ (Name _)+ ArgumentsDefinition?
TypeExtensionDefinition <- "extend" _ ObjectTypeDefinition
```

I *think* this is basically compatible with the language defined in this PR, with a few additions (directives).

Answering the questions @leebyron asked:
- description and deprecation are both expressed via directives
- not sure what top level schema means - we define a query root object like:

```
type QueryRoot{}
```

then extend it as needed (I'm assuming that's what you're referencing?)

```
extend type QueryRoot {
    __schema : __Schema!
    __type(name: String!) : __Type
}
```

- It's not clear what validation would be required. We're doing a little to catch compliance with the relay extensions and handle some of our own extensions, but nothing that passes the grammar has brought up anything interesting.

joshprice:
- Got a concrete example for implementing metadata (`description`, `isDeprecated`, etc) using directives?
- Top level schema I think refers to being able to parse the schema itself which is fairly trivial, just needs to be spec'd fully, but it looks like your example could use existing primitives
```graphql
schema {
  query: QueryType
  mutation: MutationType
}
```
- You can't mix type defs and queries which is described in the spec already, but presumably not in the execution section yet
- Schema validation would be somewhat different to query validation, but as you point out if it is parsable then it's probably right, not sure what semantics could be validated? Guidance on orphan/unreferenced types, circular references, for a start

joshprice:
For reference there are some concrete examples here: https://github.com/matthewmueller/graph.ql

Descriptions are assumed to be in comments immediately preceding the thing it's describing, assuming deprecation etc could be handled in a similar fashion with a specially formatted comment.

Another possibility for metadata might be type annotations starting with `@`:

```graphql
@description Describes a person
type Person {
  @isDeprecated true
  field: Int
}
```

joshprice:
Thinking more about how this is done with directives, you could make something quite readable:

```graphql
@metadata(description: "Describes a person")
type Person {
  @metadata(isDeprecated true)
  field: Int
}
```

joshprice:
Enums are also an issue here because `value`s are not able to be supplied using the current syntax

leebyron:
Updated to rebase atop recent changes and includes syntax for defining directives based on https://github.com/graphql/graphql-js/pull/318

leebyron:
Updated to rebase, fixed some outdated language and added `SchemaDefinition`

spec/Section 2 -- Language.md
```diff
@@ -938,3 +943,239 @@ and operations.
...
+In this example, a Scalar type called `DateTime` is defined:
+
+```graphql
+scalar DateTime
+```
+
+
+#### Object
+
+ObjectTypeDefinition : type Name ImplementsInterfaces? { FieldDefinition+ }
```
OlegIlyenko: @leebyron I think I found a mismatch with the reference implementation. It uses `any` to parse the list of fields, but according to spec it should use `many` (one or more field definitions):

https://github.com/graphql/graphql-js/blob/master/src/language/parser.js#L765

Should `ObjectTypeDefinition` be defined like this?

```
ObjectTypeDefinition : type Name ImplementsInterfaces? { FieldDefinition* }
```

(it's the same with the `InputObjectTypeDefinition`)

leebyron: Thanks for the catch and close read. I'll take a closer look and figure out which is more appropriate, but I think you're probably right.

charlieschwabacher:
@leebyron can you share anything about your thinking on a syntax for descriptions?  Do you expect it to use directives? comments?  Or is it still a ways off?

helfer:
@charlieschwabacher Lee has said somewhere that he wants it to be in comment format, because that will be the easiest to read and format. I can't find the issue(s) where he said that any more, but I'm 100% sure about it.

leebyron:
Yep, I think we should use comments as doc blocks to describe types and fields. It's a little annoying and "un-pure" from the parser's point of view since technically comments are "ignored" tokens, however it's hard to ignore the developer ergonomics and it's also quite easy to point to prior art like JavaDoc.

leebyron:
I'll update this RFC soon to reflect progress made in the experimental implementation in graphql-js. Deprecation is now supported in that implementation via a `@deprecated` directive.

charlieschwabacher:
@helfer @leebyron - I'm happy to hear that, and agree about comments being the most ergonomic.  thanks for the update!

OlegIlyenko:
FYI, as a proof of concept, I already implemented comment-based descriptions in sangria. If you would like to experiment with them, you can check it out here: http://toolbox.sangria-graphql.org/proxy (all `##` comments are treated as descriptions and all changes to the schema definition are reflected in GraphiQL and documentation panel)

exogen:
> it's hard to ignore the better developer ergonomics

Can you elaborate on how the developer ergonomics are better?

I'm just a bystander using GraphQL for hobby projects, but I have to admit I'm disappointed to see a future where comments are officially used to embed directives like this. I was hoping for something like `@description`. When you're in a position to update the GraphQL spec itself, why not just use a different syntax and remove all possible ambiguity? Even just a single different leading character instead of `#`?

(I know people widely complain about JSON's lack of comments, given it's now popularly adopted for config files, but I think the reasoning was quite prescient – if they were supported, comment directives would have sprung up in every JSON parser and hurt interoperability. Sure, this discussion would effectively make comments-as-descriptions "official", but still – why not introduce a new syntax, remove ambiguity, and let comments be comments?)

sergehuber:
Just to give another (less frequent but possibly important ?) use case for the type definition language: I'm working on a specification at OASIS that is interested in using GraphQL as part of the transport language, but for that we would need the type language to be specified and relatively stable. So this is a case where we cannot use any programming language specification.

leebyron:
@exogen - developer ergonomics are better specifically because descriptions are often multiple lines and include markdown syntax. Embedding that into a directive would make reading the schema language much more difficult, and legibility is primarily important.

For example, the description appearing before the `__TypeKind` introspection type in graphql-js:

```graphql
# The fundamental unit of any GraphQL Schema is the type. There are many kinds of
# types in GraphQL as represented by the `__TypeKind` enum.
#
# Depending on the kind of a type, certain fields describe information about that
# type. Scalar types provide no information beyond a name and description, while
# Enum types provide their values. Object and Interface types provide the fields
# they describe. Abstract types, Union and Interface, provide the Object types
# possible at runtime. List and NonNull types compose other types.
type __Type {
  kind: __TypeKind!
  name: String
  description: String
  interfaces: [__Type!]
  possibleTypes: [__Type!]
  inputFields: [__InputValue!]
  ofType: __Type
}
```

Vs

```graphql
type __Type @description(msg: "The fundamental unit of any GraphQL Schema is the type. There are many kinds of types in GraphQL as represented by the `__TypeKind` enum.\n\n Depending on the kind of a type, certain fields describe information about that type. Scalar types provide no information beyond a name and description, while Enum types provide their values. Object and Interface types provide the fields they describe. Abstract types, Union and Interface, provide the Object types possible at runtime. List and NonNull types compose other types.") {
  kind: __TypeKind!
  name: String
  description: String
  interfaces: [__Type!]
  possibleTypes: [__Type!]
  inputFields: [__InputValue!]
  ofType: __Type
}
```

leebyron:
It's also not too late to introduce a new kind of token for descriptions. There are certainly pros and cons to using comments for this.

Pros:

* It follows prior art. Comments in source code as descriptions for the following declaration is prevalent in almost every programming language. Nearly every popular language has some tool which consumes these comments to use as documentation.
* It's obvious, making it easy to fall into [the pit of success](https://blogs.msdn.microsoft.com/brada/2003/10/02/the-pit-of-success/) - essentially there is no other obvious way to add descriptions, and the reasonable assumption of just putting them in comments "just works".
* It minimizes additional syntax to the language. In fact, no additional syntax is necessary. This is beneficial for interoperation with existing tools.
* For tools which don't care about descriptions, they can be safely treated only as comments.

Cons:

* Parsers need to be more sophisticated and cannot simply skip over comments but instead need to track which comment lines appear before which definitions so they can be interpreted as descriptions.
* There is potential for ambiguity where commenting out lines could cause them to be misinterpreted as a description for the following line.
* Actual comments (ignored) need to be separated from descriptions (currently proposed via an empty line)

Other proposals suggest a different leading character, similar to comments:

```
# comment
## description
## using `##` chars
type SomeType {
  field: String
}
```

Or

```
# comment
% description
% using `%` char
type SomeType {
  field: String
}
```

Another option is to use something like a [docstring](https://en.wikipedia.org/wiki/Docstring):

```
"""
This is a python-inspired doc string style description
"""
type SomeType {
  field: String
}
```

Or

```
###
Similar style, perhaps interpreted as a multi-line comment
###
type SomeType {
  field: String
}
```

stubailo:
I've grown accustomed to the comment approach, but it would be nice to also add a way to have non-visible comments. I think the main issue for me is you can't add internal comments like "# TODO add a limit to this field later" or similar.

I kind of like the "##" idea since it reminds me of "/**" in javadoc.

tendant:
I would vote for parser friendly solution. It is easy to support description in parser, if it is defined using something like "%", "##" or "@".

It harder to support description with existing comment format. It takes multiple pass to identify which comment is description.

spec/Section 2 -- Language.md
```diff
@@ -938,3 +943,292 @@ and operations.
...
+  x: Float
+  y: Float
+}
+```
+
+### Type Extension
+
+TypeExtensionDefinition : extend ObjectTypeDefinition
+
+Type extensions may be used by client-side tools in order to represent a
```
stubailo: This could also be used on the server if you are using the IDL to define your schema, but want to split one type across multiple files. In particular, this would be common for `Query` or `Mutation`, where you might want to do:

```graphql
type Todo {
  text: String!
}

extend type Query {
  allTodos: [ Todo ]
}
```

In the super ideal case, `buildASTSchema` would take this into account and merge fields.

Relevant issue on graphql-tools: https://github.com/apollostack/graphql-tools/issues/182

leebyron:
> I would vote for parser friendly solution.

I disagree. We should be aiming for whatever will have the best developer experience. Parsers are written rarely and by people who are capable and can write tests. Schema in the language will be written far more often. That's what we should be optimizing for.

leebyron:
Using comments for descriptions has a lot of advantages, but we'll need to decide if it's worth these drawbacks:

* There is potential for ambiguity where commenting out lines could cause them to be misinterpreted as a description for the following line.
* Actual comments (ignored) need to be separated from descriptions (currently proposed via an empty line)

Using some other mechanism for adding descriptions can solve these, but could reduce the benefits of using comment lines or add drawbacks of their own.

The drawbacks for `##` is that it's easy to get wrong and miss a `##` for a `#` and accidentally lose a documentation. It adds some ambiguity with comments, and I could careless commenting out of lines repeatedly converting a `#` into a `##` and falling into the same drawbacks.

The drawbacks for `%` is that it consumes a punctuator not in use (those are fairly precious) and it's not particularly easy to remember, nor is there much precedence for such a documentation scheme in modern languages so it's more to learn for newcomers.

Multi-line comments and Doc-strings are both interesting to me in that while they're maybe slightly less ergonomic, they may add additional value elsewhere (though not in a way that solves a real need at the moment)

**Multi-line comments**:

This adds two different comment forms to GraphQL. I originally avoided this in the sake of simplicity. Any multi-line comment can be easily represented as multiple single-line comments.

```graphql
###
Multi 
line 
description
###
type Foo {
  bar: String
}
```

Perhaps there's some value in including them, even if only for this use case. Though I think they still suffer from the same drawbacks of single-line comments, just to a lesser degree. 

**Doc-strings** would really be doing two things: 

* Adding a new kind of multi-line string literal that presumably could be used anywhere string literals are used today. 
* Optionally parsing `String` before parsing anything that may have a description.

That's interesting in that very short descriptions could be inline with typical string literals:

```graphql
"""
Multiline string 
description of Foo
"""
type Foo {
  "Short description of a bar field"
  bar("arg described inline" arg: Int): String
}
```

Adding this multiline string literal probably otherwise had only minor value, but one use case that comes to mind is deprecation warnings:

```graphql
type Foo {
  bar @deprecated(reason: """
    There's a lengthy reason why bar is deprecated
    that takes multiple lines to fully describe, and
    fitting into a single line string literal would be illegible
    """)
}
```

I think that's a pretty interesting use for multi-line strings and brings some parity between descriptions and deprecation reasons.

The drawbacks of strings as descriptions I think is just ergonomic. It's perhaps non obvious that you should use a string literal instead of a comment to describe something. Instead newcomers to the language may see that string and be curious how it's affecting the schema. This is a common drawback to docstrings in languages like Python and Clojure - it's easy for those less familiar with the language to misinterpret the strings a important to runtime behavior and become confused.

exogen:
I really like the Python-like docstring approach and was thinking of suggesting exactly the same thing :)

> it's easy for those less familiar with the language to misinterpret the strings a important to runtime behavior and become confused.

Aren't they indeed important to runtime behavior and thus this isn't a misinterpretation? One of the major reasons I don't like the `#` comment approach is because we don't expect that content to affect, nor be available at, runtime. But descriptions are available at runtime and are used in the GraphiQL "Docs" browser via introspection and such.

(Maybe I'm misunderstanding what you mean by "runtime behavior" though.)

leebyron:
I like the docstring approach as well, but I do think it's important to be sober about the drawbacks.

> Maybe I'm misunderstanding what you mean by "runtime behavior" though.

For example:

```clojure
(defn myFn
  "Some fn"
  [x]
  (doSomethingWith x))
```

If you weren't familiar with clojure, you might encounter this and be curious how this argument to the `defn` function was being interpreted and used at runtime or perhaps how it might affect performance or API design. Only once you become more familiar would you realize that `defn` is a macro, not a function, and that none of those early concerns apply.

It's mostly about matching existing mental models. I think if we've been docstring documenting our stuff since the 70's that this sort of thing would be obvious to all of us.

For GraphQL I think the same thing applies:

```graphql
type Foo {
  bar("this is an argument" arg: String): String
}
```

Is that string literal another argument in some way? Is it some hint to `bar`? Is it the default value for `arg`? If that string gets longer will it affect performance? I think it's totally reasonable to expect some portion of people who encounter this for the first time to be confused with these sorts of questions or probably others I haven't thought of.

stubailo:
As a concrete example, how can I do this? https://github.com/apollostack/GitHunt-API/blob/01f91419f16c0eaf93ca74e64f270df8d9195f1b/api/sql/schema.js#L54

```
  # The SQL ID of this entry
  id: Int!
  # XXX to be changed
  vote: Vote!
}
```

This is a real "comment" in the sense that it's an internal note in the code.

Would that need to be written with a blank line in between?

```
  # XXX to be changed

  vote: Vote!
```

That kind of seems to ruin the idea of ignoring whitespace.

leebyron:
> That kind of seems to ruin the idea of ignoring whitespace.

Yeah it's definitely a drawback of the approach.

DomKM:
I'd vote for letting comments be comments, not descriptions. As mentioned above, docstrings have worked out well in Clojure. It is reasonable for some people to be confused by new syntax the first time they see the syntax. However, using comments as descriptions is confusing, regardless of familiarity with syntax. It conflates developer-targeted documentation with user-targeted documentation. Please don't use comments to provide descriptions. Any alternative (metadata, directives, totally new syntax, etc.) is better.

stubailo:
Perhaps descriptions could be directives?

```graphql
myField: String! @description(text: "
  Get the content of myField!
")
```

Also, is there any reason not to have `"` be a multiline string? I don't know if there is value in having separate syntax for single-line and multiline strings, but I could be missing something.

luolong:
Now, I may be a little late to the discussion, but I would like to weigh in with my 2 cents on the topic of description syntax.

I would greatly prefer the double-quoted string syntax over any other proposal.

Furthermore, I would make strings multiline by default. That is -- Docstrings would be terminated by a next matching unescaped double quote. Even if the matching quote character is on another line.

Escaping double quote character would be achieved by prefixing it with with a backslash (`\`). Common escape sequences `\n`, `\t`, `\\` `\"` would be accepted and replaced with appropriate characters.

I would reserve triple quoted strings for verbatim string syntax. This means, anything you write within triple quotes will be taken as a verbatim value (no escape sequences needed). Only thing that terminates a verbatim string is a matching triple quotes pair.

See how [string literals have been defined in Ceylon language](https://ceylon-lang.org/documentation/1.3/tour/basics/#string_literals) for example.

```graphql
"Long description of what the Foo type is all about
 that spans multiple lines."
type Foo {
  """
  Definition of the `bar` attribute. 

  I want to have code sample here with embedded HTML snippet,
  so I should use triple quoted doc string to make my life easier:
  <script type="application/graphql">
  query{
    foo { bar("some argument value") }
  }
  </script>
  """
  bar("this is an argument." arg: String): String
}
```

spec/Section 2 -- Language.md
```diff
@@ -938,3 +943,239 @@ and operations.
...
+interface Node {
+  id: ID
+}
+```
+
+#### Union
+
+UnionTypeDefinition : union Name = UnionMembers
+
+UnionMembers :
```
swolchok: Isn't `UnionMembers` the same as `NamedType+`?

swolchok: Oh, the `|` is meant to be taken literally. That solves an apparent grammatical ambiguity too.

emilebosch:
@leebyron Whats the status on this? I'm really waiting to get directives to be allowed on any location. Can we maybe split it in the multi-line comments and the allow-directives-on-everything to be merged first?

spec/Appendix B -- Grammar Summary.md
```diff
@@ -169,3 +170,57 @@ NonNullType :
...
+
+UnionMembers :
+  - NamedType
+  - UnionMembers | NamedType
+
+EnumTypeDefinition : enum Name Directives? { EnumValueDefinition+ }
+
+EnumValueDefinition : EnumValue Directives?
+
+EnumValue : Name
```
kassens: Enum value is already defined above.

sbussard:
Why do interface fields have to all be the same type?

leebyron:
@sbussard This is probably the wrong issue to discuss that, but implementation of interface fields do not have to be the same type, they do however have to be covariant (http://facebook.github.io/graphql/#sec-Object-type-validation)

IvanGoncharov:
@leebyron I think it should be forbidden to add directives on argument definitions of directives and on type definitions used by directives arguments.
When I tried to implement parsing of applied directives from IDL ( the first version of https://github.com/graphql/graphql-js/pull/746) it proved to be very complex. Part of complexity caused by the fact that directive argument and also types of those arguments could have directives applied on them. E.g.:
```graphql
directive @testDirective(
  arg: SomeScalar @testDirective(arg: "test")
  #                    ^^^^ here
) on ARGUMENT_DEFINITION | SCALAR
scalar SomeScalar @testDirective(arg: "test2")
#                      ^^^^ and here
```
And here is circular dependency between two directives:
```graphql
directive @testA(
  arg: SomeScalar @testB(arg: "test")
) on ARGUMENT_DEFINITION
directive @testB(
  arg: SomeScalar @testA(arg: "test")
) on ARGUMENT_DEFINITION
```
Without such restriction in place, it will be hard to implement full support of directives in tools and libraries. That’s why I think it's important to add it into this release.

spec/Section 2 -- Language.md
```diff
@@ -980,3 +985,292 @@ and operations.
...
+```
+
+**Default Root Types**
+
+While any type can be the *root type* for a GraphQL query or mutation operation,
+GraphQL type system definitions can omit the schema definition when the query
+and mutation root types are named `Query` and `Mutation`, respectively.
+
+Similarly, when serializing a GraphQL schema using the type system language, a
+schema definition should be omitted if only uses the default root type names.
```
gjtorikian: if _it_ only uses

leebyron:
@IvanGoncharov I'm curious to what that complexity is specifically so we can address it. This is a fairly sensitive balance between simplifying the schema grammar (in which currently, arguments on directives are no different from arguments anywhere else) and being restrictive as to where directives can be placed.

IvanGoncharov:
> This is a fairly sensitive balance between simplifying the schema grammar (in which currently, arguments on directives are no different from arguments anywhere else) and being restrictive as to where directives can be placed.

@leebyron Instead of forbidding to apply directives on directives argument, the spec can forbid any recursion in directive definition similar to how it's done for the Fragments:
> The graph of fragment spreads must not form any cycles including spreading itself.
--------------------------
>  I'm curious to what that complexity is specifically so we can address it. 

According to this PR, a directive on arguments is an:
> indicator that they should be evaluated differently by a validator, executor...

That allows directive to change parsing, coercion, and validation of arguments. That means that to build directive object(e.g. `GraphQLDirective`) you need to figure out if the same directive is correctly applied on its own arguments. For example:

```graphql
directive @coerceDifferently(
  arg: String @coerceDifferently(arg: "invalidValueAccordingToChangedCoertion")
) on ARGUMENT_DEFINITION
```

spec/Appendix B -- Grammar Summary.md
```diff
@@ -172,3 +173,57 @@ NonNullType :
 Directives : Directive+
 
 Directive : @ Name Arguments?
```
JeffRMoore: I think this needs to express that argument values are constants for directives used in definitions.  The following is allowed by the grammar right now, I believe.
```
scalar foo @deprecated(reason: $foo)
```
Unless I'm missing something, there's no enclosing `VariableDefinitions` for schema to define variables, so this doesn't seem to make much sense.

The arguments definitions are:
```
Arguments : ( Argument+ )
Argument : Name : Value
Value[Const] :
```

I think in this case, it should be something like:

```
Directive : @ Name Arguments[Const]?
```
That involves also changing `Arguments` and `Argument`, right?  I'm not sure what the right thing is here, but I think the current definition is ambiguous.

pksunkara:
Are we planning to include some frequently used directives in the spec? For example, `@unique` can be used for Scalar typed field whenever we want to say that the value of the field is unique.

> Investigate impact on validation rules

I do not understand what exactly this means. Can anyone explain it a little bit?

foomip:
Just out of curiosity, is EBNF or ABNF notation available? It would with make the creation of a parser for other third party libraries easier.

gjtorikian:
@foomip I'm actually very close to finishing a parser implementation like that, within the next week or so I'd say. I don't want to spam this thread but I can post it here if you like? I'll for sure tweet about it when it's done.

andimarek:
@foomip maybe this can help you: https://github.com/graphql-java/graphql-java/blob/master/src/main/antlr/Graphql.g4 

This is a grammar for graphql including IDL.

SerafimArts:
@foomip or this =) https://github.com/SerafimArts/Railgun/blob/master/resources/grammar.pp

foomip:
Thanks @andimarek @SerafimArts will have a look. @gjtorikian would be great to have a look at your implementation, once you are ready :)

Just to give you guys some background into what trying to do. I am looking at creating a parser that takes a GraphQL SDL file and converts it to an equivalent EDN definition for Clojure's Lacinia GraphQL implementation - https://github.com/walmartlabs/lacinia. Not sure if it is 100% compatible but I think most of the SDL structure should be portable to the equivalent EDN notation.

SerafimArts:
@foomip souns good. But... the current standard is a draft and contain a little flaws. Take it into account.

foomip:
@SerafimArts will do thanks. Gotta start somewhere :)

spec/Appendix B -- Grammar Summary.md
```diff
@@ -172,3 +173,57 @@ NonNullType :
...
+  - ScalarTypeDefinition
+  - ObjectTypeDefinition
+  - InterfaceTypeDefinition
+  - UnionTypeDefinition
+  - EnumTypeDefinition
+  - InputObjectTypeDefinition
+
+ScalarTypeDefinition : scalar Name Directives?
+
+ObjectTypeDefinition : type Name ImplementsInterfaces? Directives? { FieldDefinition+ }
```
gjtorikian: I have noticed that some implementations (like `GraphQL Ruby`) do emit objects in the IDL that lack fields, such as `type Foo {}`. Will that be taken into consideration here? (Making this `FieldDefinition*`.)

SerafimArts: @gjtorikian does this separation make sense?

mohawk2: Also `graphql-js` expects an empty one to work: https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/schema-parser-test.js#L168

```
const body = 'type Hello implements World { }';
```

gjtorikian:
> would be great to have a look at your implementation, once you are ready :)

@foomip Finally got around to [writing about](https://www.gjtorikian.com/posts/2017/08/15/wrapping-rust-in-ruby-to-parse-gql) and releasing the lib.

spec/Appendix B -- Grammar Summary.md
leebyron: Thanks for identifying this! While this would result in an invalid final schema - it's still useful to parse as an intermediate schema - I'll correct for this

spec/Appendix B -- Grammar Summary.md
leebyron: (Sorry for delayed feedback)

You're absolutely correct. Variables should not be allowed by the grammar. I will account for that

spec/Appendix B -- Grammar Summary.md
mohawk2: @leebyron Could you help me with how it's useful to parse something knowingly invalid? Or is that in the expectation of an `extends`?

mohawk2:
@foomip Another possibility, using Pegex: https://github.com/graphql-perl/graphql-perl/blob/master/graphql.pgx

schickling:
@leebyron would be great if this PR could be renamed to `[RFC] GraphQL SDL additions` as discussed in the working group meeting. 👍

spec/Section 2 -- Language.md
```diff
@@ -980,3 +985,292 @@ and operations.
...
+
+Since the validation of a GraphQL document includes ensuring that any directives
+used are defined and used correctly, defining a directive allows for a validator
+to be aware of all possible validation rules.
+
+As part of validation, directives are only allowed to be used in locations
+that are explicitly declared. Directive locations must be one of:
+
+ * `QUERY`
+ * `MUTATION`
```
JonasRudolph: I would suggest to add ``` * `SUBSCRIPTION` ``` to this list.
If `SUBSCRIPTION` was omitted intentionally, can someone explain why?

Kind regards :)

spec/Section 2 -- Language.md
```diff
@@ -980,3 +985,292 @@ and operations.
...
+
+ArgumentsDefinition : ( InputValueDefinition+ )
+
+InputValueDefinition : Name : Type DefaultValue? Directives?
+
+Object types represent a list of named fields, each of which yield a value of a
+specific type. Each field itself may accept a list of named arguments.
+
+Objects may implement Interface types. When implementing an Interface type, the
+Object type must supply all fields defined by the Interface.
```
JonasRudolph: I really like, that interfaces exists in GraphQL. But as a computer scientist this feels unnatural for me.
Besides the feeling I can think of at least the following problem:
1. Alice defines an Interface A with Field x in her Project.
2. Bob adds the project of Alice as dependency and `implements` her Interface A in type B of his project.
3. Alice adds Field y to Interface A and publishes her changes.
4. Bob updates his dependencies -> The GraphQL implementation he uses reports an error.

This currently means, that adding a field to an interface is a major change and thus not backwards compatible.
We would get around this problem, if `the object type MAY supply the fields defined by the interface`.

My feeling is, that forcing the user of an Interface to repeat the same fields, does not add value while reducing maintainability.

What is the reason that one has to repeat all fields that are defined in the interface? 

Kind regards :)

mohawk2: Surely that's literally the point of an interface, that those who "implement" it are actually doing so.

JonasRudolph: You agree with you about the definition of an interface and that the current usage of the `implements` keyword in GraphQL matches the definition. I was not able to correctly describe the issue.

So here is what I mean:
Let us define, that types that implement an interface automatically satisfy it. This does not require any Syntax addition or change. We only have to change one word in the languager definition (MUST-> MAY) above and do not need additions like mixins (#370).
There is at least one other person (@SerafimArts) suggesting this (https://github.com/facebook/graphql/issues/370#issuecomment-339848304).

In my view this is the simplest change in the spec that would at the same time improve the productivity of many people because they do not have to repeat the fields of the interface over and over again.

What speaks against this?

mohawk2: > What speaks against this?

The fact that this would render interfaces completely meaningless?

JonasRudolph: Seems like it is still not clear what I mean. 
I suggest, that an type that implements an interface implicitly has all the fields of the interface.

SerafimArts: @mohawk2 
>The fact that this would render interfaces completely meaningless?

Why is that? The interface defines those fields that the type (object) is obliged to contain. But this does not mean that we have to copy them from the interface to the implementation.

Interfaces in the modern world are needed in order to oblige classes to implement methods. 
In GraphQL/SDL, there are no methods, no realizations, and this turns out to be stupid Ctrl+C -> Ctrl+V.

### e.g Current behavior

```graphql
interface B {
    field: Any
}

# Type A with field "field"
type A implements B { 
    field: Any # Copy+Paste
}

 # Error: Field "field" required
type C implements B {
}
```

### e.g Proposal behavior

```
interface B {
    field: Any
}

# Type A with field "field"
type A implements B {
    field: Any # Copy+Paste
}

# Type C with field "field"
type C implements B {
    # Auto resolving. We pretend that the definition in the Object completely repeats the definition in the Interface.
}
```

mohawk2: I would suggest this means you should generate your schemas rather than manually maintaining them?

SerafimArts: @mohawk2 Schemes of introspection and so generated... Or what is the question? Generation of SDL? How does a little tuning of the gql standard and its implementation?

JonasRudolph: @mohawk2 Can you explain what you mean? I also don't understand your sentence :)

mohawk2: I am trying to indicate that I believe you are solving the wrong problem, and that the way to avoid incorrectness in IDL representation of implementers of interfaces is to automatically generate those IDL representations. In any case, I do not thing the IDL should change in the way you suggest.

SerafimArts: Generate SDL? :D Lol?

leebyron: To close the loop on this conversation, we've now discussed this thoroughly. This is a tradeoff between explicitness (requiring object definitions to declare all fields they implement) and terseness (imply an object contains the fields implements in an interface).

There are a number of reasons why we've chosen to lean in favor of explicitness in this instance:

* Readability: The SDL is designed to be read far more often than it is written (in fact, in many cases it is "written" by a code generation script). Because of this, it's highly desirable to see all defined fields in one place and not have to follow references to interfaces to understand the full set of fields defined.

* Removal of adherence ambiguity: Consider the following example schema with implicitly defined fields:

  ```graphql
  interface Foo {
    self: Foo
  }
  interface Bar {
    self: Bar
  }
  type FooBar implements Foo, Bar {}
  ```

  In this case, what is the type of the field `self`? While this sort of ambiguity can be solved by additional checks and constraints, it complicates tools and can lead to pitfalls when designing APIs. Requiring fields to be defined removes this complication and potential for ambiguity.

spec/Appendix B -- Grammar Summary.md
leebyron: To resolve this, rather than allow parsing `type Foo {}` which you correctly point out would be invalid, I've changed the definition of extends to explicitly support a case to extend a type without adding fields.

leebyron:
@vergenzt - I think that's certainly a topic worth discussion.

Currently it's not uncommon for client codebases to define both their schema extensions and set of definitions across the same documents. We would expect these definitions to be separated by tooling before validation and execution.

stubailo:
Yeah totally agree - lots of documents will include both type info and queries in one, so it would be great to have a well-defined way for them to live side by side.

OlegIlyenko:
After releasing most recent SDL spec changes in sangria, I got some user feedback:

https://github.com/sangria-graphql/sangria/issues/308

I would like to make a suggestion regarding these grammar changes:

```
ObjectTypeDefinition : Description? type Name ImplementsInterfaces? Directives[Const]? FieldDefinitions
FieldDefinitions : { FieldDefinition+ }
InterfaceTypeDefinition : Description? interface Name Directives[Const]? FieldDefinitions
EnumTypeDefinition : Description? enum Name Directives[Const]? { EnumValueDefinition+ }
InputObjectTypeDefinition : Description? input Name Directives[Const]? { InputValueDefinition+ }
```

At the moment the parser defines that the field list must contain at least one field (or enum value in case of the enum). My suggestion is to implement this restriction as a validation rule and allow field list to be empty at the parsing stage. 

Effectively, this approach yields very similar results, but also provides following advantages:

* Since the field list verification takes place at the semantic validation stage, it might produce better validation errors with deeper semantic insights
* Validation provides a very robust mechanism for semantic analysis. If the field list of generated based on, let's say, directive information, then more complex validation and overarching workflow can be implemented. Right now the validation is baked into the parser and provides very little flexibility in this regard 
* It might provide a smooth migration path for people who are using types with empty field lists right now.

Would love to hear your thoughts on this :)

leebyron:
Good feedback, @OlegIlyenko. I still had some reservations about the any/many fields in field sets. My previous understanding was that the use cause for empty fields was for type extensions, where an extension sought to add a directive or interface and no additional fields.

A minor concern is consistency with the rest of the grammar. For example [SelectionSet](http://facebook.github.io/graphql/draft/#SelectionSet) and [Arguments](http://facebook.github.io/graphql/draft/#Arguments) and [VariableDefinitions](http://facebook.github.io/graphql/draft/#VariableDefinitions) share this property of requiring the grammar to have non-empty lists - similarly these could be validation rules instead, however requiring them in the grammar ensures unambiguous text.

Another use case I've come across (and one I personally wouldn't advocate, but people seem to find value in it) is to define a type with no fields followed by numerous extensions:

```graphql
type Example {}
extend type Example {
  foo: String
}
extend type Example {
  bar: Int
}
```

This pattern seems to reduce the legibility of the document, however it certainly still seems semantically valid and unambiguous in its entirety.

While these use cases for empty field sets are not particularly compelling to me personally, I think the strongest argument is simply that people are already using them and smoothing the migration path from the experimental SDL to the spec SDL is valuable in its own right, and it's the cost we pay for experimenting in the open.

OlegIlyenko:
It's good that you mention this particular use case, I also encountered on a few occasions. I think it often desired when the schema definition is split between different files.

I think I also would prefer to keep it consistent with other languages contracts. I wonder whether we can also skip the curly braces and define the syntax like this:

```
InputObjectTypeDefinition : Description? input Name Directives[Const]? InputValueDefinitions?
ObjectTypeDefinition : Description? type Name ImplementsInterfaces? Directives[Const]? FieldDefinitions?
InterfaceTypeDefinition : Description? interface Name Directives[Const]? FieldDefinitions?
EnumTypeDefinition : Description? enum Name Directives[Const]? EnumValueDefinitions?

EnumValueDefinitions : { EnumValueDefinition+ }
InputValueDefinitions : { InputValueDefinition+ }
FieldDefinitions : { FieldDefinition+ }
```

leebyron:
I like that suggestion, @OlegIlyenko. That formulation almost makes it more immediately apparent that no fields are defined rather than an empty set being defined. However one drawback is that it still presents a breaking change against the experimental SDL, however at least it is very easy to fix by simply removing `{}`

OlegIlyenko:
Indeed, this is a concern. I think it should not be a big problem if people just need to search and replace the codebase. On the other hand, of SDL is stored in the external DB then it might involve multi-staged migration. @BjRo would be interesting to hear your thoughts on this matter.

@stubailo It would be also interesting to hear your thoughts on this suggestion. Do you think that proposed parser/validation separation might provide value beyond the ease of migration?

BjRo:
@OlegIlyenko @leebyron First of all, thx for discussing this. I like your ideas. It would definitely make things a bit easier to manage and also a bit tidier in our SDL schema. We have both cases (multiple files and empty field definitions).

Regarding the migration aspects, I think it depends on the setup. For us it's relatively easy to adopt such changes since we have the SDL files, the application code, our project documentation and how-to-guides in a single source code repository. A pull request on our side is able to change all affected portions at once and we can do this more or less manually. A strict behaviour of the parser and the `GraphQL` framework can be handled in this scenario.

SDL stored in an external DB looks like an entirely different beast to me, since you can't change all portions at once. One way to address this is to have the `GraphQL` server running in a compatibility mode (that allows both formats) and then doing the schema update separately. If I'm not mistaken, this is also what Sangria allows you currently to do (though it needs to be specifically enabled).

I think the community would benefit from a schema migration tool for such breaking changes, so that the schema updates behave fairly consistent and can be mostly automated. It would be interesting to hear @schickling thoughts on this (given that `Graphcool` has externally stored SDL schemas).

leebyron:
I believe this is now in a finalized state.

spec/Section 3 -- Type System.md
```diff
@@ -674,7 +963,68 @@ Interface types have the potential to be invalid if incorrectly defined.
...
+```
+
+Interface type extensions may choose not to add additional fields, instead only
+adding directives.
+
+In this example, a directive is added to a `NamedEntity` type without
+adding fields:
+
+```graphql example
+extend type NamedEntity @addedDirective
```
vergenzt: Should this be `extend interface NamedEntity @addedDirective`?

spec/Section 3 -- Type System.md
```diff
@@ -781,26 +1181,67 @@ should only allow such values as enum input values. Otherwise, for most
 transport serializations that do not, strings may be interpreted as the enum
 input value with the same name.
 
+**Type Validation**
+
+Enum types have the potential to be invalid if incorrectly defined.
+
+1. A Enum type must define one or more unique enum values.
```
vergenzt: Is there a reason the validation rules are saying "A Enum type" or is it a typo? Happens here and in the "Enum Extensions" type validation section.

spec/Section 3 -- Type System.md
```diff
@@ -962,21 +1424,92 @@ query withNullableVariable($var: String) {
...
-engine supports.
+DirectiveDefinition : Description? directive @ Name ArgumentsDefinition? on DirectiveLocations
+
+DirectiveLocations :
+  - `|`? DirectiveLocation
+  - DirectiveLocations | DirectiveLocation
+
+DirectiveLocation : one of
+  `QUERY`                 `SCHEMA`                `ENUM`
+  `MUTATION`              `SCALAR`                `ENUM_VALUE`
```
vergenzt: Should we be consistent throughout these names with the `_DEFINITION` suffix? I was confused at first what it meant for a directive to apply to `SCALAR`. I at first thought it meant that something like `query { someField(integerArgument: 5 @mydirective) }` would be valid before I realized it was talking about definitions.

Or maybe could we split them into two groups, `TypeSystemDirectiveLocation` and `OperationalDirectiveLocation` and say that `DirectiveLocation` must be from one of the two groups?

leebyron:
Thanks for the review @vergenzt! I'll go through and address the issues you spotted!

spec/Section 3 -- Type System.md
leebyron: Because these are in use, it would be pretty painful to change these values, however I like the idea of grammatically splitting across two groups for clarity

vergenzt: Makes sense. 👍 Thanks!!

spec/Appendix B -- Grammar Summary.md
```diff
@@ -224,7 +224,9 @@ ObjectTypeExtension :
   - extend type Name ImplementsInterfaces? Directives[Const]
   - extend type Name ImplementsInterfaces
 
-ImplementsInterfaces : implements NamedType+
+ImplementsInterfaces :
+  - implements `&`? NamedType
```
vergenzt: Where did the ampersand syntax come from? Has this been implemented anywhere and found to be preferable over simple whitespace/comma separation? 🤔 

Looks like graphql-js just uses whitespace/commas:
https://github.com/graphql/graphql-js/blob/master@{2017-12-19}/src/language/parser.js#L921-L923

leebyron: This solves an ambiguity we should have caught sooner but was found in practice.

https://github.com/graphql/graphql-js/pull/1166 documents the ambiguity

https://github.com/graphql/graphql-js/pull/1169 implements this same proposed fix

spec/Section 3 -- Type System.md
```diff
@@ -962,21 +1442,113 @@ query withNullableVariable($var: String) {
...
+```graphql example
+directive @example on
+  | FIELD
+  | FRAGMENT_SPREAD
+  | INLINE_FRAGMENT
+```
+
+Directives can also be used to annotate the type system definition language
+as well, which can be a useful tool for supplying additional metadata in order
+to generate GraphQL execution services, produce client generated runtime code,
```
acjay: > [...] produce client generated runtime code

But if I understand https://github.com/graphql/graphql-js/pull/746#issuecomment-301554231, directives won't generally be visible in client code. Or am I misreading this?

acjay: Perhaps I'm just struggling with confusion over what clients can be expected to have access to when consuming a GraphQL service. Here's my understanding: a GraphQL API can either be defined using SDL or directly in code. So, it would seem to me that if directives would be visible to and usable by client code after consuming the SDL definition of the server, the same data would be accessible via introspection of a non-SDL schema. Or that a non-SDL schema could render to a generated SDL file (e.g. Sangria's [Schema materilialization](http://sangria-graphql.org/learn/#schema-materialization)) with the full expressiveness of an "a priori SDL" schema.

After reading @leebyron's comment, I started thinking, "Oh! Directives are used in SDL to guide the _implementation_ of a GraphQL schema, not to provide client-consumable metadata." Which is why that clause is throwing me off.

chrisregnier:
I agree with @acjay, I was under the impression that directives wouldn't be available through introspection? Although I think there's tons of potential for allowing directives to be available through introspection and allowing developers better ways to grow their use of the language in a dynamic way without requiring changes to the spec for every new concept. 

Anyways, in the case that they are available for introspection, I think either the ability to add directives to directive definitions needs to be added back, or there needs to be a visibility concept for directives.

For example a `@cost` directive (used for declaring rough cost of a field) likely has the purpose of being passed along to users of the schema, so it should be available through introspection. However a `@cache(maxAge:60000)` directive is likely used more as config into the implementing framework, so it shouldn't be available through introspection, or else implementation details bleed out.

This means there's a visibility requirement, and I see two possible solutions.
1. Directives need a visibility concept and the SDL adds a public/private keyword for DirectiveDefinition. Frameworks can then use the visibility to resolve introspection queries.
2. The DirectiveDefinition allows other directives be used on it (which isn't allowed right now) and then frameworks are responsible for looking for something like a `@private` directive filtering things appropriately
eg: `directive @private @cache(maxAge: Int) on FIELD_DEFINITION`

I like 1. better, but in another scenario where I'm stitching schemas automatically through a proxy service, passing on that `@cache` directive could be very useful. Although I wouldn't want it to bleed past the proxy even though I may want other directives to be allowed through, so I'd need a custom directive anyways. Then all of my services have to look for something like a `@private(hops=2)` directive or just pass something like `@namespaced_for_proxy_private` directive through, so eventually the field will be marked private. So if I need a custom directive here and to rely on the frameworks anyways 2. as a general approach may be better.

spec/Appendix B -- Grammar Summary.md
```diff
@@ -176,6 +180,124 @@ NonNullType :
...
+ObjectTypeDefinition : Description? type Name ImplementsInterfaces? Directives[Const]? FieldsDefinition?
+
+ObjectTypeExtension :
+  - extend type Name ImplementsInterfaces? Directives[Const]? FieldsDefinition
+  - extend type Name ImplementsInterfaces? Directives[Const]
+  - extend type Name ImplementsInterfaces
+
+ImplementsInterfaces :
+  - implements `&`? NamedType
+  - ImplementsInterfaces & NamedType
```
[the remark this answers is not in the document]
bbakerman: is it for this case (aka formatting??)

    type Foo implements
        & Bar 
        & Baz 
    {
       ...

?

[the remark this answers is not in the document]
bbakerman: Again it might be I am reading the grammar decl wrong but currently I think it allows

    type Foo implements & Bar implements & Baz

(because of the repeating ??)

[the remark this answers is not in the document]
IvanGoncharov: @bbakerman Yes leading `&` supported for formatting more context here: https://github.com/facebook/graphql/pull/323

> Again it might be I am reading the grammar decl wrong but currently I think it allows

No, it doesn't. According to the grammar `implements` can be used only in the begging.

spec/Section 3 -- Type System.md
```diff
@@ -674,7 +980,66 @@ Interface types have the potential to be invalid if incorrectly defined.
...
+5. Any directives provided must not already apply to the original Interface type.
+
+
+## Unions
+
+UnionTypeDefinition : Description? union Name Directives[Const]? UnionMemberTypes?
+
+UnionMemberTypes :
+  - = `|`? NamedType
+  - MemberTypes | NamedType
```
OlegIlyenko: I should be `- UnionMemberTypes | NamedType` as above, isn't it?

leebyron: Good catch. Looks like this bug only exists here and isn't replicated in the grammar summary

ermik:
@leebyron why does the spec still dated October 2016 at [http://facebook.github.io/graphql/](http://facebook.github.io/graphql/)?

IvanGoncharov:
@ermik Because it's the last official release.
You can find the draft of the next release (including SDL) here:
http://facebook.github.io/graphql/draft/

PR #92, opened 2015-09-12
Update the type syntax in examples to match the IDL

fson:
Use `ImplementsInterfaces` syntax from GraphQL IDL in the examples.


leebyron:
Thank you for this!

PR #99, opened 2015-10-01
[RFC] Make operation name optional.

leebyron:
Based on feedback from those using GraphQL and inspired by RelayQL's syntax, this makes operation names optional.

The primary value is the ability to express mutations (or eventually subscriptions) without being required to provide a name if the name adds no value for your use case. (Value typically being query disambiguation and logging). Systems which want to enforce a name for queries should have a separate linting/validation rule - such a rule would probably also want to ban "query shorthand"


dschafer:
I like it. Ship it.

PR #100, opened 2015-10-01
[RFC] Type condition optional on inline fragments.

leebyron:
Originally proposed by @josephsavona, this allows inline fragments to omit a type condition and thus always be considered for inclusion. If a type condition is omitted, no type refinement is done to the contained fields.

This enables a new kind of usecase where the only purpose for using an inline fragment is applying a directive to a grouping of fields:

```graphql
fragment Foo on Type {
  fieldA
  fieldB
  ... @include(if: $wat) {
    fieldC
    fieldD
  }
}
```


josephsavona:
ship it!

PR #106, opened 2015-10-14
Fix invalid fragment syntax in one of the examples.

dylanahsmith:
Updated a couple of examples to use the correct fragment definition, which uses ` on ` instead of `:` between the fragment name and the type name.


leebyron:
Thanks!

PR #108, opened 2015-10-14
Mention that variable values need to be coerced for execution.

dylanahsmith:
Variable value coercion is implied during input coercion for the types, but there isn't a mention of when it should be done.  I believe this underspecification can lead to implementations missing input coercion on variables, using the raw unvalidated variable values instead, as was the case with graphql-ruby until I recently reported the issue https://github.com/rmosolgo/graphql-ruby/issues/46.

The semantics of whether input coercion errors should result in the operation to fail or if the field resolution that uses the variable should fail was unclear.  [graphql-js coerces variables when building the execution context](https://github.com/graphql/graphql-js/blob/v0.4.7/src/execution/execute.js#L198), so it looks like query errors would result in the operation to fail, thus I documented this behaviour in the specification.

The input coercion also implied that `null` variable values would be accepted for Non-Null types without raising an error, which doesn't make sense for variables which don't have their values validated like literals, so I changed the sentence to say that a query error should be raised.  [graphql-js raises an error in this case.](https://github.com/graphql/graphql-js/blob/v0.4.7/src/execution/values.js#L107-L111)

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -735,11 +735,11 @@ query withNullableVariable($var: String) {
 Hence, if the value for a Non Null type is hard-coded in the query, it is always
 coerced using the input coercion for the wrapped type.
 
-When a Non Null input has its value set using a variable, the coerced value
-should be `null` if the provided value is `null`-like in the provided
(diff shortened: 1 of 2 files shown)


leebyron:
Nice work

PR #109, opened 2015-10-21
[WIP] adding 'subscription' language to GraphQL spec

skevy:
This PR adds language around the `subscription` operation type, which landed experimentally in `graphql-js@0.4.8`, to the GraphQL specification.

This is a WIP - there's still discussion to be had around subscriptions and what they mean for GraphQL, but I wanted to get the ball rolling on this so the can didn't get kicked too far down the road and also so that there was a place to discuss subscriptions in the context of the actual GraphQL specification.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -2,8 +2,8 @@
 
 Clients use the GraphQL query language to make requests to a GraphQL service.
 We refer to these request sources as documents. A document may contain
-operations (queries and mutations are both operations) as well as fragments, a
-common unit of composition allowing for query reuse.
+operations (queries, mutations, and subscriptions are all operations) as well as
+fragments, a common unit of composition allowing for query reuse.
 
 A GraphQL document is defined as a syntactic grammar where terminal symbols are
 tokens (indivisible lexical units). These tokens are defined in a lexical
@@ -192,12 +192,13 @@ OperationDefinition :
   - OperationType Name? VariableDefinitions? Directives? SelectionSet
   - SelectionSet
 
-OperationType : one of `query` `mutation`
+OperationType : one of `query` `mutation` `subscription`
 
-There are two types of operations that GraphQL models:
+There are three types of operations that GraphQL models:
 
   * query - a read-only fetch.
   * mutation - a write followed by a fetch.
+  * subscription - a read-only fetch followed by a subscription.
 
 Each operation is represented by an optional operation name and a selection set.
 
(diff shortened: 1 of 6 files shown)


taion:
Are there any blockers on this PR? graphql-js now expects `subscriptionType` to be present under `__schema` on introspection queries, but this isn't part of the spec at all.

skevy:
@taion I would presume that if there's a blocker at all, it's that subscription support in graphql-js is still experimental. Not sure how the GraphQL team wants to handle cases like that with the spec.

taion:
Possibly just update the spec then? I think this is causing the Python guys some trouble with setting up the Relay babel plugin; specifically on running the packaged introspection query against a non-graphql-js server.

spec/Section 2 -- Language.md
```diff
@@ -214,6 +215,19 @@ mutation {
 }
 ```
 
+This subscription operation will retrieve the current number of likes and create
```
leebyron: This bit is over-specified. A subscription doesn't have to do a retrieve then subscription, it probably only creates the subscription.

spec/Section 6 -- Execution.md
```diff
@@ -17,14 +17,16 @@ operations” section.
 ## Evaluating operations
 
 The type system, as described in the “Type System” part of the spec, must
-provide a “Query Root” and a “Mutation Root” object.
+provide a “Query Root” and a “Mutation Root” object, and can provide an optional
```
leebyron: This existing phrase isn't quite correct. Do you mind amending to:

The type system, as described in the "Type System" part of the spec, must provide a "Query Root" and may optionally also provide a “Mutation Root” or a "Subscription Root" object.

spec/Section 6 -- Execution.md
```diff
@@ -17,14 +17,16 @@ operations” section.
...
+“Subscription Root” object.
 
 If the operation is a mutation, the result of the operation is the result of
 evaluating the mutation’s top level selection set on the “Mutation Root”
 object. This selection set should be evaluated serially.
 
-If the operation is a query, the result of the operation is the result of
-evaluating the query’s top level selection set on the “Query Root” object.
+If the operation is a query or a subscription, the result of the operation is
```
leebyron: This is over-specified. The evaluation of a subscription should be considered experimental and intentionally under-specified at the moment.

leebyron:
Adding the introspection bit is a good idea, but the actual execution of these subscription queries should be intentionally under-defined for now. The spec should make it clear that subscription queries are experimental in a note.

skevy:
@leebyron np and totally understood. I will update based on your comments.

D1plo1d:
@skevy Any updates on this? We would like to implement subscriptions in Ruby but are waiting on an official spec and libgraphqlparser to do so.

skevy:
Closing in favor of #267

PR #112, opened 2015-10-22
Add table of contents for easier navigation

weblancaster:
Add table of contents for easier navigation.

After read some of the content one day I came back days later and wanted to just jump to the section I was looking for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,16 @@ data models.
 
 [![Public Slack Discussion](https://graphql-slack.herokuapp.com/badge.svg)](https://graphql-slack.herokuapp.com/)
 
+## Table of Contents
+
+  1. [Technical Preview Contents](#technical-preview-contents)
+  1. [Getting Started](#getting-started)
+  1. [Type System](#type-system)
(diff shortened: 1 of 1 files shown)


leebyron:
It would be better if the auto-generated TOC was just easier to navigate, which would be a change to https://github.com/leebyron/spec-md

weblancaster:
Sorry @leebyron but I did not understand what you mean by that.. are you suggesting that spec-md auto generate navigation based on the md? or I got wrong.

leebyron:
Yes. spec-md currently already generates a table of contents, so any improvements to it should be made there 



—

leebyron:
Sorry for the delayed response here (holidays).

The built spec http://facebook.github.io/graphql/ already includes a table of contents, so I believe this diff would create a second redundant one which we want to avoid.

leebyron:
The TOC is automatically generated by [spec-md](https://github.com/leebyron/spec-md/) so any improvements to the table of contents should be applied to that tool.

PR #122, opened 2015-11-16
Remove "is the result" typo

benbjohnson:
Remove duplicate "is the result" in _Evaluating a grouped field set_ > _Field entries_.


leebyron:
thank you thank you

PR #123, opened 2015-11-17
[Schema] Implementing interfaces with covariant return types.

leebyron:
This proposes loosening the definition of implementing an interface by allowing an implementing field to return a subtype of the interface field's return type.

This example would previously be an illegal schema, but becomes legal after this diff:

```graphql
interface Friendly {
  bestFriend: Friendly
}

type Person implements Friendly {
  bestFriend: Person
}
```

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -408,8 +408,20 @@ of rules must be adhered to by every Object type in a GraphQL schema.
 2. An object type must be a super-set of all interfaces it implements.
    1. The object type must include a field of the same name for every field
       defined in an interface.
-      1. The object field must be of a type which is equal to the
-         interface field (invariant).
+      1. The object field must be of a type which is equal to or a sub-type of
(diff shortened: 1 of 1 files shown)


dschafer:
This looks good. If we introduce union/interface types into the input system, we'll want to be careful (since the input type system should probably be contravariant and not covariant, but for the output type system this looks exactly right.

leebyron:
Yeah, currently the only possible pseudo-contravariance in input types in required (non-null) vs optional (nullable) but supporting that causes other issues. I think we'll need to divorce argument requirement from argument type nullability to support argument contravariance in the future.

dschafer:
> Yeah, currently the only possible pseudo-contravariance in input types in required (non-null) vs optional (nullable) but supporting that causes other issues. I think we'll need to divorce argument requirement from argument type nullability to support argument contravariance in the future.

Yep, I don't think we should do anything differently right now, just making a mental note if we ever do.

PR #129, opened 2015-12-15
Grammar fixes and section adjustment

bjornbytes:
- Fixes a few grammar issues.
- Currently in the validation section there are two subsections named "Operations" (numbered 5.1 and 5.7).  This naming collision means the skiplink in the TOC for 5.7 sends you to 5.1.  I believe the intention was to have the "Variables" section as 5.7, so I removed the second "Operations" section and adjusted the child sections accordingly.  If this isn't what was intended, let me know and I can rebase this out of the PR.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -870,7 +870,7 @@ fragment ownerFragment on Dog {
 
 ** Formal Specification **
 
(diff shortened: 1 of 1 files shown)


leebyron:
Most excellent. That fixed the link issue but also makes the docs clearer, so thank you!

PR #139, opened 2016-01-06
Fix a few issues with the validation tests

nuance:
Mostly tweaking the defined type system to include everything referenced, plus a few typos


leebyron:
This is nice work, thank you!

PR #142, opened 2016-02-12
Clarify that typed descriptions are required

admdikramr:
Clarify that "null" descriptions aren't OK. Metadata are very important.

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -214,7 +214,7 @@ Fields
 
 * `kind` must return `__TypeKind.SCALAR`.
 * `name` must return a String.
-* `description` may return a String or {null}.
+* `description` must return a String.
 * All other fields must return {null}.
 
 #### Object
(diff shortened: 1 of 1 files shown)


leebyron:
I wonder if this is the right thing? I think perhaps it should be okay for some things to be description-less.

admdikramr:
I think it maybe makes more sense to say that some things are "trivially
described," such as self-explanatory types (that can be marked as such
easily) or as types with inner types (where the description just wraps the
inner type).  But as this is now implemented in the code so that
getDescription is not nullable, updating the docs makes sense in any case...

leebyron:
It's currently implemented as allowing descriptions to be nullable - https://github.com/graphql/graphql-js/blob/master/src/type/introspection.js#L191

This was one of the things we intentionally loosened from FB's earlier version. When writing the spec for this version of graphql, we decided it would be too burdensome to always require descriptions at the introspection layer - but leave the decision on requiring descriptions as an implementation or lint-rule decision for individual users.

PR #145, opened 2016-02-18
add Validation/Variables/Variable Uniqueness

jjergus:


leebyron:
NB: implemented in the reference implementation in https://github.com/graphql/graphql-js/pull/296

PR #152, opened 2016-03-18
[RFC] Proposed change to directive location introspection

leebyron:
This specifies https://github.com/graphql/graphql-js/pull/317

This proposes a change to how we represent the ability to validate the locations of directives via introspection.

Specifically, this *removes* `onField`, `onFragment`, and `onOperation` in favor of `locations` which is a list of `__DirectiveLocation`.

In addition, this changes the specification of the `@skip` and `@include` directives to no longer allow their use on fragment directives (but are still allowed on fragment spreads and inline fragments). Because of this, a section on "merging fragment directives" is no longer useful.

**Rationale:**

This allows for a more fine-grained validation of directive placement, now you can assert that a directive is allowed on queries but not mutations, or allowed on fragment definitions but not on fragment spreads.

Also, this makes expanding the locations a directive is allowed to be placed easier to do, as future expansions will not affect the introspection API. This should be considered a prereq to supporting directives in more locations.

Finally, this is a prereq to a forthcoming RFC to add directives to the type schema language, one of the last missing pieces to represent a full schema using this language.

**Drawbacks:**

Any change to the introspection API is a challenge. Tools like Graph*i*QL should continue to support the older representation if possible.


andimarek:
@leebyron we just removed these fields from graphql java: 3,5 years after this PR ;-)

PR #153, opened 2016-03-19
[RFC] Ordering of queried fields

leebyron:
This removes language referring to responses as unordered. In fact responses should always be ordered in the order in which they were queried.

This adds clarifying language around both object types and input objects about ordering and provides more examples.

I've also added a caveat that response formats which don't support ordering are not required to uphold it.

**Impact:**

Some servers may return responses in an unordered fashion due to the way their executor behaves, though most which are ported from or inspired by graphql-js, our reference implementation, already behave this way.

**Motivation:**

This enables custom parsers which expect JSON values to be a very specific shape for performance reasons, but it also ensures an easier level of understandability when a response's shape so tightly matches the query - both for new learners and for debugging.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -779,6 +779,25 @@ curly-braces `{ }`.  The values of an object literal may be any input value
 literal or variable (ex. `{ name: "Hello world", score: 1.0 }`). We refer to
 literal representation of input objects as "object literals."
 
+**Input object fields are unordered**
+
+Input object fields may be provided in any syntactic order and maintain
+identical semantic meaning.
+
+These two queries are semantically identical:
+
+```graphql
+{
+  nearestThing(location: { lon: 12.43, lat: -53.211 })
+}
+```
+
+```graphql
+{
+  nearestThing(location: { lat: -53.211, lon: 12.43 })
+}
+```
+
 **Semantics**
 
 ObjectValue : { }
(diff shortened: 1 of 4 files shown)


spec/Section 2 -- Language.md
```diff
@@ -779,6 +779,25 @@ curly-braces `{ }`.  The values of an object literal may be any input value
 literal or variable (ex. `{ name: "Hello world", score: 1.0 }`). We refer to
 literal representation of input objects as "object literals."
 
+**Input object fields are unordered**
```
leebyron: This is just added clarification - this section matches a similar section for Field Arguments.

jjergus:
There's an additional edge case that needs to be addressed:

```
{
  foo @skip(if: true)
  bar
  foo
}
```

According to the text from this RFC the fields should be ordered as `foo,bar`, but using the implementation of `collectFields` from the specification, the fields would be ordered as `bar,foo`.

I think we should stick to the current implementation of `collectFields` and update this RFC to say something like:

> If a field is queried multiple times in a selection, it is ordered by the first time it is encountered. However fragments for which the type does not apply does not affect ordering. **Similarly, instances which are omitted due to `@skip` or `@include` directives, do not affect ordering.**

leebyron:
goodcatch will update.

leebyron:
Added language about the excluding directives, but also rewrote a bit to clarify that it's the order encountered *during execution* and specifically made reference to the CollectFields algorithm below.

PR #156, opened 2016-03-22
Fix formal spec of lone anonymous operation

frankleonrose:
Lone means lone. Set of ALL operations must be set of one.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -142,7 +142,7 @@ mutation dogOperation {
 
 ** Formal Specification **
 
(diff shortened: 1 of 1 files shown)


leebyron:
Thanks for highlighting this! I'm not sure your proposed change is correct though. It seems to read that only one operation of any kind may exist within a document, but this rule seeks to only ensure that is the case if the one query is *anonymous*.

frankleonrose:
Good point. I was thinking of the spec in isolation rather than globally.

leebyron:
Awesome, thanks!

PR #158, opened 2016-03-23
Clarify rules for validation + execution

leebyron:
There is some ambiguity of what it means to execute a query relative to if it has passed validation. Is it safe to ignore validation errors and execute anyhow? (no) Is it safe to skip validation altogether? (no) What about requests that at some point in the past were validated, do you need to validate them again every time you execute them? (only one validation required per request)

This adds more language around validation and its relationship to execution to help clarify these ambiguities and talks more about what happens when time passes between validation and execution.

There is some repetition here, just to reinforce these ideas between sections.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1,19 +1,36 @@
 # Validation
 
-GraphQL does not just verify if a request is syntactically correct.
+GraphQL does not just verify if a request is syntactically correct, but also
+ensures that it is unambiguous and mistake-free in the context of a given
+GraphQL schema.
 
-Prior to execution, it can also verify that a request is valid
-within the context of a given GraphQL schema. Validation is primarily
-targeted at development-time tooling. Any client-side tooling
-should return errors and not allow the formulation of queries
-known to violate the type system at a given point in time.
+An invalid request is still technically executable, and will always produce a
+stable result as defined by the procedures in the Execution section, however
+that result may be ambiguous, surprising, or unexpected relative to the request
+containing validation errors, so execution should not occur for invalid requests.
 
-Total request validation on the server-side during execution is optional. As
-schemas and systems change over time existing clients may end up emitting
-queries that are no longer valid given the current type system.  Servers
-(as described in the Execution section of this spec) attempt to satisfy as
-much of the request as possible and continue to execute in the presence
-of type system errors rather than cease execution completely.
+Typically validation is performed in the context of a request immediately
+before execution, however a GraphQL service may execute a request without
+explicitly validating it if that exact same request is known to have been
+validated before. For example: the request may be validated during development,
+provided it does not later change, or a service may validate a request once and
+memoize the result to avoid validating the same request again in the future.
+Any client-side or development-time tool should report validation errors and not
+allow the formulation or execution of requests known to be invalid at that given
+point in time.
+
+**Type system evolution**
+
+As GraphQL type system schema evolve over time by adding new types and new
+fields, it is possible that a request which was previously valid could later
+become invalid. Any change that can cause a previously valid request to become
+invalid is considered a *breaking change*. GraphQL services and schema
+maintainers are encouraged to avoid breaking changes, however in order to be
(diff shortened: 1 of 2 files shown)


spec/Section 5 -- Validation.md
```diff
@@ -1,19 +1,35 @@
...
-much of the request as possible and continue to execute in the presence
-of type system errors rather than cease execution completely.
+GraphQL does not just verify if a request is syntactically correct, but also
+ensures that it is unambiguous and mistake-free in the context of a given
+GraphQL schema.
+
+An invalid request is still technically executable, and will always produce a
+stable result as defined by the procedures in the Execution section, however
+that result may be ambiguous, surprising, or unexpected relative to the request
+containing invalid mistakes, so execution should not occur for invalid requests.
```
dschafer: "invalid mistakes" reads slightly weird to me. Are the mistakes invalid, or is the request invalid because it has mistakes.

jjergus: remove "invalid" from "invalid mistakes"

spec/Section 5 -- Validation.md
```diff
@@ -1,19 +1,35 @@
...
+
+Typically validation is performed in the context of a request immediately
+before execution, however a GraphQL service may execute a request without
+explicitly validating it if that exact same request is known to have been
+validated before. For example: a client may ensure the request is valid
+before sending it, or if the request may be validated during development,
```
dschafer: the "if" seems unneeded.

spec/Section 6 -- Execution.md
```diff
@@ -16,6 +16,24 @@ should be the result of evaluating the operation according to the “Evaluating
...
+without execution.
+
+Typically validation is performed in the context of a request immediately
+before execution, however a GraphQL service may execute a request without
+explicitly validating it if that exact same request is known to have been
+validated before. For example: a client may ensure the request is valid
+before sending it, or if the request may be validated during development,
+provided it does not later change.
+
+A GraphQL service may only execute requests which *at some point* were
```
dschafer: This is an odd use of "may"; I'd suggest this be "should" instead.

spec/Section 5 -- Validation.md
```diff
@@ -1,19 +1,35 @@
...
+
+An invalid request is still technically executable, and will always produce a
+stable result as defined by the procedures in the Execution section, however
+that result may be ambiguous, surprising, or unexpected relative to the request
+containing invalid mistakes, so execution should not occur for invalid requests.
+
+Typically validation is performed in the context of a request immediately
+before execution, however a GraphQL service may execute a request without
+explicitly validating it if that exact same request is known to have been
+validated before. For example: a client may ensure the request is valid
```
jjergus: I understand this is just an example, and it's technically not wrong, I still don't think it's a good idea to encourage people to implement servers that rely on clients for validation.

Maybe a better example would be something like

> a server may validate a query once and memoize this information, then if it receives the same query, not validate it again

leebyron:
Great feedback, you guys rock

leebyron:
Updated with feedback

dschafer:
Ship it.

spec/Section 5 -- Validation.md
```diff
@@ -1,19 +1,36 @@
...
-much of the request as possible and continue to execute in the presence
-of type system errors rather than cease execution completely.
+GraphQL does not just verify if a request is syntactically correct, but also
+ensures that it is unambiguous and mistake-free in the context of a given
+GraphQL schema.
+
+An invalid request is still technically executable, and will always produce a
+stable result as defined by the procedures in the Execution section, however
+that result may be ambiguous, surprising, or unexpected relative to the request
+containing validation errors, so execution should not occur for invalid requests.
```
asiandrummer: This paragraph is a tiny bit confusing to me. I think it's saying an invalid request, although still technically executable, should not be executed because the result may be unexpected?

Feel free to ignore if you think it's my misunderstanding ;) (which is completely possible).

PR #159, opened 2016-03-30
Union type definition must be lower case

helfer:
Minor typo in union type defition. Capitalization is significant, so `Union` has to be changed to `union`, otherwise the parser will complain.


leebyron:
Thank you :)

PR #172, opened 2016-05-07
Fixing about 15 typos and grammatical errors, with another 4 lines be…

danielkwinsor:
…ing italicized.


leebyron:
Nice work, thank you for the careful read!

PR #173, opened 2016-05-07
Unifying name of Non-Null

danielkwinsor:
For better searchability, use 1 name for Non-Null.
Ignoring case, there are 12 occurrences of the phrase "Non-Null", 5 of "Non Null" and 1 of "not-null".  Side note, my chrome browser doesn't even find not-null in a search.

I have not made changes to case, as many searches are not case sensitive.


leebyron:
consistency FTW!

PR #177, opened 2016-05-12
CompleteValue function for lists and non-null need subSelectionSet

danielkwinsor:
This will add a bit more clarity that when you have a list of list of list of object, you don't need to specifically address all the lists, but just write the field selection to the object and GraphQL will traverse the lists as necessary.


leebyron:
Good catch, thanks!

PR #191, opened 2016-07-01
Loosen must to should for serialization supporting ordered maps.

### LINKED ISSUE #168: JSON Key Order
Hey folks,

Per https://github.com/facebook/graphql/commit/d4b4e67f083ff6699a8f5fac5e069faba89c5bbb it seems to imply that GraphQL implementations must guarantee a particular key order in JSON responses sent back from GraphQL servers.

I'm struggling to figure out how this isn't plainly at odds with the JSON spec which states: http://www.json.org/

```
An object is an unordered set of name/value pairs. 
```

leebyron:
GraphQL should work correctly with many serialization techniques, not just JSON. Even within JSON, the GraphQL spec mentions ordered maps while the JSON spec describes unordered maps. This changes the language in the GraphQL spec to:

* Use should instead of must where ordered map support might not be possible, thus supporting more environments.

* Add clarifying language to JSON encoding section describing expectations for ordering.

Fixes #168

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -241,7 +241,7 @@ While Scalar types describe the leaf values of these hierarchical queries, Objec
 describe the intermediate levels.
 
 GraphQL Objects represent a list of named fields, each of which yield a value of
-a specific type. Object values are serialized as ordered maps, where the
+a specific type. Object values should be serialized as ordered maps, where the
 queried field names (or aliases) are the keys and the result of evaluating
 the field is the value, ordered by the order in which they appear in the query.
 
@@ -261,8 +261,9 @@ that will yield an `Int` value, and `picture` a field that will yield a
 
 A query of an object value must select at least one field. This selection of
 fields will yield an ordered map containing exactly the subset of the object
-queried, in the order in which they were queried. Only fields that are declared
-on the object type may validly be queried on that object.
+queried, which should be represented in the order in which they were queried.
+Only fields that are declared on the object type may validly be queried on
+that object.
 
 For example, selecting all the fields of `Person`:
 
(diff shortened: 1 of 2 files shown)


bruce:
I added some commit comments to 5eaa640 with some suggestions.

leebyron:
Updated with your feedback

spec/Section 7 -- Response.md
```diff
@@ -53,6 +52,24 @@ the following JSON concepts:
...
 | Enum Value    | String            |
 
+**Object Property Ordering**
+
+While JSON Objects are specified as an
+[unordered collection of key-value pairs](https://tools.ietf.org/html/rfc7159#section-4)
+the pairs are represented in an ordered manner. In other words, while the JSON
+strings `{ "name": "Mark", "age": 30 }` and `{ "age": 30, "name": "Mark" }`
+encode the same value, they also have observably different property orderings.
+While not every JSON generator allows for control over the order in which
```
dschafer: I might remove this bit, and do something like:

  From the above, the spec notes that "Serialization formats which can represent an ordered map should preserve the order of requested fields as defined by query execution." Hence, if the query was `{ name, age }`, a GraphQL server vending JSON should respond with `{ "name": "Mark", "age": 30 }`, preserving the order of the requested field in the JSON serialization.

spec/Section 3 -- Type System.md
```diff
@@ -357,9 +358,15 @@ excluding fragments for which the type does not apply and fields or
 fragments that are skipped via `@skip` or `@include` directives. This ordering
 is correctly produced when using the {CollectFields()} algorithm.
 
-Response formats which support ordered maps (such as JSON) must maintain this
-ordering. Response formats which do not support ordered maps may disregard
-this ordering.
+Response formats and runtime environments capable of representing ordered maps
+should maintain this ordering when possible. Serialization formats which can
```
dschafer: I might remove "when possible" here; if it's capable of representing ordered maps, then it is possible!

spec/Section 3 -- Type System.md
```diff
@@ -357,9 +358,15 @@ excluding fragments for which the type does not apply and fields or
...
+only represent unordered maps should if possible retain this order
```
dschafer: same, would remove "if possible"

leebyron:
Simplified a bit, including @dschafer's feedback

PR #217, opened 2016-10-03
Clarify the names regular expression

### LINKED ISSUE #216: Section 2.1.9 Names RegExp ambiguous
It looks as though the regular expression used to define valid names within GraphQL (https://facebook.github.io/graphql/#sec-Names) is incorrect. In particular it doesn't include anchors or in the text specify that the entire name must match the given regular expression. I would propose that the specification be updated to `/^[_A-Za-z][_0-9A-Za-z]*$/`.

NeoPhi:
clarify that the regular expression applies to the entire name, fixes #216

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -150,7 +150,7 @@ lacks the punctuation often used to describe mathematical expressions.
 
 ### Names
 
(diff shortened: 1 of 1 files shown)


NeoPhi:
Based on the discussion in #216 this change isn't needed.

PR #219, opened 2016-10-14
Correct an error in the Punctuator grammar

chris-morgan:
In Appendix B `Punctuator` includes `|`, but it was missing here.


leebyron:
Thanks!

PR #221, opened 2016-10-21
Clarify execution section.

leebyron:
This adds algorithms to the section on execution and reorders content to better follow the flow of execution.

Note that no additional semantics are being introduced in this PR. This is simply algorithmic clarification of the execution process.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -775,13 +775,26 @@ An input object is never a valid result.
 
 **Input Coercion**
 
-The input to an input object should be an unordered map, otherwise an error
-should be thrown. The result of the coercion is an unordered map, with an
-entry for each input field, whose key is the name of the input field.
-The value of an entry in the coerced map is the result of input coercing the
-value of the entry in the input with the same key; if the input does not have a
-corresponding entry, the value is the result of coercing null. The input
-coercion above should be performed according to the input coercion rules of the
+The value for an input object should be an input object literal or an unordered
+map, otherwise an error should be thrown. This unordered map should not contain
+any entries with names not defined by a field of this input object type,
+otherwise an error should be thrown.
+
+If any non-nullable fields defined by the input object do not have corresponding
+entries in the original value, were provided a variable for which a value was
+not provided, or for which the value {null} was provided, an error should
+be thrown.
+
+The result of coercion is an environment-specific unordered map defining slots
+for each field of the input object type.
+
+For each field of the input object type, if the original value has an entry with
+the same name, and the value at that entry is a literal value or a variable
+which was provided a runtime value, an entry is added to the result with the
+name of the field.
+
+The value of that entry in the result is the outcome of input coercing the
+original entry value according to the input coercion rules of the
 type declared by the input field.
 
 #### Input Object type validation
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -786,7 +786,8 @@ not provided, or for which the value {null} was provided, an error should
 be thrown.
 
 The result of coercion is an environment-specific unordered map defining slots
-for each field of the input object type.
+for each field both defined by the input object type and provided by the
+original value.
 
 For each field of the input object type, if the original value has an entry with
 the same name, and the value at that entry is a literal value or a variable
@@ -796,6 +797,27 @@ name of the field.
 The value of that entry in the result is the outcome of input coercing the
 original entry value according to the input coercion rules of the
 type declared by the input field.
+
+Following are examples of Input Object coercion for the type:
+
+```graphql
+input ExampleInputObject {
+  a: String
+  b: Int!
+}
+```
+
+Original Value          | Variables       | Coerced Value
+-------------------------------------------------------------------------------
+`{ a: "abc", b: 123 }`  | `{}`            | `{ a: "abc", b: 123 }`
+`{ a: 123, b: "123" }`  | `{}`            | `{ a: "123", b: 123 }`
+`{ a: "abc" }`          | `{}`            | Error: Missing required field {b}
+`{ b: $var }`           | `{ var: 123 }`  | `{ b: 123 }`
+`{ b: $var }`           | `{ var: null }` | Error: {b} must be non-null.
+`{ b: $var }`           | `{}`            | Error: {b} must be non-null.
+`{ b: $var }`           | `{}`            | Error: {b} must be non-null.
+`{ a: $var, b: 1 }`     | `{ var: null }` | `{ a: null, b: 1 }`
+`{ a: $var, b: 1 }`     | `{}`            | `{ b: 1 }`
 
 #### Input Object type validation
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -81,19 +81,23 @@ CoerceVariableValues(schema, operation, variableValues):
   * For each {variableDefinition} in {variableDefinitions}:
     * Let {variableName} be the name of {variableDefinition}.
     * Let {variableType} be the expected type of {variableDefinition}.
-    * If no value was provided in {variableValues} for the name {variableName}:
-      * If {variableType} is a Non-Nullable type, throw a query error.
-      * Otherwise, continue to the next variable definition.
+    * Let {defaultValue} be the default value for {variableDefinition}.
     * Let {value} be the value provided in {variableValues} for the name {variableName}.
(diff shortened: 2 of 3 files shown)


spec/Section 3 -- Type System.md
```diff
@@ -775,13 +775,26 @@ An input object is never a valid result.
...
-coercion above should be performed according to the input coercion rules of the
+The value for an input object should be an input object literal or an unordered
+map, otherwise an error should be thrown. This unordered map should not contain
+any entries with names not defined by a field of this input object type,
+otherwise an error should be thrown.
+
+If any non-nullable fields defined by the input object do not have corresponding
+entries in the original value, were provided a variable for which a value was
+not provided, or for which the value {null} was provided, an error should
+be thrown.
```
jjergus: With the introduction of a null literal, do we need a separate concept of "non-nullable" and "required"? For example, should it be possible to declare an input object field which must be provided, but its value may be null?

    { required_field: 42 } # valid
    { required_field: null } # valid
    { } # invalid
    { required_field: $var } # valid if $var is required but $var may be nullable...I guess?

spec/Section 3 -- Type System.md
```diff
@@ -775,13 +775,26 @@ An input object is never a valid result.
...
+
+The result of coercion is an environment-specific unordered map defining slots
+for each field of the input object type.
```
jjergus: "Each field of the input object type" or "each field included in the value before coercion"?

If the former, how do we distinguish between fields that weren't provided and fields that had explicit null value?

spec/Section 3 -- Type System.md
```diff
@@ -775,13 +775,26 @@ An input object is never a valid result.
...
+
+For each field of the input object type, if the original value has an entry with
+the same name, and the value at that entry is a literal value or a variable
+which was provided a runtime value, an entry is added to the result with the
+name of the field.
```
jjergus: If I understand it correctly

  { optional_field: $var }

coerces to

  { optional_field: 42 } # if $var == 42
  { optional_field: null } # if $var is explicitly null
  { } # if $var was not provided

I guess that makes sense but we should probably have some examples here.

spec/Section 6 -- Execution.md
```diff
@@ -42,44 +73,106 @@ of variable's declared type. If a query error is encountered during
...
-If the operation is a query, the result of the operation is the result of
-evaluating the query’s top level selection set on the “Query Root” object.
+It is expected that the top level fields in a mutation operation perform
+side-effects on the underlying data system. Serial execution of the provided
+mutations ensures against race conditions during these side-effects.
+
+ExecuteMutation(mutation, schema, variableValues, initialValue):
+
+  * Let {variableValues} be the set of variable values to be used by any
+    field argument value coercion.
```
jjergus: This step seems redundant, doesn't say anything useful (and we don't have it in the "query" algorithm above).

spec/Section 6 -- Execution.md
```diff
@@ -42,44 +73,106 @@ of variable's declared type. If a query error is encountered during
...
+  * Let {mutationType} be the root Mutation type in {schema}.
+  * Assert: {mutationType} is an Object type.
```
jjergus: Same as above (schema validation).

spec/Section 3 -- Type System.md
leebyron: I'll make sure to specifically address this point in the null literal proposal. 

I think it's actually not worth separating these notions. I think there probably isn't a ton of value beyond fairly pedantic specificity in the cases of required but nullable and optional but non-nullable.

spec/Section 3 -- Type System.md
leebyron: I'll clarify this for this section, but leave the null value handling to the null literal proposal.

spec/Section 3 -- Type System.md
leebyron: That's the current behavior at least :/ - I'll definitely include examples, that's a great idea.

If the behavior seems wrong we can discuss proposing to change it.

spec/Section 6 -- Execution.md
leebyron: goodcatch - this was a mistake :)

spec/Section 6 -- Execution.md
```diff
@@ -42,44 +73,106 @@ of variable's declared type. If a query error is encountered during
...
+ExecuteSelectionSet(selectionSet, objectType, objectValue, variableValues):
 
-If the selection set is being evaluated on the `null` object, then the result
-of evaluating the selection set is `null`.
+  * Initialize {visitedFragments} to be the empty set.
+  * Let {groupedFieldSet} be the result of
+    {CollectFields(objectType, selectionSet, visitedFragments, variableValues)}.
+  * Initialize {resultMap} to an empty ordered map.
+  * For each {groupedFields} in {groupedFieldSet}:
+    * Let {entryTuple} be {GetFieldEntry(objectType, objectValue, groupedFields, variableValues)}.
```
leebyron: The tuple approach was decided when @dschafer and I were writing this section last year. Namely we need to distinguish between a field value of null vs a field not being included at all. Otherwise as you're pointing out, a larger part of the `GetFieldEntry` would end up in this algorithm instead. It aligns to the reference implementation (which in JS returns either `null` as a value or `undefined` as a skip).

I like this suggestion though - I'd like to explore this as a follow up since it might also suggest changing the reference implementation as well.

spec/Section 6 -- Execution.md
```diff
@@ -42,44 +73,106 @@ of variable's declared type. If a query error is encountered during
...
+
+This ensures all fields with the same response key (alias or field name)
+included via referenced fragments are executed at the same time.
+
+CollectFields(objectType, selectionSet, visitedFragments, variableValues):
 
   * Initialize {groupedFields} to an empty ordered list of lists.
-  * For each {selection} in {selectionSet};
+  * For each {selection} in {selectionSet}:
     * If {selection} provides the directive `@skip`, let {skipDirective} be that directive.
```
leebyron: This actually isn't already a validation rule and is valid, though it does weird things. Changing this is probably not appropriate in an editing PR, but we should definitely fix this. 

Filed #223

spec/Section 6 -- Execution.md
```diff
@@ -350,3 +485,39 @@ cannot be `null` the error is propagated to be dealt with by the parent field.
...
+
+CoerceArgumentValues(objectType, field, variableValues)
+  * Let {argumentValues} be the argument values provided in {field}.
+  * Let {fieldName} be the name of {field}.
+  * Let {argumentDefinitions} be the arguments defined by {objectType} for the
+    field named {fieldName}.
+  * Let {coercedArgumentValues} be an empty Map.
+  * For each {argumentDefinitions} as {argumentName} and {argumentType}:
+    * If no value was provided in {argumentValues} for the name {argumentName}:
+      * Continue to the next argument definition.
```
leebyron: That's not how things work today, but I agree that's problematic. Filed #224 to follow up with this.

spec/Section 6 -- Execution.md
tendant: Can we have a little bit more explanation or example regarding {responseKey: groupedFields}, in what case a responseKey might have grouped fields? In such case how to determine the result of responseKey.

leebyron:
Added some additional changes.

Here's what these changes add up to for easier review https://ns-dtlgzkzepm.now.sh/#sec-Execution

spec/Section 6 -- Execution.md
```diff
@@ -135,110 +152,30 @@ response map.
...
-    {CollectFields(objectType, selectionSet, visitedFragments, variableValues)}.
+    {CollectFields(objectType, selectionSet, variableValues)}.
   * Initialize {resultMap} to an empty ordered map.
-  * For each {groupedFields} in {groupedFieldSet}:
-    * Let {entryTuple} be {GetFieldEntry(objectType, objectValue, groupedFields, variableValues)}.
-    * If {entryTuple} is not {null}:
-      * Let {responseKey} and {responseValue} be the values of {entryTuple}.
-      * Set {responseValue} as the value for {responseKey} in {resultMap}.
+  * For each {groupedFieldSet} as {responseKey} and {fields}:
+    * Let {fieldName} be the name of the first entry in {fields}.
```
jjergus: I like the phrasing you used for a similar case in CollectFields better:

    Let {fieldName} be the name shared by all fields in {fields}.

I think it makes it more intuitive what's going on.

Another option would be to add a note explaining that all {fields} have the same name.

PR #226, opened 2016-10-26
[RFC] No ignored within Variables.

leebyron:
This adds a specific limitation to the Variable parsing rules to make it impossible to put ignored characters between the `$` and the name of the variable.

This was suggested by @martijnwalraven as he was building syntax highlighters for some IDEs.

I'm not sure if this is necessary or good - but wanted to capture it so we can have a conversation about it.

To be specific about the issue and the impact of this change, currently the following are legal variables:

``` graphql
$variable

$ variable

$
variable

$       variable

$,,,variable
```

With this change, only `$variable` will be a parsable Variable form, the others with ignored characters will not be allowed.

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -152,7 +152,7 @@ VariableDefinitions : ( VariableDefinition+ )
 
 VariableDefinition : Variable : Type DefaultValue?
 
-Variable : $ Name
+Variable : $ [lookahead ! Ignored] Name
 
 DefaultValue : = Value[Const]
 
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -855,7 +855,7 @@ ObjectValue : { ObjectField+ }
 
 ## Variables
 
-Variable : $ Name
+Variable : $ [lookahead ! Ignored] Name
 
 VariableDefinitions : ( VariableDefinition+ )
 


martijnwalraven:
:) Thanks! Not sure if it is worth changing this. It would complicate the grammar and force all implementations to make changes to their parser to fix something that probably isn't a real issue for anyone. Plus I've actually grown fond of this little quirk. If we ever release a TCK, I suggest we include a rule that makes sure a spec compliant implementation allows these forms.

leebyron:
This is also not the only case where punctuation comes before a Name and we would expect no spaces. Another example is `...Fragment` That's also possible to write as `... Fragment` or `...,,,Fragment`.

Perhaps the right fix for this is a linter with a lint rule that looks for this sort of thing?

aweiker:
Without this being in the spec, then a linter is just a perspective and tools would need to support whatever the spec said, otherwise it would be a bug. So while a linter would help a query writer, it does not help those writing the tools.

stubailo:
Interesting, I would expect that `... fragmentName` should be allowed. I can imagine a style preference where you would write `... myFragment` to match `... on MyType`.

aweiker:
I was focused on the variable use case (title of the issue).

For the fragment style, having a space does allow for easier readability. So making this scoped to variable is perfect.

stubailo:
Looks like a great change to me, perhaps it could be merged? Or we could make a branch for the next version of the spec (April, I suppose?) and merge it there?

stubailo:
This might be controversial, but after a few more months of experience with GraphQL, I still believe this is a good change.

stubailo:
@wincent any reason this was closed?

wincent:
Ah sorry, just some overzealous spring cleaning, as it has been sitting mostly untouched for the better part of a year (save for your comment, which I didn't even see). Happy to revive this if there is still interest.

stubailo:
No worries! I agree it's not a high priority item :)

PR #229, opened 2016-10-28
[RFC] Add Validation rule for unique directives per location.

### LINKED ISSUE #223: @skip or @include should only be used once per selection
As mentioned in https://github.com/facebook/graphql/pull/221#discussion_r84519559

leebyron:
This rule was suggested by @jjergus (#223) to remove ambiguity from execution where the `@skip` and `@ignore` directive rules make the assumption that only one is defined per selection.

In general I think this assumption should be upheld by valid queries overall, not just for `@skip` and `@include` so that it can be relied upon for any usage of directives. For example, this allows a potential optimization to use a hashmap keyed on directive name to represent the directives at a location rather than an array.

Closes #223

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1242,6 +1242,49 @@ query @skip(if: $foo) {
 ```
 
 
+### Directives Are Unique Per Location
+
+** Formal Specification **
+
+  * For every {location} in the document for which Directives can apply:
+    * Let {directives} be the set of Directives which apply to {location}.
+    * For each {directive} in {directives}:
+      * Let {directiveName} be the name of {directive}.
+      * Let {namedDirectives} be the set of all Directives named {directiveName}
+        in {directives}.
+      * {namedDirectives} must be a set of one.
+
+** Explanatory Text **
+
+Directives are used to describe some metadata or behavioral change on the
+definition they apply to. When more than one directive of the same name is used,
+the expected metadata or behavior becomes ambiguous, therefore only one of each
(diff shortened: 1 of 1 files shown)


stubailo:
I guess this means that certain directives will need to include array arguments, rather than having one argument and using the directive multiple times, eg:

```
{
  comments @refetchOn(channels: ["newComment", "removedComment"])
}
```

Rather than:

```
{
  comments @refetchOn(channel: "newComment") @refetchOn(channel: "removedComment")
}
```

Seems reasonable, to require the directive to decide if it wants to accept multiple inputs.

leebyron:
Exactly. I think the latter API is more explicit to the possible behavior.

spec/Section 5 -- Validation.md
```diff
@@ -1242,6 +1242,45 @@ query @skip(if: $foo) {
...
+
+However this example is valid because `@skip` has been used only once per
+location, despite being used twice in the query:
+
+```graphql
+query ($foo: Boolean = true, $bar: Boolean = false) {
+  fieldA @skip(if: $foo)
+  fieldB @skip(if: $bar)
+}
+```
```
leebyron: Great point - I should add that

leebyron:
Further rationale after an offline discussion with @dschafer:

While this is explicitly a limiting change and therefore potentially breaking, by reducing what is possible now it *increases* future option value. This change is one that we could always change back in some capacity in the future in a non-breaking way should an explicit need arise. Barring any such need today, it's in our best interest to preserve that option-value.

Issue #214, opened 2016-09-29
Allow unicode characters outside the Basic Multilingual Plane


dylanahsmith:
Currently a GraphQL document is only allows a [SourceCharacter :: `/[\u0009\u000A\u000D\u0020-\uFFFF]/`](http://facebook.github.io/graphql/#sec-Source-Text) and [EscapedUnicode :: `/[0-9A-Fa-f]{4}/`](http://facebook.github.io/graphql/#EscapedUnicode) also prevents unicode characters above U+FFFF from being included into a GraphQL string.

Unicode code points are actually in the range 0 to 0x10FFFF.  For example, [unicode emoji characters](http://unicode.org/emoji/charts/full-emoji-list.html) like 😀  (U+1F600) have code points above U+FFFF.

Is there any reason why the source document doesn't allow unicode characters above U+FFFF?  Or can we remove that restriction?  Without that restriction the limitation of the unicode escape doesn't seem problematic.

If supporting a unicode escape for all unicode characters is desired, then one way of handling that is the way [swift supports unicode escapes](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/StringsAndCharacters.html):

> An arbitrary Unicode scalar, written as \u{n}, where n is a 1–8 digit hexadecimal number with a value equal to a valid Unicode code point

chris-morgan:
I also was reading the spec and realised this. Given the paragraph around it:

> GraphQL documents are expressed as a sequence of Unicode characters. However, with few exceptions, most of GraphQL is expressed only in the original non‐control ASCII range so as to be as widely compatible with as many existing tools, languages, and serialization formats as possible and avoid display issues in text editors and source control.

It sounds to me an error of ignorance rather than intent.

Rust formerly had `\uXXXX` and `\UXXXXXXXX`, but changed to `\u{x}` (the same as Swift does) some time before Rust 1.0.

JSON (which is probably the main inspiration for the GraphQL syntax) does `\uXXXX` and uses the abomination that is [UTF-16 surrogate pairs](https://en.wikipedia.org/wiki/UTF-16#U.2B10000_to_U.2B10FFFF) as a way of representing higher-order characters, e.g. U+1F600 (😀) is escaped as `"\ud83d\ude00"` in JSON.

[...]

PR #231, opened 2016-10-28
[RFC] Support full Unicode character range

leebyron:
This proposes support for the entire unicode character range in a source document, rather than explicitly excluding certain characters.

Because this doesn't add any new characters to the ignored sequences, in practice this allows for adding more characters to comments (low value) and strings (high value).

This allows strings to include literal unicode points as a way to represent supplemental plane characters. In fact, it almost allows ignoring escape sequences altogether as all characters can be represented literally.

Suggested by #214

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -1,6 +1,6 @@
 # B. Appendix: Grammar Summary
 
-SourceCharacter :: /[\u0009\u000A\u000D\u0020-\uFFFF]/
+SourceCharacter :: "Any Unicode code point"
 
 
 ## Ignored Tokens
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -13,27 +13,43 @@ double-colon `::`).
 
 ## Source Text
 
-SourceCharacter :: /[\u0009\u000A\u000D\u0020-\uFFFF]/
+SourceCharacter :: "Any Unicode code point"
 
 GraphQL documents are expressed as a sequence of
-[Unicode](http://unicode.org/standard/standard.html) characters. However, with
-few exceptions, most of GraphQL is expressed only in the original non-control
-ASCII range so as to be as widely compatible with as many existing tools,
-languages, and serialization formats as possible and avoid display issues in
-text editors and source control.
+[Unicode](http://unicode.org/standard/standard.html) code points (referred to in
+this specification as characters). All Unicode code point values from U+0000 to
+U+10FFFF, including surrogate code points, may appear in this sequence where
+allowed by the grammatical rules below.
+
+A [combining character sequence](http://unicode.org/faq/char_combmark.html) is
+treated as a sequence of individual Unicode code points and a sequence of
+individual {SourceCharacter}, even though they may appear to a user as a
+single character.
+
+However, with the exceptions of {StringValue} and {Comment}, most of GraphQL is
+expressed only in the original non-control ASCII range so as to be as widely
+compatible with as many existing tools, languages, and serialization formats as
+possible and avoid display issues in text editors and source control.
+
+Note: The encoding used to represent a GraphQL document source is irrelevant to
+this specification. A document is not required to be stored or transmitted in an
+encoding which can represent every Unicode code point. Instead, given any
+encoding format, and the range of code points which it can encode, GraphQL
+documents may consist of any of those code points.
 
 
-### Unicode
+### Byte Order Mark
 
 UnicodeBOM :: "Byte Order Mark (U+FEFF)"
-
-Non-ASCII Unicode characters may freely appear within {StringValue} and
-{Comment} portions of GraphQL.
 
 The "Byte Order Mark" is a special Unicode character which
 may appear at the beginning of a file containing Unicode which programs may use
 to determine the fact that the text stream is Unicode, what endianness the text
 stream is in, and which of several Unicode encodings to interpret.
+
+GraphQL ignores this character anywhere ignored tokens may occur, regardless of
+if it appears at the beginning of a GraphQL document, as it may appear within a
+document due to file concatenation.
 
 
 ### White Space
@@ -65,7 +81,11 @@ text, any amount may appear before or after any other token and have no
 significance to the semantic meaning of a GraphQL query document. Line
 terminators are not found within any other token.
 
-Note: Any error reporting which provide the line number in the source of the
+Note: GraphQL intentionally does not consider Unicode line or paragraph
+separators outside the ASCII range as line terminators, avoiding
+misinterpretation by text editors and source control tools.
+
+Note: Any error reporting which provides the line number in the source of the
 offending syntax should use the preceding amount of {LineTerminator} to produce
 the line number.
 
@@ -83,9 +103,13 @@ A comment can contain any Unicode code point except {LineTerminator} so a
 comment always consists of all code points starting with the {`#`} character up
 to but not including the line terminator.
 
-Comments behave like white space and may appear after any token, or before a
+Comments behave like white space and may appear after any token, or before any
 line terminator, and have no significance to the semantic meaning of a GraphQL
 query document.
+
+Any Unicode code point may appear within a Comment. Comments do not include
+escape sequences, so the character sequence `\n` or `\u000A` must not be
+interpreted as the end of a Comment.
 
 
 ### Insignificant Commas
@@ -704,13 +728,13 @@ EscapedUnicode :: /[0-9A-Fa-f]{4}/
 
 EscapedCharacter :: one of `"` \ `/` b f n r t
 
-Strings are sequences of characters wrapped in double-quotes (`"`). (ex.
-`"Hello World"`). White space and other otherwise-ignored characters are
-significant within a string value.
+Strings are sequences of zero or more source characters wrapped in double-quotes
+(`"`). (ex. `"Hello World"`).
 
-Note: Unicode characters are allowed within String value literals, however
-GraphQL source must not contain some ASCII control characters so escape
-sequences must be used to represent these characters.
+Any Unicode code point other than those explicitly excluded may appear literally
+within a String value. White-space and other characters otherwise ignored
+outside of string values are significant and included. Unicode code points may
+also be represented with escape sequences.
 
 **Semantics**
 


leebyron:
Drawbacks and concerns I'd like feedback on from folks with more experience dealing with unicode corner cases:

By encouraging use of non-ascii characters in strings, we will inevitably see utf8 files that include supplemental plane and surrogate pair points. Will this break any common tools or create an unnecessary burden?

This removes the restriction of excluding control flow characters from the source. This is in the name of simplicity, however it may have real drawbacks for interaction with other tools. For example, could sources that contain U+0000 be interpreted as the end of a file? If this remains a real concern, it might be worth adding back that restriction just as a point of practical convenience.

leebyron:
Check out a rendering of this patch applied here: https://ns-gxioxiaqyc.now.sh

spec/Appendix B -- Grammar Summary.md
```diff
@@ -1,6 +1,6 @@
 # B. Appendix: Grammar Summary
 
-SourceCharacter :: /[\u0009\u000A\u000D\u0020-\uFFFF]/
+SourceCharacter :: "Any Unicode code point from U+0000 to U+10FFFF"
```
chris-morgan: Viz. any Unicode code point.

spec/Section 2 -- Language.md
```diff
@@ -13,28 +13,39 @@ double-colon `::`).
...
-languages, and serialization formats as possible and avoid display issues in
-text editors and source control.
+[Unicode](http://unicode.org/standard/standard.html) code points (referred to in
+this specification as characters). All Unicode code point values from U+0000 to
+U+10FFFF, including surrogate code points, may appear in this sequence where
+allowed by the grammatical rules below.
+A [combining character sequence](http://unicode.org/faq/char_combmark.html) are
+treated as a sequence of individual Unicode code points and a sequence of
+individual {SourceCharacter}, even though they may appear to a user as a
+single character.
```
chris-morgan: And this introduces the potential to seriously mess with some syntax highlighters by combining characters with quotation marks and things like that. Ah well. Nothing new here, that’s how everything treats such things.

BTW, is there supposed to be a paragraph break before this sentence?

spec/Section 2 -- Language.md
```diff
@@ -13,28 +13,39 @@ double-colon `::`).
...
 GraphQL documents are expressed as a sequence of
-[Unicode](http://unicode.org/standard/standard.html) characters. However, with
-few exceptions, most of GraphQL is expressed only in the original non-control
-ASCII range so as to be as widely compatible with as many existing tools,
-languages, and serialization formats as possible and avoid display issues in
-text editors and source control.
+[Unicode](http://unicode.org/standard/standard.html) code points (referred to in
+this specification as characters). All Unicode code point values from U+0000 to
+U+10FFFF, including surrogate code points, may appear in this sequence where
+allowed by the grammatical rules below.
```
chris-morgan: “including surrogate code points”: I’m not sure what you mean by this. Do you mean that you will allow things like U+DEAD? This is problematic: UTF-8 does not allow surrogates, and a language that works with UTF-8 strings will then break. Rust, for example.

spec/Section 2 -- Language.md
```diff
@@ -13,28 +13,39 @@ double-colon `::`).
...
+UnicodeBOM :: "Byte Order Mark (U+FEFF)"
 
 The "Byte Order Mark" is a special Unicode character which
 may appear at the beginning of a file containing Unicode which programs may use
 to determine the fact that the text stream is Unicode, what endianness the text
 stream is in, and which of several Unicode encodings to interpret.
 
+GraphQL ignores this character regardless of if it appears at the beginning of a
+source, as it may appear within due to concatenation of sources.
+
```
chris-morgan: What about in the middle of a string or a token?

spec/Section 2 -- Language.md
```diff
@@ -83,10 +98,14 @@ A comment can contain any Unicode code point except {LineTerminator} so a
...
 to but not including the line terminator.
 
-Comments behave like white space and may appear after any token, or before a
+Comments behave like white space and may appear after any token, or before any
 line terminator, and have no significance to the semantic meaning of a GraphQL
 query document.
 
+Any Unicode code point may appear within a Comment. Comments do not include
+escape sequences, so the character sequence `\n` or `\000A` must not be
+interpretted as the end of a Comment.
```
chris-morgan: s/interpretted/interpreted/

spec/Section 2 -- Language.md
```diff
@@ -704,13 +723,13 @@ EscapedUnicode :: /[0-9A-Fa-f]{4}/
...
+Strings are sequences of zero or more source characters wrapped in double-quotes
+(`"`). (ex. `"Hello World"`).
 
-Note: Unicode characters are allowed within String value literals, however
-GraphQL source must not contain some ASCII control characters so escape
-sequences must be used to represent these characters.
+Any Unicode code point other than those explicitly excluded may appear literally
+within a String value. White-space and other characters otherwise ignored
+outside of string values are significant and included. Unicode code points may
+also be represented with escape sequences.
```
chris-morgan: … where escape sequences are? (Looks like `\u EscapedUnicode` still isn’t explained precisely or clearly?)

spec/Section 2 -- Language.md
leebyron: The hope is to both be as non-restrictive as possible while being compatible with as many unicode encodings as possible. There's a note below that the encoding should be irrelevant.

Any given unicode encoding can represent some form of a stream of units between U+0000 and U+10FFFF. The fact that UTF-8 cannot use UTF-16 surrogate pairs, or that UTF-8 can encode invalid points that cannot be encoded in UTF-16 should be irrelevant.

Explicitly, we don't want to make any promises that the GraphQL language will do some specific unicode-specific operations that would be an undesirable burden for implementors who need to use existing languages with existing Unicode quirks. Hence the clauses above and below that GraphQL won't combine surrogate pairs for you nor assemble combining sequences before tokenizing.

If there's a better way to word this that's more clear, I'm definitely open to suggestions.

spec/Section 2 -- Language.md
leebyron: Good suggestion to add one

spec/Section 2 -- Language.md
leebyron: I'll clarify for those cases

spec/Section 2 -- Language.md
leebyron: They're explained in the Semantics section right below here. I'm open to suggestions for improvements to that if you feel there's something specific that is not precise or clear enough

leebyron:
Updated to incorporate feedback

spec/Section 2 -- Language.md
chris-morgan: OK, I’m happy with that part now.

spec/Appendix B -- Grammar Summary.md
```diff
@@ -1,6 +1,6 @@
 # B. Appendix: Grammar Summary
 
-SourceCharacter :: /[\u0009\u000A\u000D\u0020-\uFFFF]/
+SourceCharacter :: "Any Unicode code point"
```
bhamiltoncx: One thing to note is that XML 1.0 forbids much of the same C0 control characters the previous definition did:

https://www.w3.org/TR/xml/#charsets

```
	Char	   ::=   	#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]	/* any Unicode character, excluding the surrogate blocks, FFFE, and FFFF. */
```

so if we want to maintain compatibility with XML 1.0 transport, we might want to keep that restriction so we don't mess up existing clients.

spec/Section 2 -- Language.md
bhamiltoncx: I would be explicit that this standard reflects the contents of the source document **after** it's been decoded from its binary encoding (e.g. UTF-8, UTF-16, etc.) to Unicode.

As such, after decoding, there are a number of code points which just cannot appear, including surrogates.

You shouldn't even include that language. I would just remove it.

spec/Section 2 -- Language.md
```diff
@@ -13,28 +13,44 @@ double-colon `::`).
...
+U+10FFFF, including surrogate code points, may appear in this sequence where
+allowed by the grammatical rules below.
 
+A [combining character sequence](http://unicode.org/faq/char_combmark.html) is
+treated as a sequence of individual Unicode code points and a sequence of
+individual {SourceCharacter}, even though they may appear to a user as a
+single character.
 
-### Unicode
+However, with the exceptions of {StringValue} and {Comment}, most of GraphQL is
```
bhamiltoncx: I don't really see what the point of this paragraph is. Either GraphQL source documents require non-ASCII values to be escaped, or they don't. I would remove it.

spec/Section 2 -- Language.md
```diff
@@ -13,28 +13,44 @@ double-colon `::`).
...
+encoding format, and the range of code points which it can encode, GraphQL
+documents may consist of any of those code points.
 
-UnicodeBOM :: "Byte Order Mark (U+FEFF)"
 
-Non-ASCII Unicode characters may freely appear within {StringValue} and
-{Comment} portions of GraphQL.
+### Byte Order Mark
+
+UnicodeBOM :: "Byte Order Mark (U+FEFF)"
```
bhamiltoncx: Like surrogates, this code point is irrelevant after decoding the document's encoding to Unicode. It won't appear.

spec/Section 2 -- Language.md
bhamiltoncx: I would recommend a new syntax to support SMP escapes. You currently support only code points up to U+FFFF via `\u EscapedUnicode`, but modern languages support SMP code points via a `\u{Hex}` syntax which supports between 4 and 6 hexadecimal digits.

spec/Appendix B -- Grammar Summary.md
bhamiltoncx: Also note that allowing `U+0000` (NUL) will be interesting for a number of implementations as well as security checks, since C string APIs will assume that NUL is the end of a string.

leebyron: Great reference, thanks for pointing that out. I had similar concern about allowing U+0 (see my 2nd comment on this issue) so happy to see that concern raised elsewhere.

spec/Section 2 -- Language.md
leebyron: I'm borrowing this idea from ECMAScript spec where the rationale was based on reading crappy files resulting from concatenating utf-16 files, and the parser breaking. This code point is irrelevant after decoding if it's the first point in the sequence, but IIUC Unicode doesn't do anything about BOM found within a sequence?

bhamiltoncx: It's up to the decoder to properly decode files before passing them to GraphQL, including decoding before concatenating.

spec/Section 2 -- Language.md
chris-morgan: Correction: between *one* and six hexadecimal digits (`\u{a}` is quite legal).

spec/Appendix B -- Grammar Summary.md
langpavel: `NULL` handling is hardcore thing :-)
By my opinion it shall be best to forbid unescaped NULL character in source AND should be explicitly declared that it must be rejected by tokenizer as an error anywhere in source.
In each reliable implementations should be test case for this case.

andimarek:
@wincent @leebyron can you give an explanation why this PR was closed? thanks

PR #233, opened 2016-10-31
Fix formating in section #2

IvanGoncharov:
I'm reading through the spec on GitHub at the moment.
So I see some discrepancy in GitHub render output.
So here are small changes to make markdown more GFM-friendly.

\+ I fixed some inconsistency in how tokens are defined.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -907,7 +907,7 @@ size `60` width:
 }
 ```
 
-** Variable use within Fragments **
+**Variable use within Fragments**
 
 Query variables can be used within fragments. Query variables have global scope
 with a given operation, so a variable used within a fragment must be declared
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -716,25 +716,25 @@ sequences must be used to represent these characters.
 
 StringValue :: `""`
 
-  * Return an empty Unicode character sequence.
+  + Return an empty Unicode character sequence.
 
 StringValue :: `"` StringCharacter+ `"`
 
-  * Return the Unicode character sequence of all {StringCharacter}
+  + Return the Unicode character sequence of all {StringCharacter}
     Unicode character values.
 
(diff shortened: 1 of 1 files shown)


spec/Section 2 -- Language.md
```diff
@@ -166,9 +166,9 @@ characters to support interoperation with as many other systems as possible.
 
 ## Query Document
 
-Document : Definition+
+Document :: Definition+
```
leebyron: The `::` is semantically meaningful - it distinguishes between character level rules and token level rules.

spec/Section 2 -- Language.md
```diff
@@ -716,28 +716,28 @@ sequences must be used to represent these characters.
 
 StringValue :: `""`
 
-  * Return an empty Unicode character sequence.
+  + Return an empty Unicode character sequence.
```
leebyron: Why change `*` to `+`? The `*` is fairly standard for bullet points and already renders correctly in GFM

spec/Section 2 -- Language.md
leebyron: For more detail on this, see http://facebook.github.io/graphql/#sec-Lexical-and-Syntactical-Grammar

leebyron:
I just fixed the table and bold rendering issues you pointed out here, so thank you for the careful read!

spec/Section 2 -- Language.md
IvanGoncharov: `1.` looks very strange if it's only one element in the list:
image

spec/Section 2 -- Language.md
IvanGoncharov: In this case, it should be referenced here: https://github.com/facebook/graphql/blame/master/spec/Section%202%20--%20Language.md#L8-L11
Because right now it very confusing since above text mentioning only `::` and not `:`

IvanGoncharov:
@leebyron Thank you, for incorporating formatting changes into spec.
I opened #238 issue with a suggestion on how to prevent confusion about lexical grammar in future.

PR #244, opened 2016-11-22
Clarify: Adds more mentions of `__` as reserved by introspection system.

leebyron:
The spec chapter on introspection mentions that type artifacts must not be named with "__" since it is reserved. This clarification repeats this information closer to the spec on type system itself along with the rules for determining if object/interface types are valid.


stubailo:
Does this exclude the possibility of using `__id` for IDs, which are not strictly introspection, or will that involve changing this language?

leebyron:
Certainly if ids end up both using `__` and not being part of the introspection system - then the existing language would need to change. I'd hope we can avoid that outcome though. So if ids are not defined in the schema and are accessed as meta-fields, they should use `__`, however if ids *are* defined in the schema (not a metafield) then they should not use `__`.

PR #261, opened 2017-01-31
Improved note on input coercion for null value via runtime variable value

sogko:
Follow-up to #260

Improved the note on input coercion for null value via runtime variable value for a list type.

The previous form was a little bit hard to read. Hopefully the changes improve its readability.

Thanks!


wincent:
Thanks @sogko!

PR #262, opened 2017-02-03
Forbid to implement same same interface twice

IvanGoncharov:
Both spec and reference implementation allow you to implement the same interface twice:
```
type Cat implements Pet, Pet {
  name: String
  currentlyPurring: Boolean
}
```

Moreover, I've discovered this issue in production GitHub's GraphQL API
image

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -539,7 +539,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
    no two fields may share the same name.
 3. Each field of an Object type must not have a name which begins with the
    characters {"__"} (two underscores).
-4. An object type must be a super-set of all interfaces it implements:
+4. An object type may declare that it implements one or more unique interfaces.
+5. An object type must be a super-set of all interfaces it implements:
    1. The object type must include a field of the same name for every field
       defined in an interface.
       1. The object field must be of a type which is equal to or a sub-type of


wincent:
On the surface that does seem like something that shouldn't be allowed.

On your wording:

> An object can not implement particular interface more than once.

I think there are two aspects to "implementing" an interface here that we should be explicit and tease apart here:

1. *Declaring* that a type implements an interface.
2. Actually *implementing* that interface, by implementing the fields declared in the interface.

The other rules in that section are enough to prevent the object type from doubly "implementing" an interface in the second sense already:

> The fields of an Object type must have unique names within that Object type; no two fields may share the same name.

So in the interests of clarity here I'd prefer wording that explicitly referred to sense "1.". Something like:

> An object type may declare that it implements 1 or more unique interfaces

Or, we could move that down to the "Interfaces" section where the `implements` keyword is first shown/demonstrated.

htc2u:
pull request

IvanGoncharov:
@wincent Agree. \+ I changed the order of rules.

> Or, we could move that down to the "Interfaces" section where the implements keyword is first shown/demonstrated.

I think rules about validating `1` and `2` on your list should live together.

wincent:
I like the new approach.

leebyron:
Great! Do you also have a PR for applying this to the reference implementation?

IvanGoncharov:
@leebyron Not at the moment but I can start working on it.

PR #263, opened 2017-02-06
fixed input object example

firede:
`type` => `input`


wincent:
Good catch. Thanks a lot @firede.

PR #266, opened 2017-02-14
Forbid to define duplicated member types in Union

IvanGoncharov:
This PR is intended to prevent a problem similar to #262 but for Union types.


wincent:
Makes sense, and it's a minimal change.

leebyron:
Great! Do you also have a PR for applying this change to the reference implementation, graphql-js?

IvanGoncharov:
@leebyron Not at the moment but I can start working on it.

PR #267, opened 2017-02-16
[RFC] GraphQL Subscriptions

robzhu:
Let's add GraphQL Subscriptions to the Spec!


rfcs/Subscriptions.md
```diff
@@ -0,0 +1,94 @@
...
+
+There are multiple valid answers to these questions, several of which have been proposed by the open-source community already [1][2][3]. At Facebook, we have been using “GraphQL Subscriptions” [4] at scale since 2015. Following our open-source philosophy, we now believe this system is useful and generalizable. We would like to contribute our ideas to the conversation.
+
+One of GraphQL's superpowers is allowing clients to specify exactly the data they need, sidestepping a problem we call “over/under-fetching”. With push-based solutions, we have a similar problem: “over/under pushing.” Under-pushing means we didn't send enough data, and the client now needs to make a follow-up network request. Over-pushing means we sent the client too much data; in the best case this data is irrelevant and doesn't waste anybody's mobile data plan, in the worst case, this data contains sensitive information that the client should not see. Just as request/response GraphQL addresses over and under-fetching, we believe GraphQL Subscriptions addresses over/under-pushing.
+
+**Possible Solutions**
+
+We broadly categorize realtime API solutions into three types:
```
syrusakbary: Not sure if is a typo, but there are *six* types defined below (instead of the *three* mentioned here).

stubailo: (If you look at the rendered view, it's a nested list)

stubailo:
This is really great! It was awesome that this took into account feedback from a variety of sources.

I think this sketch of the subscriptions lifecycle is compatible with the one currently implemented in our `graphql-subscriptions` package here, which is really nice: https://github.com/apollographql/graphql-subscriptions/

We've designed it in such a way that anyone currently using GraphQL.js can just drop it in and start using subscriptions right away, and we have some adapters for Redis, MQTT, etc.

There's also a transport protocol over websockets which we and some other organizations are currently using in production, which isn't coupled to any particular client or web server package, and implements all of the important lifecycle events necessary: https://github.com/apollographql/subscriptions-transport-ws

I encourage people interested in this proposal to try out those packages with their existing server and test out GraphQL subscriptions today!

leebyron:
Awesome! Here's a rough schedule of what happens next:

This is a really well written RFC, but this should not deter anyone from asking questions - both getting into the weeds and challenging core assumptions and aspects of the proposal. An RFC is the *beginning* of a discussion. Let's leave this PR open as a place for discussion for the time being.

OlegIlyenko:
Really happy to see progress on subscriptions! Good job, it looks great for far! 🏆 

> the input query and variables are mapped to a set of events

It would be interesting to see the semantics of this mapping. In particular, do the root subscription fields have any special semantics or the interpretation of the root field semantics can be defined by a GraphQL server (for example, does every root field represents specific type of event like `commentCreated` or topic `comments` or something else)?

Also another remark. The RFC assumes and emphasizes that subscription communication is bi-directional. I wonder whether this restriction is necessary or required. I definitely see appeal of having bi-directional communication. For instance, it can multiplex several subscriptions in a single WebSocket connection and provide extended life-cycle phases for further optimizations. 

But in the most basic form, uni-directional communication (server → client) can be sufficient to implement semantics described in the RFC. A while back I implemented [an example GraphQL subscription service that is based on SSE, which is uni-directional](https://github.com/sangria-graphql/sangria-subscriptions-example). As far as I can tell, it satisfies the semantics described in this RFC, but simplifies life-cycle a bit:

* `Subscribe` (client → server) - represented by server receiving a client request with subscription query
* `Subscription active` (server → client) - absent/implicit
* `Unsubscribe` (client → server) - client just closes the connection

Would love to hear your thoughts on this aspect of the RFC :)

Also regarding bi-directional communication. As far as I can tell, it defines a small protocol - something similar to [apollo WS transport protocol](https://github.com/apollographql/subscriptions-transport-ws#client-server-communication). Does this mean that in future we will have additional specification, alongside of GraphQL, that will define and standardize this bi-directional communication protocol (life-cycle phases, data format, command types like `SUBSCRIPTION_START`, `SUBSCRIPTION_FAIL`, `UNSUBSCRIBE`, etc.)?

stubailo:
@OlegIlyenko my interpretation is that "bidirectional" means that in your description there are arrows that go from client -> server and also server -> client.

taion:
Aren't any client -> server legs just mutations?

This looks consistent with what's implemented between [relay-subscriptions](https://github.com/edvinerikson/relay-subscriptions) and [graphql-relay-subscription](https://github.com/taion/graphql-relay-subscription) as well.

A few comments:
- An explicit "subscription active" message may not be necessary. I don't use one, anyway.
- https://github.com/robzhu/graphql/blob/80f17faac4838a3afcff3213c785ebaa533d53c8/rfcs/subscriptions_01.png might be clearer if there were a box for "other backend components" that "domain layer" and "event B" connect to

stubailo:
We started our implementation without a "subscription active" notification, but later had to add it. Because the first result from your subscription could arrive a long time after the initial request, you don't know if you're actually receiving those updates. In some cases, you want to be able to look at the state of the subscription to decide if you should be polling for data or simply waiting for events to arrive, especially if you don't want to miss some crucial updates to the information you're looking for.

Essentially, it can be very hard to know what happened if you subscribe and get back total silence - did the subscription somehow fail or stall, or did it successfully start and you should expect to get results in the future? @taion I think you're using socket.io right? Maybe they have a system that checks if the subscription is successful under the hood?

taion:
@stubailo 

To clarify, I'm saying that "subscription active" notifications aren't required in all cases.

It really depends on the app. If it's something like chat, then "subscriptions are down" is a big deal and you want to show some user feedback. For something like, say, showing "new comment" prompts on a GitHub issue page, you don't really care.

A general purpose subscription layer probably wants to expose something like those "subscription active" notifications to handle all uses cases, but they're not always necessary in more specialized cases.

stubailo:
Good call - I agree that it is not necessary in cases where the subscription is not providing critical functionality. I think one other factor to consider is that returning an error about the initial subscription definitely is necessary - for example, if the query is not valid at all you want to receive some validation error in response to your initial subscription. Without a message that there was a successful subscription (essentially it just tells you the query validated etc) there's this time when you're not sure if there will be an initial error or not.

taion:
Yup – error handling is necessary in general. I guess it depends how in-detail this spec wants to go on what the network layer should do.

deanius:
Questions like whether a subscription can do events singly or must be multiple, as asked in the blog post, point to the possbible usefulness of an abstraction like Observable, which is not inherently tied to multiplicity of responses. Also, that it is moving through standardization in the language would be helpful too.

I'm also a fan of the DDP 'ready' event that announces the change from serving up existing old data to serving up new responses. While not strictly necessary, it lets you do things like remove loading graphics as you already know quite well @stubailo :)

My 2c on unsubscribe is it should be a message as well (as in DDP), but obviously the server should do the right thing if a connection goes stale and no unsubscribe was received.

PS I'm in the category of someone whose org may adopt GraphQL if it had subscriptions. Right now we're rolling something on Meteor DDP until we see if Apollo could give us a compatible-enough experience.

taion:
Observable is really a JS-level implementation detail, no? I don't think it's pertinent at the level of this spec.

rmosolgo:
Thanks for writing this up, I'm looking forward to seeing how it evolves! I had a couple questions:

- Somehow, I was previously under the impression that subscription requests _also_ returned an immediate result, as if they were queries. That's not mentioned here, right? So I was mistaken about that? 
- Would it make sense to add one more arrow to this diagram? I thought there might be one like this: 
  
  image

  After the subscription system is notified of an event, it tells the GraphQL server to re-evaluate the query, and the GraphQL server gives a new response (in magenta) to the subscription system. (Then, the subscription system sends it to the client over the particular transport layer.) Did I understand correctly? That's how I made sense of this arrow in the later diagram:

  image

taion:
I don't think in general subscription queries should return an immediate result. If I have a subscription on e.g. items being added, what would that result be?

deanius:
@taion - The point of it wasn't to prescribe an implementation, but to convey semantics by referring to an existing spec to broaden the discussion. The contract covers issues like you mentioned above such as whether you get a result upon subscribe. http://reactivex.io/documentation/contract.html

smolinari:
Whatever the solution may be. I 👍 the addition to the spec, because without a subscription system, GraphQL can hardly be considered as "state of the art", when if fact, it really should be. 

Scott

theorygeek:
One thing that's not necessarily clear to me (and to be honest, I haven't been down in the trenches with subscriptions, like others commenting here) is this:

When the client wants to create a subscription, is it just executing a GraphQL query with a special operation? Or is it sending a package of "something" (_including_ a GraphQL query, variables, etc) to something else, and then that something else is interacting with GraphQL?

I was previously under the impression that if I sent this query to my GraphQL server, using the exact same transport/etc mechanics as any other query, that it was creating a subscription:
```graphql
subscription {
  someField {
    # sub-selections
  }
}
``` 

In this scenario, the GraphQL library ([graphql-ruby](github.com/rmosolgo/graphql-ruby) in my case) is handling the "book-keeping" of the subscription, and somehow I'm plugging in:
- Events
- A special network layer

But after reading this spec, it sounds like I'm instead directing this to something _other_ than my GraphQL library, handling the book-keeping myself, and periodically running a not-so-special query and sending the (black box) payload to the client.

Essentially, what I'm getting at, is should we expect authors of GraphQL frameworks to be providing the book-keeping for subscriptions, etc and then exposing some fancy integration points (events and networking)? Or should we expect a new eco-system of subscription frameworks to grow up alongside existing GraphQL frameworks?

taion:
I imagine implementation-wise there's not much actual change, right? As set up right now, GraphQL.js can execute a `subscription` query, but the actual semantics of the full request live outside. It's like the split between `graphql` and `express-graphql` for the standard query/mutation side.

stubailo:
> Or is it sending a package of "something" (including a GraphQL query, variables, etc) to something else, and then that something else is interacting with GraphQL?

Yes, the idea is that you can add subscriptions on top of the existing GraphQL execution libraries by adding a subscription manager or subscription gateway. And that thing's job is to re-execute the subscription query against the GraphQL server in response to the event.

> I imagine implementation-wise there's not much actual change, right? As set up right now, GraphQL.js can execute a subscription query, but the actual semantics of the full request live outside.

I think that's about right - in terms of a library like GraphQL.js I think the one thing that could be added is a way to map the subscriptions fields to events. I think there would be a lot of value in putting that right next to the resolver code.

Oh, and one more thing - the resolver for a subscription field has one more parameter now, which is the "payload", so now that resolver takes in `rootValue`, `args`, `context`, and `payload`.

That could come in three different places:

1. The payload could be passed as the root value, that would be consistent with looking at the subscription as a continuation of the mutation - so the mutation can push its result over the message system, and the subscription field picks up where it left off. But that could be confusing given that root values are used for other stuff as well.
2. The payload is attached to context, but that feels like a bit of a hack to me especially since it means _all_ fields will be able to access it.
3. There's a new parameter that only subscriptions field resolvers get access to.

I think I'm in favor of (3), which requires a change to GraphQL.js to have the execution function take an extra `subscriptionPayload` parameter or similar.

taion:
It'd be great if running a subscription could return some description of the relevant event set as a first-order concern, though the current workarounds aren't too bad.

I'd prefer subscription payload come through as `rootValue` for consistency.

stubailo:
> I'd prefer subscription payload come through as rootValue for consistency.

That's what the `graphql-subscriptions` npm package does right now, and it's not too bad. But we don't really use `rootValue` for anything else since we basically put everything on `context`.

taion:
Yeah, I think that's the right thing, no? Now that `context` exists, `rootValue` is exactly just for injecting payload in that manner, I think.

smolinari:
I've read through everything once more and I think I might be missing something (probably general understanding). But, how does or should the subscription system fit in with Introspection?

Scott

PaulLeCam:
Great to see this added to the spec!

We have been using relay-subscriptions for the last few months at Mainframe, and I wonder how much of this spec differs from relay-subscriptions’ implementation?
In our case, with the only exception of handling notifications, all our subscriptions are linked to queries, so they notably use the same fragments, as in [relay-subscriptions’ example](https://github.com/edvinerikson/relay-subscriptions/blob/b66edefeb9dff6ba1dea3fe19a42d0611540c419/examples/todo/js/subscriptions/AddTodoSubscription.js#L22).

One issue we needed to address was the possibility to miss relevant events between the time a query is resolved and the subscription is created on the server, as the current flow using relay-subscriptions is client queries -> server responds -> client creates subscription using response payload -> server receives subscription request -> server publishes subscription events when receiving internal events.
Our current workaround is to have the server provide a “version” (opaque string for the client) when responding to any query, that the client sends along with the subscription requests related to the query, allowing the server to look back at the events log and publish the needed subscription events to the client so it can catch-up on any change.
We implemented this logic in our custom network layer, but considering it is affected by the relations between queries and subscriptions, I wonder if it is something that could be part of this spec?

taion:
Again, I think that's the sort of thing that really depends. In many cases I think that's just not required, and would instead just be extra overhead.

rmosolgo:
That's interesting, it's actually a case where my _misunderstanding_ of subscriptions would come in handy. 

In that scenario, a subscription request also returned an immediate response, so you could get the current server state _and_ make a subscription in a single operation. 

However, I'm not sure how you'd sort it out in a case like this:

```graphql 
subscription getNewComment($postId: $ID!) {
  commentAdded(postId: $postId) {
    newComment { author { name }, message }
    post { title, comments { message } }
  }
}
```

In that case, the first response comes with `newComment: null` but `post: { ... }` is present, so you can initialize the UI with that state. 

The problem is, on later requests, you'd _still_ get `post: { ... }`, which is overfetching. 

So, I don't think it's much of an answer, but I thought I'd share anyways 😆

taion:
Presumably at the implementation level, you can put something on `context` indicating whether this is the initial subscription request, then do whatever is appropriate in the resolver.

There's nothing that _prevents_ an implementation from immediately firing an update corresponding to the current state – I just don't think it's required in the general case.

paralin:
I'm going to just drop a link to my Real-Time GraphQL Project here - http://github.com/rgraphql/magellan - where I define a protocol for two-way communication between a GraphQL client and server, including real-time updates to results **and queries**.

Subscriptions seem like an inelegant solution to me. They are extremely limited, one-way, one-time subscribe, unable to update query parameters after the subscription has started, etc. As you're finding now in this thread, these limitations make using and implementing subscriptions a muddy deal.

If something's not simple, it's probably wrong. And subscriptions to me seem to have a lot of cludginess as a result of jerry-rigging them in to an existing one-way one-off GraphQL spec. With the assumption that we have a two-way communication path with the server, a lot of optimizations and improvements can be made. I've written a proof of concept of this at the rgraphql repo I linked above. A demo is also available here: http://rgraphql.org

Curious to hear your opinions on this.

dschafer:
Liking the looks of the RFC so far!

One thing that's not in the RFC right now but that might be valuable: do we want to require that the sequence of payloads sent to the subscriber is determined only by the root subscription field (and its arguments), and that in particular the *fieldset* of that subscription field does not affect the sequence of payloads?

----

I think a concrete example probably demonstrates the value of this best; I imagine a developer on the client writing

```graphql
subscription DeveloperSubscription {
  commentCreate(postId: 1234) {
    comment {
      text
      timeCreated
      author {
        ...CommentAuthorFragment
      }
    }
  }
}
```

This is a pretty basic subscription; every time there's a new comment, push a payload that includes the text of the comment, time created, and the author information (where that's delegated to another fragment, one that presumably defined by `CommentAuthorView` or something).

Now I imagine using a client-side GraphQL framework to issue that subscription; if that framework has a normalized cache, it might want to augment the subscription to fetch IDs for some of the objects so that it can update the cache. So rather than issuing `DeveloperSubscription`, it would instead issue:

```graphql
subscription FrameworkAugmentedSubscription {
  commentCreate(postId: 1234) {
    comment {
      id # Added by framework
      text
      timeCreated
      author {
        id # Added by framework
        ...CommentAuthorFragment
      }
    }
  }
}
```

Meanwhile, our developer wants to debug this, and opens up Graphiql. The whole author fragment might be big and unwieldy and all the developer wants to do is verify that the subscription is working as expected, so the developer might issue the following in Graphiql:

```graphql
subscription DebuggingSubscription {
  commentCreate(postId: 1234) {
    comment {
      text
    }
  }
}
```

----

Of course, the *payloads* received by these three subscriptions will be different between the three, since the field sets are different. But because the root subscription field on all three was the same, `commentCreate(postId: 1234)`, it feels like we should be able to guarantee the sequence is the same: if one subscription receives a payload, the other two should as well. It seems like without the guarantee, we might have three potential issues.

 - A client-side framework can't safely modify the field set without risking changing the functionality of the subscription.
 - A debugging developer can't reduce the size of the field set in Graphiql while guaranteeing that they are still seeing the same sequence as the client they are debugging.
 - We kinda break encapsulation: in the `DeveloperSubscription` example, the developer's intention is to subscribe to all comments created... but if the field set can alter the sequence, then changes to `CommentAuthorFragment` can alter the sequence. This feels kind of weird, the developer only spread in `CommentAuthorFragment` to ensure it fetched the right fields in the payload to render the comment author; the idea that a change to `CommentAuthorFragment` could alter the functionality of `DeveloperSubscription` feels odd.

Thoughts?

rmosolgo:
@dschafer The spec contains special rules for root fields on a `Mutation` type. It sounds like you're suggesting that the root fields of `Subscription` type have a special property also: _only_ the root fields determine how a client is subscribed, so if two queries have the same root fields, they result in the same subscribed state (regardless of any child selections).

Is that correct? If so, big :+1: from me. I somehow had that assumption, but I didn't notice it was missing from this RFC!

taion:
I wonder if that might not be better as a best-practice type thing for typical use cases rather than a part of the spec?

Imagine a subscription like:

```graphql
subscription {
  commentUpdate(commentId: 3) {
    author {
      karma
    }
  }
}
```

With my current implementation, if I care about author karma changes here, I would just create a new subscription, but it seems like it'd be odd to prohibit this subscription from firing on e.g. author karma changes, or else force something like:

```graphql
subscription {
  commentUpdate(commentId: 3) {
    score
  }
}
```

to also fire when this sort of thing happens.

And what about optimizations, e.g. for a partial update if none of the fields on a given subscription have changed in that partial update?

OlegIlyenko:
@dschafer This sounds great 👍. This is precisely what I mean in my question:

> It would be interesting to see the semantics of this mapping. In particular, do the root subscription fields have any special semantics or the interpretation of the root field semantics can be defined by a GraphQL server (for example, does every root field represents specific type of event like `commentCreated` or topic `comments` or something else)?

This definitely helps to make the subscription semantics a bit more clear (as you mentioned, it has a big advantage for generic tools). In my experiments, I also found that this semantics for root subscription fields makes it quite easy to use `Observable`s instead of `Promise`s for root query fields.

dschafer:
> @dschafer The spec contains special rules for root fields on a Mutation type. It sounds like you're suggesting that the root fields of Subscription type have a special property also: only the root fields determine how a client is subscribed, so if two queries have the same root fields, they result in the same subscribed state (regardless of any child selections).

So subscription root fields will end up being special no matter what, since they both end up "resolving" to a value and identifying a "sequence of payload events" (which query/mutation root fields don't do). I guess my claim is that the "sequencer" should obey similar constraints to the resolver; in particular, in the spec the field resolver looks like:

```
ResolveFieldValue(objectType, objectValue, fieldName, argumentValues)
```

Notably, `fieldSet` is *not* a parameter there; it's provided in `graphql-js` and other implementations to allow for reasonable optimizations (so you can do `SELECT name FROM users WHERE id=1` if they only asked for `{name}`), but the field value itself shouldn't change based on the field set. It would be very strange if `user(id:1){name}` returned `{name: "Dan Schafer"}` but `user(id:1){name, firstName}` returned `{name:"Lee Byron", firstName:"Lee"}`, and I'd argue that would be spec non-compliant.

I'd envision the "sequencer" acting the same way; the "sequence of payload events" that it identifies should only vary depending on those four values (objectType, objectValue, fieldName, argumentValues) and not on the field set. So in some ways, I think this makes subscription fields *less* special, since it increases the parallelism between how they sequence and how all other fields resolve.

----

> I wonder if that might not be better as a best-practice type thing for typical use cases rather than a part of the spec?

I think the primary benefit to having it be in the spec is for client tooling. If this guarantee isn't in the spec, then a client can't safely add a field to any subscription (and by extension, it can't really add a field to any *fragment*, because the fragment could be used in the subscription).

I guess it feels like there's a pretty important guarantee that exists today in GraphQL: you can always safely add a field, and know the only thing that will change about the response is that there will be a corresponding key-value pair in the response for your field; the rest of the response will be the same. So while adding this guarantee in the subscriptions spec adds a restriction on subscriptions, it more importantly maintains an existing guarantee about field sets.

----

> With my current implementation, if I care about author karma changes here, I would just create a new subscription, but it seems like it'd be odd to prohibit this subscription from firing on e.g. author karma changes
> And what about optimizations, e.g. for a partial update if none of the fields on a given subscription have changed in that partial update?

So because there's a lot of freedom in how we define the subscription root type, it feels like all of these possibilities/optimizations would still be available; they would just have to be more explicit. The intent would be that this change would just prevent the field set from being used as the trigger for a behavior change; instead, the behavior change would have to be made by calling a different subscription field, providing a different set of arguments to the subscription field, or using a directive, which feels like a reasonable tradeoff to me. The downside is that the field set cannot change the behavior, so we have to be explicit if we want to change behavior; but the upside is that the field set **cannot** change the behavior, so users/tools can safely manipulate the field set without worrying that we've changed something.

taion:
Thinking a bit more on something like the "empty partial update" case – even if the GraphQL subscription itself uniquely identifies a sequence of events, would the network layer be required to not apply optimizations like not sending empty updates?

Contra request/response semantics on typical GraphQL operations, it seems like there's more scope for the network layer (on the server side) to do particular things here, such that spec guarantees aren't the final word for behavior in practice with an actual server.

stubailo:
I think it would be very useful to recommend that the subscriptions system doesn't even know about anything other than the root subscription field, because it would be pretty unfortunate to have the server return different events based on the selection set. That would be like a mutation making a different change to the server based on the fields fetched afterwards.

However, I think it's accurate that some subscriptions implementations would want to do throttling or batching of some kind, which means you don't get exactly one result for one event fired. So I'm not sure how the specification could have an opinion on how many messages are delivered, even though it would be very nice to have the re-running be independent of the selection set.

stubailo:
For your case @taion I think `commentAuthorKarmaUpdated` would be good to be a separate subscription from `commentScoreUpdated`. Or, even something like `commentUpdated(events: [SCORE, AUTHOR_KARMA])`.

taion:
To be clear, I'm just trying to work through hypotheticals here.

@stubailo: you'd have a better answer for this than me – what would it look like to build something like a live query on top of GraphQL subscriptions? Naively I'd have imagined that it would require subscriptions with events that are payload-dependent.

smolinari:
@stubailo - my argumentation exactly! Live queries would need to be specialized queries and contrary to the Facebook argumentation, not the standard read queries set to live. That would be very problematic. Live queries should be specific to the data needing updating and it will be that way 80-90% of the time. It seems like the live query idea was only half way thought through....

@taion - My words! However, let's have GraphQL subscriptions BE a live query system. How userland gets the data to be live should be up to them. It is a userland concern, whether it be events, databases that do live queries themselves, polling some other API (which defeats the purpose of live queries, but sometimes it just isn't possible otherwise). There are a myriad of possibilities and now GraphQL is forcing something on its users with this semi-wild events system. KISS 😄  

Scott

paralin:
@smolinari @taion this is exactly what I have built in rGraphQL - http://rgraphql.org - live data. Users decide how to feed data into the resolver system. It adjusts automatically to your code via reflection. Then query changes and result changes are streamed over a two way connection.

taion:
It would not be appropriate for GraphQL subscriptions to mandate live queries. In general doing live queries in a scalable manner is an unsolved problem, and it's not in general the correct solution for all use cases anyway.

smolinari:
@taion - It would be just as appropriate as the event system being (possibly) mandated, except considerably simpler. I also don't understand your comment about live queries being an unsolved problem in terms of scaling. What problem is there?

Scott

Siyfion:
> It would not be appropriate for GraphQL subscriptions to mandate live queries. 

I cannot agree with this enough, we're talking about two vastly different things here @smolinari, live queries trigger on *any* update. Subscriptions are fine-grained and specific. Also live queries involve a lot of backend logic in order to support them, something GraphQL has managed to stay away from as yet.

smolinari:
@Siyfion - maybe my definition of a live query is wrong then, because in my book, a live query is a particular query that is "deemed" live. In other words, you call certain queries as live through a different subscription method. You could say it is like a subscription system, but without all the event stuff. How updates are fed back to the GraphQL server is domain specific and yes, it could get hairy, as we can see from the solution offered here. 

This is an example of something similar to what I have in mind: http://blog.parse.com/announcements/parse-server-goes-realtime-with-live-queries/

> To use live queries, you just construct a Query like you normally would and call subscribe() instead of find().

Scott

leebyron:
Awesome conversation around this so far.

One concern I wanted to throw in which I've discussed with @dschafer and @robzhu offline about is what the right abstraction should be for the source of triggering the execution of payloads. Right now the RFC suggests:

> When the Subscription is created, the input query and variables are mapped to a set of events to which the Subscription listens.

I think this actually exposes too much about the implementation details of a subscription and makes too strong of an assumption that a Pub Sub system is the *only* way rather than the *primary* way to implement subscriptions. The wording is actually pretty close to what I would recommend, but how we read it is importantly different.

I think the appropriate abstraction requires two things:

1. A relationship between a client and service must be *bidirectional*. That is, in addition to a client being able to initiate connection with a service, the service must also be able to communicate directly to the client. This differs from a request/response model where a service can only "speak when spoken to".
2. Each subscription instance corresponds to a single sequence of notifications emitted over time, e.g. a single *Observable*. Each emit from the Observable results in a Subscription payload.

A PubSub system can pretty clearly satisfy both of these, "Pub" is the reverse direction of "Sub" fulfilling bidirectional, and if more than one "events" or "topics" are relevant, this can still be modeled from GraphQL's point of view as a single merged Observable.

Importantly, service statefulness is not required. Webhook APIs can also satisfy these statelessly. Webhooks include a URL to POST payloads to fulfilling bidirection, and the "Observable" exists only abstractly via some other part of the service that's run as a result of some other incoming request (i.e. GitHub `post-receive` git hook).

leebyron:
Following up on that, this different framing of the base abstraction also captures @dschafer's concerns above:

May a change in selected fields result in a different sequence of payloads?

stubailo:
@leebyron I agree that it might be better to avoid mentioning events explicitly. In the proposal I shared with you guys before I had framed it as a "callback" provided to the application code by the subscriptions system, where the callback can be called as many times as desired with payload objects. Perhaps "observable" could be used as a term, but I find that term pretty heavily overloaded by specific implementations like RxJS that bring a lot of baggage.

> the input query and variables are mapped

I'd also propose this wording could be changed to "root subscription field and arguments" to suggest that implementations should avoid depending on parts of the query other than the root field.

@smolinari Subscriptions as proposed here, IMO, should almost never be used as a live query because they don't provide some relevant guarantees, and I think it might be good to ensure that there's a very clear distinction. Live queries would be a great feature to have in GraphQL, but I don't think this proposal should be expanded to also include that separate concept.

taion:
@leebyron / @stubailo:

To what extent is it actually necessary to codify the behavior of the subscription system? At a minimum, GraphQL itself only needs to deal with two cases, no?

- On a subscription being received, given the subscription query, produce some sort of `NewSubscriptionResult`, e.g. for the subscription to use to identify a set of events
- At future points in time, given the same subscription query, produce some sort of `SubscriptionUpdateResult`, e.g. to send on to the subscriber

Does the GraphQL spec need to outline what the subscription system does? For example, the guarantee mentioned above can just be stated as the `NewSubscriptionResult` being only a function of the root query; further semantics don't concern the GraphQL engine directly.

@stubailo 

I think the best scenario would be that the subscription support is event-based, but that it would be possible to use it as a lower-level building block for a live query system. Do you think this is impractical?

stubailo:
> To what extent is it actually necessary to codify the behavior of the subscription system?

I think the value of specifying some part of GraphQL is to make it predictable. You want to know what the life cycle of a subscription looks like, how it might change when requesting different fields, etc.

> Do you think this is impractical?

Live query systems can easily be built as extensions to GraphQL completely independent of subscriptions. Do you think some part of the current proposal makes live queries harder to implement than they were before?

f:
Hi!

I implemented a naive and **simple** GQL Subscription implementation using `EventSource` and [`graphql-ruby`](https://github.com/rmosolgo/graphql-ruby).

```js
var source = new EventSource(`/graphql?query=
  subscription ($id: ID!) {
    todoAdded(id: $id) {
      id
      text
    }
    todoUpdated(id: $id) {
      id
      isCompleted
    }
    todoRemoved(id: $id) {
      id
    }
  }`)
source.addEventListener('graphql.todoAdded', function (message) {
  // "todoAdded" field resolved.
})
source.addEventListener('graphql.todoRemoved', function (message) {
  // "todoRemoved" field resolved.
})
source.addEventListener('graphql.todoUpdated', function (message) {
  // "todoUpdated" field resolved.
})
```

The problem here (which is not a problem for implementation but the RFC) is it's executing queries lazily with back-end generated variables (publisher at back-end sets the variables).

Since this approach works on this example, I have no experiment on a more complex application.

This is the working demo:

Gif

I know this approach has some pitfalls. E.g. I cannot subscribe to an individual field at back-end since I'm listening to back-end publisher to take the variables and execute the query comes from client. And it's not very intuitive. But works exactly how I need.

Here is the link you can try out [https://github.com/f/graphql.js-demo](https://github.com/f/graphql.js-demo)

romandvoskin:
I've read the RFC (excellent job @robzhu!) and all of the (insightful) comments. I want to understand if GraphQL Subscriptions can be used for implementing hot + cold observables, where a user first gets the initial state (cold observable) and then receives subsequent changes (hot observable)? In such a system, the hot observable (subscription payload) must be combine-able with the current client state. 

1. Should the RFC include details about behavior regarding the first payload vs all subsequent payloads. For example, the initial state (cold observable) must be sent as the first payload, if the value is present and the context makes sense?

2. Should the RFC include details about the order of subscription payloads as they relate to the firing order of the underlying events. 

3. Should the RFC propose a special metadata field in a subscription payload to serve as place for data like: eventIds that triggered the subscription payload, a sequential payload id to serve as a data anchor to detect payload loss when using non-reliable transport libraries like socket.io (where loss can occur when protocol switching)?

taion:
@stubailo On live queries, that makes sense, thanks.

@romandvoskin To me, (1) seems application-specific, while (2) and (3) seem transport-specific. To e.g. the points earlier, these don't seem like things that should be specified at the level of the query language, as they relate to properties that you may or may not want in a specific implementation, in an application-specific manner.

rzachariah:
I wrote some feedback to @robzhu privately, which he asked me to share here. I will.

I think this proposal will enable a useful subset of realtime/reactive use cases.

I see the subset as event-triggered schema projection. Whenever one of a specified set of underlying events happens, run a user specified projection of the GraphQL schema.

Imagine we have event sources `IObservable<X> xs`, `IObservable<Y> ys`, `IObservable<Z> zs`, ..., a GraphQL schema, plus a function that maps from `TSchema` (the GraphQL schema) to `TPayload` (the user specified subset of the schema). We could model the proposal as an Rx Observable query.

```
public IObservable<X, Y, Z, ..., TSchema, TPayload> GraphQLObservable(IObservable<X> xs, IObservable<Y> ys, IObservable<Z> zs, ..., TSchema schema, Func<TSchema, TPayload> query) {
  var xSignals = xs.Select(_ => Unit.Default);
  var ySignals = ys.Select(_ => Unit.Default);
  var zSignals = zs.Select(_ => Unit.Default);
  return Observable.Merge(xSignals, ySignals, zSignals)
    .Select(_ => query(schema));
}
```

This Observable query is quite general, as the user may specify whatever query he wants, and we can select any list of underlying events as triggers.

However, for some use cases, this approach may not be significantly more efficient than polling.

Consider the following case.

Let's say I want to produce a trending topics live dashboard. This dashboard will show me a top 10 list of the most popular Twitter hashtags in the last 6 hours.

There is one underlying event stream: Tweets.

One approach, which could be implemented with GraphQL subscriptions, would be trigger a re-query across all the tweets for the past 6 hours every time a new tweet occurred.

A more efficient approach would be to set up a stream processing pipeline, say with Apache Storm, that processed the tweet stream through a directed graph of computation.

Spout: Tweet stream
Edge: Random partitioning (load balancing)
Bolt Type 1: Parse hashtags from tweets (stateless)
Edge: Partition by hashtag
Bolt Type 2: Perform 6 hour rolling counts of hashtags (maintain intermediate state)
Edge: Partition by count object
Bolt Type 3: Intermediate rankings of counts (maintain intermediate state)
Edge: Global grouping
Bolt Type 4 (singleton): Total rankings of counts (maintain intermediate state)
Edge: Random partitioning
Bolt Type 5: Publish (stateless)

Here's a post that details how to implement this stream processing pipeline in Storm.
http://www.michael-noll.com/blog/2013/01/18/implementing-real-time-trending-topics-in-storm/

Here, we set up the processing pipeline once, at the dawn of time, or when a user expresses interest. Once we set up the processing pipeline, there's no querying required when an event happens. We take an underlying event stream and transform it into a more interesting stream, which pushes data at the user. It's all reactive.

I guess the distinction is push versus pull. Querying is a pull activity. In the proposed (RFC) model, we pull at sources, not on a fixed interval trigger like polling, but on an event-based trigger. That's a possible and useful approach to solving realtime use cases, but it may not be practical (efficient) for some applications.

All that said, I still think the proposed spec covers a useful subset of cases. It will be wonderful to avoid under-pushing/over-pushing in those cases.

romandvoskin:
@rzachariah This is an interesting observation. I wonder if the feature you're getting with Storm is possible here by creating an event (becoming an event source) as a consumer of a subscription. To use your use case, an initial subscription to tweets is consumed by an aggregator (which as an event source) publishes an event whenever the result (top x tweets) changes. What do you think?

rzachariah:
@romandvoskin I think that's right. I could create a GraphQL Subscription server that offered both a Tweets subscription based on raw tweet events, and a Trends subscription based on manufactured top-x rankings events. There is a relationship between these event streams, but perhaps those are implementation details of the server. Perhaps the GraphQL Subscription spec should be ignorant of any distinction between raw and manufactured events.

smolinari:
@rzachariah  - Correct me if I am wrong, but your example is one where a polling type push of data (the event to GraphQL) is going to be the result of the aggregation of top tweets. In other words, you wouldn't want to push data for each tweet, but rather only every minute or 10 seconds or whatever is smart for your users and your system. It is a great example of how diverse the events could be, which trigger publishing data to subscribers.

On a side note: The more I read and learn, the more I see that the pub-sub system, which is put forward in this RFC, is actually what I consider a live query system. So, I apologize for my ramblings. 

I've also noticed that "live query" has been defined in different ways for different technologies, from databases to jQuery to Meteor. So, seeing as there is no one clear standard for what a live query is, I'd like to also suggest we call this system "GraphQL Live Query", simply because that name is a lot sexier than "pub-sub" or "GraphQL Subscriptions". 😄 

Or is there a clear definition of what a live query is? Let's forget what was put forward as a live query in this RFC, as it also isn't a standard either. 😉 How about it?

Scott

rzachariah:
Hi @smolinari. Yes, in my example, the top trends event stream is a streaming aggregation of the tweet stream. The throughput (events/second) of this derived event stream would be much lower than the raw event stream that produced it. There's some natural filtering as a result of the top-x operation. In addition, yes, it would be reasonable to introduce additional delays/batching to reduce the computational load. A top 10 trends dashboard that's live as of a few seconds ago, or even a minute ago, seems reasonable.

I was also trying to suggest a more efficient alternative to simply iterating over the last 6 hours of tweet data every time you're interested in updating the trends dashboard. I proposed a storm topology as an approach that maintains intermediate state about the computation (sliding window counts by topic, current rankings), so there's very little work to do whenever it's time to process a new tweet. It's an event processing pipeline made up of a chain of subscriptions. An Rx query would be another example of an event processing pipeline made up of a chain of subscription, and it's possible to write one for this example (though the Rx version would not be as scalable).

One could think of a storm topology or an Rx query as a live query. It's a query that, once deployed or subscribed to, is always running and automatically pushes updates in reaction to new input events.

jamesgorman2:
Some thoughts on this I've had reading the RFC and this thread (in no particular order):

* the case for subscriptions _inside_ graphql needs to be made more strongly. 'Ah well, you can just implement that in your transport layer' seems to have come up a few times. 
* in `img_01`, why does the subscription layer talk directly to graphql? This feels like a break in the abstraction.
* there feels like there is a gap between Data-transform pipelines and Live Queries for entity-state updates. The main problem with live queries is not business logic decisions[1] but computation and dependency complexity. If I have an _n_ depth query I am subscribed and an object at depth _n_ changes, propagating this back down to the root subscriptions is difficult (not impossible, but I'd rather not have to think about how to get it to wire up!). Comments by @leebyron, @rmosolgo and @stubailo about restricting subscription semantics to the root parameters all speak to this. 
  * I think restricting subscription semantics to the root parameters is generally a good restriction to make, but it needs to be made clearly and with reasons given if it is to be more than a recommendation
  * talking about live queries is unclear (and has already caused some confusion). This seems to be a DB-land phrase originally, but it does appear to be causing confusion. It would be better to frame everything in terms of the construction of graphql queries and their result sets.
  * distinguishing between events (state mutation events), actions/commands (not in here yet, but closely related in the eventing world) and whole state objects is immaterial given the above, particularly since they are all types of event and the distinction is unenforceable.
* in `img_01` most the green subscription layer seems unnecessary. A graphql subscription would work without any bi-directional traffic from downstream. Leaving this in may reduce the clarity of the message (it could perhaps go in an addendum to describe a more complex system).
* there is definitely room for best practice stuff around designing events, but this should be demarcated as best practice since it's stuff that is contingently true.

[1] here I disagree with Dan Schafer and Laney Kuenzel's assessment of Live Queries in Relay (though not their decision). I think their domain problem (ie, when do I update?) is technically solved (eg using some sort of buffer/aggregation or debouncing, ignoring the computational cost aspect) but requires a product decision as to what should updates be seen when. Their second point about computational complexity speaks only to that specific use case, not all use cases.

taion:
FWIW, I'd strongly recommend taking a look at graphql-subscriptions and graphql-relay-subscription, if you haven't already.

I think there already _are_ implementations that hew quite closely to what's in the proposal. Some of these are even already live.

Those systems, their relevant API warts, and the current GraphQL reference implementation (which has already moved ahead of the spec here) are in some sense existence proofs for what makes sense engineering-wise to split between core and transport layers.

jamesgorman2:
Thanks Jimmy. While I've read around them I haven't looked at them in
detail yet. My main goal wasn't to dispute that subscription is useful, or
even that this is heading in the right direction,[1] but to improve the
clarity of the whys. Once this is tighter, it is easier to debate 'is this
the problem we want to solve' and 'is this a solution to the problem'
separately.


[1] except maybe the definition of events

smolinari:
@rzachariah

> It's a query that, once deployed or subscribed to, is always running and automatically pushes updates in reaction to new input events.

That sounds like exactly what the one-sentence-definition of GraphQL Subscriptions or ehem.... Live Query 😄  should be. 😉 

Scott

paralin:
@smolinari I hate to keep mentioning it, but I have actually proven in rgraphql that "live" Graphql queries with arbitrary data sources (no DB related logic) and mixes of static and dynamic data is easily possible without changing the current spec at all. What these guys are working on is a extension to the spec that works around the cludge of typical one-way data sources like Pub/Sub or long polling on requests. Otherwise there's no advantage to the flat limited subscription model when compared to a streaming system.

smolinari:
Ok. @paralin Still, I am confused. Are you making that point for me, because I am misunderstanding something or are you making that point to everyone else, to say all this isn't really necessary to make a live query system within the spec? Or maybe even both? LOL! 😄 

Scott

paralin:
The spec already covers everything for live queries, other than I suppose directives to control the execution. But that's more of a server specific thing anyway. 

What's the difference of a live query to the spec? You're still specifying what data is required in the same syntax.

robzhu:
Hey all, amazing feedback and discussion so far, but it’s clear the github PR single-comment-thread feedback format is not suited for hosting multiple conversations at once. 

I’m going to go through the feedback and organize everything into **open** and **closed** topics. **Closed** topics are either obvious or have clear consensus and will be merged into the proposal in the next edit. **Open** topics are still being discussed (I will try to tag everyone who’s been involved/interested on each topic). Once I’ve organized this list, I’ll need everyone’s help:

* Make sure the list is complete and the categorization of open/closed topics is accurate
* Make sure your name is tagged on the topics you care about.

After that, I’ll merge this PR (clearly marked as "WIP") and open issues for each of the open topics so that we can have a more focused conversation. As each open topic is resolved, we will make the appropriate edits as part of a new PR. My main goal here is to make sure all the great conversations here are captured and that we can make incremental progress on getting this into the Spec.

smolinari:
@paralin 

> What's the difference of a live query to the spec? You're still specifying what data is required in the same syntax.

Somehow a query needs to be designated as "live" by the client. Or, as you said, and I think we are on the same wavelength on this, the queries that can be called as "live" need to be defined on the server in some way and be obvious/ findable through introspection.   

Scott

robzhu:
### Closed
* Add clearer notice that the RFC is a work-in-progress.
* Add backward arrow from GraphQL to Subscriptions box in diagram (@rmosolgo)
* Make "subscription active" part of the lifecycle optional and recommended for client-side error-handling (@taion, @stubailo)

### Open
* Do we need a "bi-directional" channel between client and server? 
@taion @stubailo @OlegIlyenko 
* Where should subscription state be stored? (inside GraphQL-js or as a separate component?)
@theorygeek @taion @stubailo @rmosolgo @jamesgorman2 
* How should we pass event payloads/data into the subscription resolvers? 
@stubailo @taion @leebyron @dschafer 
* Re-model the concept of "events" as an abstract "event stream". 
@dschafer @leebyron @rmosolgo @taion @OlegIlyenko @stubailo 
* The "event stream" should be derived purely from the subscription root field (and ignore selection set). 
@dschafer @leebyron @rmosolgo @taion @OlegIlyenko @stubailo 
* Fetching initial state when subscribing. 
@stubailo @rzachariah @romandvoskin @taion 
* Re-define Live Queries? Can Live Queries make Subscriptions unnecessary? 
@smolinari @paralin @laneyk @taion @Siyfion @jamesgorman2 @leebyron 
* Intermediate Subscription event processing 
@rzachariah @romandvoskin 

Please let me know if I've missed your topic or if you would like to be added/removed from anything. I'll leave this post here for a bit so everyone tagged has a chance to review. After that, I'm going to merge this PR and file issues for all open topics.

taion:
>Do we need a "bi-directional" channel between client and server?
The use of SSE v WS is an application-level implementation detail. All that's needed is the server being able to push messages to the client somehow, which distinguishes this from request/response.

>Where should subscription state be stored? (inside GraphQL-js or as a separate component?)
I'd prefer to have this state be stored outside of GraphQL.js. I want to e.g. manage my own Redis subscriptions backing my GraphQL subscriptions, and I feel like the control logic is more straightforward if I have something I can call when my subscription fires, rather than passing e.g. an observable into GraphQL.js.

>How should we pass event payloads/data into the subscription resolvers?
This is a framework-level implementation question. I'd prefer GraphQL.js use the existing `rootValue` field here.

>The "event stream" should be derived purely from the subscription root field (and ignore selection set).
I agree with this. It's the "natural" way to write the spec.

>Fetching initial state when subscribing.
I think this is an application-level thing. "State" is not a concept that is inherent to GraphQL. Subscribing to e.g. streams of newly created items also doesn't necessarily admit an easy representation of current state. Imagine subscription updates that don't reflect idempotent operations – in such cases there isn't really a coherent initial state.

>Re-define Live Queries? Can Live Queries make Subscriptions unnecessary?
@stubailo's point is exactly right here. Live queries should be thought of separately from event-based subscriptions. They can be used to solve similar problems. However, scaling event-stream-based subscription systems is relatively well-understood; the same does not apply for live queries, and for anyone looking to ship something scalable now, live queries aren't really an option.

>Intermediate Subscription event processing
This is an application-level detail. If someone wants to subscribe to e.g. a "top X" stream but not re-process that stream, then there's nothing at the framework level that prevents e.g. the "trending tweets" subscription from using different events than the "new tweet" subscription.

paralin:
@taion Live queries certainly could be scalable, especially with GraphQL's emphasis on separating logic for each field. I'm looking at implementing a scheduler for [Magellan](http://github.com/rgraphql/magellan) that could do that.

But, you're right. Subscriptions are a different topic from live queries. And IMO, live queries don't need a change to the spec to be implemented. Directives are server-specific with introspection, and that should be all you need.

taion:
Just to clarify, when I say "live queries", I mean something like the `@live` directive mentioned in talks, where the interface presented to the client is one where the client can request a particular query be live, rather than one where the client explicitly requests specific streams.

For an example of a subscription-based API, see https://github.com/edvinerikson/relay-subscriptions/blob/v1.0.0/examples/todo/js/components/Todo.js#L124-L148.

If this were using live queries, then perhaps it would just look like:

```graphql
fragment on Todo @live {
  id
  complete
  text
  # ...
}
```

However, if you're doing various things such as using microservices and splitting up GraphQL objects across multiple backend services, &c., how to efficiently implement live queries can be much less than obvious, while event-based systems are much more straightforward.

leebyron:
Since there's excitement and support for the direction of this RFC, I'm going to commit it as an official tracked RFC. @robzhu just opened up issues for all of the various topics still under discussion - so this is still very much an open design!

stubailo:
So excited!

smolinari:
Me too! 👍 

Scott

thetwosents:
AHHHHHHHHHHHHH!!!!!!!

PR #270, opened 2017-02-26
Add missing description for '__EnumValue' Type

IvanGoncharov:
Spec is also missing descriptions for `__Schema` and `__DirectiveLocation` types.


leebyron:
Excellent work, many thanks for the careful read and careful fix of this accidental omission

IvanGoncharov:
It was my pleasure. We are working on a few GraphQL tools so rereading spec every day 😄

PR #287, opened 2017-03-15
Forbid alias names to start with `__`

IvanGoncharov:
In on of our application, we need to know the exact type of every object in response so we inject `__typename` into every selection. 
According to the current version of the spec, following query is valid:
```
{
   __typename: someField
}
```
So injected selection will look like:
```
{
   __typename: someField
   __typename
}
```
And will produce the following error:
```
"Fields \"__typename\" conflict because someField and __typename are different fields. Use different aliases on the fields to fetch both if this was intentional."
```
Based on [this line](https://github.com/facebook/relay/blame/45fd7206f8e003d283acf7ef5462731176cfd04f/CHANGELOG.md#L182) and a quick look through source code I suspect that Relay also injects `__typename` into queries and therefore is also affected by the same problem.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -212,7 +212,8 @@ query getName {
 **Explanatory Text**
 
 The target field of a field selection must be defined on the scoped type of the
-selection set. There are no limitations on alias names.
+selection set. Alias names must not begin with {"__"} (two underscores), as they
+can conflict with names used by GraphQL's introspection system.
 
 For example the following fragment would not pass validation:
 


calebmer:
Why is the conflict validation not enough? I appreciate the ability to start an alias with `__` from a client perspective. Because `__` is reserved in the schema a client may insert fields with aliases starting with `__` knowing that they won’t conflict.

For example, in a GraphQL client with no schema it may be useful to fetch type information at runtime like so:

```graphql
{
  search {
    ... on Human {
      firstName
    }
    ... on Droid {
      modelNumber
    }
  }
  __Human: __type(name: "Human") {
    possibleTypes {
      name
    }  
  }
  __Droid: __type(name: "Droid") {
    possibleTypes {
      name
    }  
  }
}
```

The `__` here allows us to avoid collisions with user defined fields. This approach was discussed in https://github.com/apollographql/apollo-client/issues/1337.

IvanGoncharov:
@calebmer 

> Why is the conflict validation not enough? 
> Because __ is reserved in the schema a client may insert fields with aliases starting with __ knowing that they won’t conflict.

Users can also specify aliases in their queries starting with `__` as I have shown in my initial example. Judging from [this test case](https://github.com/apollographql/apollo-client/blob/84db80208e7ecd7536eee4fb576cc6ed9349fb08/test/queryTransform.ts#L33) Apollo also uses `__typename` injecting technique.
So it is also affected by the same problem from my initial comment.

> For example, in a GraphQL client with no schema it may be useful to fetch type information at runtime like so:

What if I use `typename` as a name of a type ? (non-zero probability)
Will you generate an invalid query like below?
```
{
  search {
    ... on typename {
      direvedTypes
    }
    ... on fileldname {
      type
    }
  }
  __typename: __type(name: "typename") { # <-- result field __typename here
    possibleTypes {
      name
    }  
  }
  __fieldname: __type(name: "fieldname") {
    possibleTypes {
      name
    }  
  }
  __typename # <-- and here
}
```

In general, it's very concerning that developers use `__` prefix since it may clash with the current and future introspection fields. Why not just add some random prefix like `_apollo__`?

For example, our new project is a smart proxy which injects some additional fields into query before passing it to the original server. So what if we also decide to use `__` for our own aliases? It will make our tool incompatible with Apollo clients.

**So from purely theoretical this issue became very practical.**
I think we should ASAP discourage tooling authors from using `__` and switch to unique prefixes. Or we, as GraphQL community, will start experiencing interoperability issues in the future.

On a practical side: @calebmer can you please switch to unique prefix in the future version of Apollo?
\+ I can change the wording of this PR to say that using `__` in aliases is deprecated/discouraged and may result in errors in the future versions of spec.

stubailo:
(Note that the current version of Apollo does not do this - that was simply an example!)

IvanGoncharov:
@stubailo Thanks for the update.
So you don't use labels named `__<Something>` anywhere in Apollo?

To give example of potential interoperability issues lets use following scenario for GraphQL request:
`query in source code` => [VolcanoJS](https://github.com/telescopejs/telescope) => Apollo => proxy tool we are working on => GraphQL Server 

And anyone in this chain (including frontend developer) can come up with `__<Name>` alias thinking it's unique and safe to use.

IMHO it's still not too late to ban `__<name>` aliases until some major library/tool start using them.
At the same time, we can add the recommendation to use `_<tool/library name>__` pattern for naming tool/library specific types, fields, aliases, etc. What do you think?

stubailo:
> So you don't use labels named `__<Something>` anywhere in Apollo?

That's correct.

I think there's a decision here - does `__` refer to anything "internal" or specifically to reserved GraphQL stuff. And I guess now in the spec `__` is reserved for spec and introspection fields.

I can imagine, in the future, wanting to inject fields for stuff like cache invalidation, so having a convention to avoid collisions would be nice.

calebmer:
@IvanGoncharov I agree that if a client (or any tool) is going to inject an aliased field into a GraphQL query then it should be namespaced. However, I don’t think that’s relevant to this discussion as `_apollo_xyz` and `__apollo_xyz` are both currently options.

IvanGoncharov:
> And I guess now in the spec __ is reserved for spec and introspection fields.

I think the same way

> However, I don’t think that’s relevant to this discussion as _apollo_xyz and __apollo_xyz are both currently options.

@calebmer Agree 👍. Discussing naming schemes for vendor fields, types and aliases are beyond the scope of this PR.  However, it's very likely to treat `__<name>` as a unique name without adding any vendor prefix, and https://github.com/apollographql/apollo-client/issues/1337 is a good example of this cognitive bias.

Also, nothing is preventing GraphQL spec from adding a new introspection field with your unique prefix. In that sense `Apollo` is in the better situation comparing to some other projects with very common names like [graphql-server](https://github.com/apollographql/graphql-server) since `__graphql_server_<name>` or `__server_<name>` is something that may be added in the future versions of the spec.

Returning back to my initial comment.
@calebmer @stubailo Do you see any alternative solutions to the problem with conflicting `__typename`s?

```
{
   __typename: someField # Alias inside user query
   __typename # Injected by Apollo, Relay or other tool
}
```

stubailo:
I think the above problem is something that could be solved when it comes up.

leebyron:
I don't see this as a GraphQL-wide problem, only specifically a problem for individual client frameworks to be cognizant of. Specifically if a client framework decides to automatically include a field, that they run the risk of colliding with a field added by the developer.

The original example posted here is a good example of this happening:

```graphql
{
   __typename: someField
   __typename
}
```

But note that it's not the only example of this happening. For example if a client wants to include an `ID` type field in an attempt to create client side cache keys, it might include a field that conflicts in the same way, but does not start with `__`:

```
{
  id: someField
  id
}
```

Therefore I don't think this issue is specifically related to the `__` prefix. Instead it is a concern with client frameworks injecting potentially colliding fields. I don't think that a change to the spec helps with that concern. Instead I think client frameworks like Apollo or Relay just need to produce clear and helpful error messages when a developer writes a query which collides with those frameworks' own limitations.

I'm inclined to close this issue, but am open to other points of view before doing so

stubailo:
I think a smart enough client framework can always generate non-conflicting fields, so it's not really an issue from that perspective. Apollo does inject `__typename` if it doesn't exist, but it could just as easily alias that to `_____asdfasdsdafasdf___unique: __typename` and always pick a key that doesn't conflict.

IvanGoncharov:
@leebyron Thank you for great example with `id` 👍 I agree that this problem is more general than `__typename`. So I will close this PR. 
However I still think that there should be some naming convention for injected fields. But this is a topic for a separate issue.

leebyron:
> However I still think that there should be some naming convention for injected fields. But this is a topic for a separate issue.

I agree with you that client frameworks should be aware of the potential for collision when injecting fields. I also agree that is a separate topic - likely specific to each client framework to be discussed per-project. A naming convention is one possible solution to this problem, but not the only one or necessarily the best one when considering the context. For example, it could also be reasonable to simply make something like `__typename: someField` produce an error during build.

PR #291, opened 2017-04-01
Use sub-bullets for sub-classes of subscriptions

h3h:
Just update the Markdown to make the sub-classes clearer.


wincent:
Thanks @h3h. Good catch.

Issue #292, opened 2017-04-02
Specification Clarification: Multiple Operations for Query/Mutations 


mhamrah:
The spec states in the language section: ```When submitting a Document with multiple operations to a GraphQL service, the name of the desired operation to be executed must also be provided.``` and for Execution: ```...and a selected operation name to run if the document defines multiple operations.``` but I cannot find where the caller should specify this operation name in the request. Is it explicitly specified, or is it up to the implementation to determine the method for specifying the operation name?

IvanGoncharov:
> is it up to the implementation to determine the method for specifying the operation name?

@mhamrah Correct.
From here: http://facebook.github.io/graphql/#sec-Executing-Requests
> To execute a request, the executor must have a parsed Document (as defined in the “Query Language” part of this spec) and a **selected operation name** to run if the document defines multiple operations, 

There is also a recommendation on how it should be implemented over HTTP:
http://graphql.org/learn/serving-over-http/#post-request

wincent:
Thanks for answering, @IvanGoncharov. If you still have any questions, @mhamrah, please feel free to comment again.

PR #293, opened 2017-04-03
Clarify how operation name is handled in the spec.

mhamrah:
As #292 I believe it will be beneficial to clarify that
specifying the operation name when handling multiple operations
is up to the implementation, not the query language.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -184,7 +184,7 @@ represented in the shorthand form, which omits both the query keyword and
 operation name. Otherwise, if a GraphQL query document contains multiple
 operations, each operation must be named. When submitting a query document with
 multiple operations to a GraphQL service, the name of the desired operation to
-be executed must also be provided.
+be executed must also be provided. How the name is provided is up to the GraphQL implementation.
 
 
 ## Operations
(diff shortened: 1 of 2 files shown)


spec/Section 6 -- Execution.md
```diff
@@ -22,7 +22,7 @@ to be formatted according to the Response section below.
 To execute a request, the executor must have a parsed `Document` (as defined
 in the “Query Language” part of this spec) and a selected operation name to
 run if the document defines multiple operations, otherwise the document is
-expected to only contain a single operation. The result of the request is
+expected to only contain a single operation. How this information is passed to the executor is determined by the implementation. The result of the request is
```
leebyron: I think this addition is redundant and potential confusing. Of course this is the case, by definition an implementation's role is to implement the below algorithm, which describes the arguments which are expected to be provided

spec/Section 2 -- Language.md
```diff
@@ -184,7 +184,7 @@ represented in the shorthand form, which omits both the query keyword and
 operation name. Otherwise, if a GraphQL query document contains multiple
 operations, each operation must be named. When submitting a query document with
 multiple operations to a GraphQL service, the name of the desired operation to
-be executed must also be provided.
+be executed must also be provided. How the name is provided is up to the GraphQL implementation.
```
leebyron: This line is confusing - no where else do we specify that the information required to perform an execution needs to be provided by the implementation. For example, how the graphql document is provided is also defined by the implementation.

Perhaps to help alleviate the concern you mentioned, this should instead make it clear that the name should be provided in addition to the document. Suggestion:

When submitting a query document with multiple operations to a GraphQL service, the name of the desired operation to be executed must be provided *in addition to the document*.

PR #305, opened 2017-05-02
Spec: Formal Subscriptions Definition

robzhu:
This PR contains my proposal for formally describing GraphQL Subscriptions in the Spec. It will be accompanied by a PR against GraphQL-js (will provide link when that PR is available).

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -2,7 +2,7 @@
 
 Clients use the GraphQL query language to make requests to a GraphQL service.
 We refer to these request sources as documents. A document may contain
-operations (queries and mutations are both operations) as well as fragments, a
+operations (queries, mutations, and subscriptions) as well as fragments, a
 common unit of composition allowing for query reuse.
 
 A GraphQL document is defined as a syntactic grammar where terminal symbols are
@@ -193,12 +193,14 @@ OperationDefinition :
   - OperationType Name? VariableDefinitions? Directives? SelectionSet
   - SelectionSet
 
-OperationType : one of `query` `mutation`
+OperationType : one of `query` `mutation` `subscription`
 
-There are two types of operations that GraphQL models:
+There are three types of operations that GraphQL models:
 
   * query - a read-only fetch.
   * mutation - a write followed by a fetch.
+  * subscription - a long-lived request that returns data whenever a domain
+    event triggers.
 
 Each operation is represented by an optional operation name and a selection set.
 
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -199,8 +199,8 @@ There are three types of operations that GraphQL models:
 
   * query - a read-only fetch.
   * mutation - a write followed by a fetch.
-  * subscription - a long-lived request that returns data whenever a domain
-    event triggers.
+  * subscription - a long-lived request that fetches data in response to source
+    events.
 
 Each operation is represented by an optional operation name and a selection set.
 
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -198,6 +198,70 @@ query getName {
 }
 ```
 
+### Subscription Operation Definitions
+
+#### Single root field
+
+**Formal Specification**
+
+  * For each subscription operation definition {subscription} in the document
+  * Let {rootFields} be the first level of fields on the operation's selection set.
+    * {rootFields} must be a set of one.
+
+**Explanatory Text**
+
+Subscription operation must have exactly one root field.
+
+Valid examples:
+
+```graphql
+subscription sub {
+  newMessage {
+    body
+    sender
+  }
+}
+```
+
+```graphql
+fragment newMessageFields on Message {
+  body
+  sender
+}
+
+subscription sub {
+  newMessage {
+    ... newMessageFields  
+  }
+}
+```
+
+Invalid:
+
+```!graphql
+subscription sub {
+  newMessage {
+    body
+    sender
+  }
+  # a second root field will cause this operation to fail validation
+  alarms {
+    name
+    snoozeCount
+  }
+}
+```
+
+```!graphql
+subscription sub {
+  newMessage {
+    body
+    sender
+  }
+  # this also applies to introspection fields
+  __typename
+}
+```
 
 ## Fields
 
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -210,7 +210,7 @@ query getName {
 
 **Explanatory Text**
 
-Subscription operation must have exactly one root field.
+Subscription operations must have exactly one root field.
 
 Valid examples:
 
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -244,13 +244,11 @@ subscription sub {
     body
     sender
   }
-  # a second root field will cause this operation to fail validation
-  alarms {
-    name
-    snoozeCount
-  }
+  disallowedSecondRootField
 }
 ```
+
+Introspection fields are counted. The following example is also invalid:
 
 ```!graphql
 subscription sub {
@@ -258,7 +256,6 @@ subscription sub {
     body
     sender
   }
-  # this also applies to introspection fields
   __typename
 }
 ```
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -205,7 +205,7 @@ query getName {
 **Formal Specification**
 
   * For each subscription operation definition {subscription} in the document
-  * Let {rootFields} be the first level of fields on the operation's selection set.
+  * Let {rootFields} be the top level selection set on {subscription}.
     * {rootFields} must be a set of one.
 
 **Explanatory Text**
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -169,23 +169,21 @@ which are no longer necessary.
 #### Subscribe
 
 Executing a subscription creates a persistent function on the server that
-maps an underlying event stream to the Publish Stream. The logic to create the
-underlying event stream is application-specific and takes the root field and
-query variables as inputs.
+maps an underlying Source stream to the Publish Stream. The logic to create the
+Source stream is application-specific and takes the root field and query
+variables as inputs.
 
-Subscribe(schema, subscription, operationName, variableValues, initialValue):
+Subscribe(subscription, schema, variableValues, initialValue):
+
+  * Let {sourceStream} be the result of running {CreateSourceEventStream(subscription, schema, variableValues, initialValue)}.
+  * Let {responseStream} be the result of running {MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues)}
+  * Return {responseStream}.
+
+CreateSourceEventStream(subscription, schema, variableValues, initialValue):
 
   * Let {subscriptionType} be the root Subscription type in {schema}.
   * Assert: {subscriptionType} is an Object type.
   * Let {selectionSet} be the top level Selection Set in {subscription}.
-  * Let {sourceStream} be the result of running {CreateSourceEventStream(schema, subscription, selectionSet, variableValues, initialValue)}.
-  * Let {responseStream} be the result of running
-    {MapSourceToResponseEvent(sourceStream)}
-  * Return {responseStream}.
-
-CreateSourceEventStream(schema, subscriptionType, selectionSet, variableValues,
-  initialValue):
-
   * Let {rootField} be the first top level field in {selectionSet}.
   * Let {argumentValues} be the result of {CoerceArgumentValues(subscriptionType, rootField, variableValues)}.
   * Let {fieldStream} be the result of running {ResolveFieldEventStream(subscriptionType, initialValue, rootField, argumentValues)}.
@@ -201,31 +199,38 @@ ResolveFieldEventStream(subscriptionType, rootValue, fieldName, argumentValues):
 Each event in the underlying event stream triggers execution of the subscription
 selection set.
 
-MapSourceToResponseEvent(sourceStream, subscriptionType, selectionSet, variableValues):
+MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
 
-  * Return a new event stream {publishStream} which yields events as follows:
+  * Return a new event stream {responseStream} which yields events as follows:
   * For each {event} on {sourceStream}:
-    * Let {publishPayload} be the result of running
-      {ExecuteSelectionSet(selectionSet, subscriptionType, event, variableValues)}
-      *normally* (allowing parallelization).
-    * Let {errors} be any *field errors* produced while executing the
-      selection set.
-    * Let {response} be an unordered map containing {publishPayload} and, optionally,
-      {errors}.
+    * Let {response} be the result of running
+      {ExecuteSubscriptionEvent(subscription, schema, variableValues, event)}.
     * Yield an event containing {response}.
 
-Note: in large scale subscription systems, ExecuteSelectionSet and Subscribe
-algorithm may be run on separate services to maintain predictable scaling
-properties. See the section below on Supporting Subscriptions at Scale.
+ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
+
+  * Let {subscriptionType} be the root Subscription type in {schema}.
+  * Assert: {subscriptionType} is an Object type.
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {data} be the result of running
+    {ExecuteSelectionSet(selectionSet, subscriptionType, initialValue, variableValues)}
+    *normally* (allowing parallelization).
+  * Let {errors} be any *field errors* produced while executing the
+    selection set.
+  * Return an unordered map containing {data} and {errors}.
+
+Note: in large scale subscription systems, the {ExecuteSubscriptionEvent} and
+{Subscribe} algorithms may be run on separate services to maintain predictable
+scaling properties. See the section below on Supporting Subscriptions at Scale.
 
 #### Unsubscribe
 
-Unsubscribe cancels the Publish Stream. This is also a good opportunity for the
+Unsubscribe cancels the Response Stream. This is also a good opportunity for the
 server to clean up the underlying event stream and any other resources used by
 the subscription. Here are some example cases in which to Unsubscribe: client
 no longer wishes to receive payloads for a subscription; the source event stream
 produced an error or naturally ended; the server encountered an error during
-ExecuteSelectionSet.
+ExecuteSubscriptionEvent.
 
 Unsubscribe()
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -154,7 +154,8 @@ the following capabilities:
   * **must** support observation of the associated publish stream (for example,
   via iteration, callbacks, or reactive semantics).
   * **must** support cancellation of the subscription (aka unsubscribe).
-  * **must** support a unique mapping to the subscriber (for example, a unique Id).
+  * **must** support a way to identify the subscriber/subscription pair (for
+    example, a GUID/callback table or closure over the callback).
   * **may** include an initial response associated with executing the selection
   set defined on the subscription operation.
 
@@ -182,11 +183,16 @@ Publish(subscription, schema, variableValues, payload, publishStream)
       {publishStream}.
 
 ### Unsubscribe
+The unsubscribe operation can be implemented in a number of ways. For example,
+by using a dedicated subscription manager, defining it as a method on the
+subscription object, or cancelling the iterator.
+
 Unsubscribe()
 
   * Let {publishStream} be a mapping of {null}
+  * Terminate and clean up {eventStream}
 
-### Recommendations and Considerations for supporting Subscriptions
+### Recommendations and Considerations for Supporting Subscriptions
 Supporting subscriptions is a large change for any GraphQL server. Query and
 mutation operations are stateless, allowing scaling via cloning GraphQL server
 instances. Subscriptions, by contrast, are stateful. The pieces of state for a
@@ -202,10 +208,11 @@ We recommend thinking about the behavior of your system when this state is lost
 due to single-node failures. We can improve durability and availability by using
 dedicated sub-systems to manage this state. For example, event streams can be
 built using modern pub-sub systems, and client channels can be handled with a
-dedicate client gateway.
+dedicated client gateway.
 
-Rather than mixing stateful and stateless systems, we recommend keeping the
-GraphQL server stateless and delegating all state persistence these sub-systems.
+Rather than mixing stateless (queries and mutations) and stateful
+(subscriptions), we recommend keeping the GraphQL server stateless and
+delegating all state persistence these sub-systems.
 
 ## Executing Selection Sets
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -166,15 +166,19 @@ Subscribe(subscription, schema, variableValues, initialValue):
   * Assert: {subscriptionType} is an Object type.
   * Let {selectionSet} be the top level Selection Set in {subscription}.
   * Let {rootField} be the first top level field in {selectionSet}.
-  * Let {eventStream} be a domain-specific transform of {rootField, variableValues}
+  * Let {eventStream} be the result of running {MapSubscriptionEvents(rootField, variableValues)}.
   * Let {publishStream} be a mapping of {eventStream} where each {event} is
     mapped via Publish().
+
+MapSubscriptionEvents(rootField, variableValues):
+
+  * *Application-specific logic to map from root field/variables to events*
 
 ### Publish
 
 Publish(subscription, schema, variableValues, payload, publishStream)
 
-  * For each {event, payload} in {eventStream}
+  * For each {eventStream} as {event} and {payload}:
     * Let {data} be the result of running
       {ExecuteSelectionSet(selectionSet, subscriptionType, payload, variableValues)}
       *normally* (allowing parallelization).
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -103,8 +103,7 @@ Note: This algorithm is very similar to {CoerceArgumentValues()}.
 ## Executing Operations
 
 The type system, as described in the “Type System” section of the spec, must
-provide a query root object type. If mutations/subscriptions are supported, it must also
-provide a mutation/subscription root object type, respectively.
+provide a query root object type. If mutations or subscriptions are supported, it must also provide a mutation and subscription root object type, respectively.
 
 ### Query
 
@@ -149,9 +148,9 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
 
 ### Subscription
 
-Unlike queries and mutations, subscriptions have a lifetime that consists of three
-phases. Between the the subscribe and unsubscribe phases, the subscription is
-considered to be in the "active" state.
+If the operation is a subscription, the result of the operation is the creation
+of a persistent object on the server that exists for the lifetime of the
+subscription.
 
 #### Subscribe
 
@@ -160,7 +159,7 @@ the following capabilities:
 
   * **Must** support observation of the associated publish stream (for example,
   via iteration, callbacks, or reactive semantics).
-  * **Must** support cancellation of the subscription, see {Unsubscribe}.
+  * **Must** support cancellation of the observation, see {Unsubscribe}.
   * **Must** support a way to send data to the subscriber.
   * **May** include an initial response associated with executing the selection
   set defined on the subscription operation.
@@ -189,7 +188,7 @@ MapSubscriptionEvents(rootField, variableValues):
 
 #### Publish
 
-Once a subscription is active, we listen for events on its event stream. Each
+Once a subscription is created, we listen for events on its event stream. Each
 event carries an optional payload, which we combine with the arguments from
 Subscribe() to resolve the selection set.
 
@@ -215,6 +214,35 @@ Unsubscribe()
   * Terminate {publishStream} (For example, cancel iteration, detach mapping function)
   * Terminate and clean up {eventStream}
 
+#### Example
+
+As an example, consider a chat application. To subscribe to new messages posted
+to the chat room, the client sends a request like so:
+
+```graphql
+subscription NewMessages {
+  newMessage(roomId: 123) {
+    sender
+    text
+  }
+}
+```
+
+While the client is subscribe, whenever new messages are posted to chat room
+with ```ID 123```, the selection for ```sender``` and ```text``` will be
+evaluated and published to the client, for example:
+
+```js
+{
+  "data": {
+    "newMessage": {
+      "sender": "Hagrid",
+      "text": "You're a wizard!"
+    }
+  }
+}
+```
+
 #### Recommendations and Considerations for Supporting Subscriptions
 
 Supporting subscriptions is a large change for any GraphQL server. Query and
@@ -231,12 +259,14 @@ subscription are:
 We recommend thinking about the behavior of your system when this state is lost
 due to single-node failures. We can improve durability and availability by using
 dedicated sub-systems to manage this state. For example, event streams can be
-built using modern pub-sub systems, and client channels can be handled with a
-dedicated client gateway.
+built using modern pub-sub systems, and payload delivery to clients can be
+handled by a dedicated client gateway tier.
 
-Rather than mixing stateless (queries and mutations) and stateful
-(subscriptions), we recommend keeping the GraphQL server stateless and
-delegating all state persistence to these sub-systems.
+For systems with high capacity, availability, and durability requirements, we
+recommend keeping the GraphQL server stateless and delegating all state
+persistence to sub-systems that are designed for stateful scaling. Note that
+subscription types are still defined in the original schema along with queries 
+and mutations.
 
 ## Executing Selection Sets
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -229,8 +229,8 @@ subscription NewMessages {
 ```
 
 While the client is subscribe, whenever new messages are posted to chat room
-with ```ID 123```, the selection for ```sender``` and ```text``` will be
-evaluated and published to the client, for example:
+with ID "123", the selection for "sender" and "text" will be evaluated and
+published to the client, for example:
 
 ```js
 {
@@ -265,7 +265,7 @@ handled by a dedicated client gateway tier.
 For systems with high capacity, availability, and durability requirements, we
 recommend keeping the GraphQL server stateless and delegating all state
 persistence to sub-systems that are designed for stateful scaling. Note that
-subscription types are still defined in the original schema along with queries 
+subscription types are still defined in the original schema along with queries
 and mutations.
 
 ## Executing Selection Sets
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -155,14 +155,27 @@ If the operation is a subscription, the result is an event stream called the
 "Response Stream" where each event in the event stream is the result of
 executing the operation for each new event on an underlying "Source Stream".
 
+Executing a subscription creates a persistent function on the server that
+maps an underlying Source Stream to a returned Response Stream.
+
+Subscribe(subscription, schema, variableValues, initialValue):
+
+  * Let {sourceStream} be the result of running {CreateSourceEventStream(subscription, schema, variableValues, initialValue)}.
+  * Let {responseStream} be the result of running {MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues)}
+  * Return {responseStream}.
+
+Note: In large scale subscription systems, the {Subscribe} and {ExecuteSubscriptionEvent} 
+algorithms may be run on separate services to maintain predictable scaling 
+properties. See the section below on Supporting Subscriptions at Scale.
+
 An event stream represents a sequence of discrete events over time which can be
 observed. As an example, a "Pub-Sub" system may produce an event stream when
 "subscribing to a topic", with an event occurring on that event stream for each
 "publish" to that topic. Event streams may produce an infinite sequence of
 events or may complete at any point. Event streams may complete in response to
 an error or simply because no more events will occur. An observer may at any
-point decide to stop observing an event stream, after which it must receive no
-more events from that event stream.
+point decide to stop observing an event stream by cancelling it, after which it 
+must receive no more events from that event stream.
 
 As an example, consider a chat application. To subscribe to new messages posted
 to the chat room, the client sends a request like so:
@@ -207,18 +220,11 @@ single machine in a service. Durability and availability may be improved by
 having separate dedicated services for managing subscription state and client
 connectivity.
 
-#### Subscribe
+#### Source Stream
 
-Executing a subscription creates a persistent function on the server that
-maps an underlying Source stream to the Response Stream. The logic to create the
-Source stream is application-specific and takes the root field and query
-variables as inputs.
-
-Subscribe(subscription, schema, variableValues, initialValue):
-
-  * Let {sourceStream} be the result of running {CreateSourceEventStream(subscription, schema, variableValues, initialValue)}.
-  * Let {responseStream} be the result of running {MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues)}
-  * Return {responseStream}.
+A Source Stream represents the sequence of events, each of which will 
+trigger a GraphQL execution corresponding to that event. Like field value
+resolution, the logic to create a Source Stream is application-specific.
 
 CreateSourceEventStream(subscription, schema, variableValues, initialValue):
 
@@ -232,16 +238,17 @@ CreateSourceEventStream(subscription, schema, variableValues, initialValue):
 
 ResolveFieldEventStream(subscriptionType, rootValue, fieldName, argumentValues):
   * Let {resolver} be the internal function provided by {subscriptionType} for
-    determining the resolved value of a field named {fieldName}.
+    determining the resolved event stream of a subscription field named {fieldName}.
   * Return the result of calling {resolver}, providing {rootValue} and {argumentValues}.
 
-Note: this algorithm is intentionally similar to {ResolveFieldValue} to enable
-consistency when defining resolvers on any operation type.
+Note: This {ResolveFieldEventStream} algorithm is intentionally similar 
+to {ResolveFieldValue} to enable consistency when defining resolvers 
+on any operation type.
 
 #### Response Stream
 
-Each event in the underlying event stream triggers execution of the subscription
-selection set.
+Each event in the underlying Source Stream triggers execution of the subscription
+selection set using that event as a root value.
 
 MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
 
@@ -250,6 +257,7 @@ MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
     * Let {response} be the result of running
       {ExecuteSubscriptionEvent(subscription, schema, variableValues, event)}.
     * Yield an event containing {response}.
+  * When {responseStream} completes: complete this event stream.
 
 ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
 
@@ -263,22 +271,17 @@ ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
     selection set.
   * Return an unordered map containing {data} and {errors}.
 
-Note: in large scale subscription systems, the {ExecuteSubscriptionEvent} and
-{Subscribe} algorithms may be run on separate services to maintain predictable
-scaling properties. See the section above on Supporting Subscriptions at Scale.
-This algorithm is intentionally similar to {ExecuteQuery} since this is where
-the subscription's selection set is executed.
+Note: The {ExecuteSubscriptionEvent} algorithm is intentionally similar to 
+{ExecuteQuery} since this is how the each event result is produced.
 
 #### Unsubscribe
 
-Unsubscribe cancels the Response Stream. This is also a good opportunity for the
-server to clean up the underlying event stream and any other resources used by
-the subscription. Here are some example cases in which to Unsubscribe: client
-no longer wishes to receive payloads for a subscription; the source event stream
-produced an error or naturally ended; the server encountered an error during
-{ExecuteSubscriptionEvent}.
+Unsubscribe cancels the Response Stream when a client no longer wishes to receive 
+payloads for a subscription. This may in turn also cancel the Source Stream.
+This is also a good opportunity to clean up any other resources used by
+the subscription.
 
-Unsubscribe()
+Unsubscribe(responseStream)
 
   * Cancel {responseStream}
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -106,7 +106,7 @@ Note: This algorithm is very similar to {CoerceArgumentValues()}.
 
 The type system, as described in the “Type System” section of the spec, must
 provide a query root object type. If mutations or subscriptions are supported,
-it must also provide a mutation and subscription root object type, respectively.
+it must also provide a mutation or subscription root object type, respectively.
 
 ### Query
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -148,30 +148,19 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
 
 ### Subscription
 
-If the operation is a subscription, the result of the operation is the creation
-of a persistent object on the server that exists for the lifetime of the
-subscription.
+We define an event stream as a sequence of discrete events over time that can be
+observed. An observer of an event stream may cancel observation to avoid stop
+future events.
+
+If the operation is a subscription, the result is an event stream called the
+"Publish Stream" where each event in the stream is called a "Publish Payload".
 
 #### Subscribe
 
-The result of executing a subscription operation is a subscription object with
-the following capabilities:
-
-  * **Must** support observation of the associated publish stream (for example,
-  via iteration, callbacks, or reactive semantics).
-  * **Must** support cancellation of the observation, see {Unsubscribe}.
-  * **Must** support a way to send data to the subscriber.
-  * **May** include an initial response associated with executing the selection
-  set defined on the subscription operation.
-
-Most subscriptions can only be evaluated when event data is available. For
-example, a subscription tells us when users log on and off would require event
-data that tells us *who* logged on or off. Without this event data, the
-subscription cannot be evaluated. However, it is possible to define
-subscriptions that can be evaluated without event data (for example, the current
-time on the server, synthetically triggered every second). Subscriptions that do
-not require event data for evaluation can optionally return an initial response
-to the Subscribe() operation.
+Executing a subscription creates a persistent function on the server that
+maps an underlying event stream to the Publish Stream. The logic to create the
+underlying event stream is domain-specific and takes the root field and query
+variables as inputs.
 
 Subscribe(schema, subscription, operationName, variableValues, initialValue):
 
@@ -179,40 +168,44 @@ Subscribe(schema, subscription, operationName, variableValues, initialValue):
   * Assert: {subscriptionType} is an Object type.
   * Let {selectionSet} be the top level Selection Set in {subscription}.
   * Let {rootField} be the first top level field in {selectionSet}.
-  * Let {eventStream} be the result of running {MapSubscriptionEvents(rootField, variableValues)}.
-  * Optionally: let {data} be the result of running {ExecuteRequest(schema, document, operationName, variableValues, initialValue)}, and return {data} on the subscription object.
+  * Let {eventStream} be the result of running {CreateUnderlyingEventStream(rootField, variableValues)}.
+  * Let {publishStream} be the result of running {MapEventToPayload(eventStream)}
 
-MapSubscriptionEvents(rootField, variableValues):
+CreateUnderlyingEventStream(rootField, variableValues):
 
   * *Application-specific logic to map from root field/variables to events*
 
-#### Publish
+#### Publish Stream
 
-Once a subscription is created, we listen for events on its event stream. Each
-event carries an optional payload, which we combine with the arguments from
-Subscribe() to resolve the selection set.
+Each event in the underlying event stream triggers execution of the subscription
+selection set.
 
-  * For each {event} and {eventData} on {eventStream}:
-    * Let {data} be the result of running
-      {ExecuteSelectionSet(selectionSet, subscriptionType, eventData, variableValues)}
+MapEventToPayload(eventStream):
+
+  * For each {event} on {eventStream}:
+    * Let {publishPayload} be the result of running
+      {ExecuteSelectionSet(selectionSet, subscriptionType, event, variableValues)}
       *normally* (allowing parallelization).
     * Let {errors} be any *field errors* produced while executing the
       selection set.
-    * If {errors} is not empty, optionally, Unsubscribe() the subscription.
-    * Yield an unordered map containing {data} and, optionally, {errors} on
-      {publishStream}.
-  * Let {publishStream} be the sequence of outputs yielded above.
+    * Yield an unordered map containing {publishPayload} and, optionally,
+      {errors} on {publishStream}.
+  * At any time while the publish stream is active, the client or server may
+  Unsubscribe().
+
+Common reasons for Unsubscribe() include:
+  * client no longer wants to receive payloads for a subscription.
+  * The underlying event stream has produced an error or has naturally ended.
 
 #### Unsubscribe
 
-The unsubscribe operation can be implemented in a number of ways. For example,
-by using a dedicated subscription manager, defining it as a method on the
-subscription object, or cancelling the iterator.
+Unsubscribe cancels the Publish Stream. This is also a good opportunity for the
+server to clean up the underlying event stream and any other resources used by
+the subscription.
 
 Unsubscribe()
 
-  * Terminate {publishStream} (For example, cancel iteration, detach mapping function)
-  * Terminate and clean up {eventStream}
+  * Cancel {publishStream}
 
 #### Example
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -148,6 +148,7 @@ phases. Between the the subscribe and unsubscribe phases, the subscription is
 considered to be in the "active" state.
 
 ### Subscribe
+
 The result of executing a subscription operation is a subscription object with
 the following capabilities:
 
@@ -170,6 +171,7 @@ Subscribe(subscription, schema, variableValues, initialValue):
     mapped via Publish().
 
 ### Publish
+
 Publish(subscription, schema, variableValues, payload, publishStream)
 
   * For each {event, payload} in {eventStream}
@@ -183,6 +185,7 @@ Publish(subscription, schema, variableValues, payload, publishStream)
       {publishStream}.
 
 ### Unsubscribe
+
 The unsubscribe operation can be implemented in a number of ways. For example,
 by using a dedicated subscription manager, defining it as a method on the
 subscription object, or cancelling the iterator.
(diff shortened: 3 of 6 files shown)


spec/Section 5 -- Validation.md
```diff
@@ -198,6 +198,70 @@ query getName {
...
+
+Invalid:
+
+```!graphql
+subscription sub {
+  newMessage {
+    body
+    sender
+  }
+  # a second root field will cause this operation to fail validation
```
wincent: Everywhere else in this file we rely on naming rather than comments to illustrate counterexamples. Perhaps we should do the same for consistency here. (ie. call the field `disallowedSecondRootField`)?

spec/Section 5 -- Validation.md
```diff
@@ -198,6 +198,70 @@ query getName {
...
+```
+
+```!graphql
+subscription sub {
+  newMessage {
+    body
+    sender
+  }
+  # this also applies to introspection fields
+  __typename
```
wincent: Not clear what we would name this one if we wanted to get rid of the comment.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
     selection set.
   * Return an unordered map containing {data} and {errors}.
```
wincent: Putting a comment here because silly GitHub won't let me comment on the lines above. Up on line 106-107 it says:

> If mutations are supported, it must also provide a mutation root object type.

That should probably say:

> If mutations or subscriptions are supported, it must also provide a mutation and subscription root object type, respectively.

Or words to that effect.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
 
+Unlike queries and mutations, subscriptions have a lifetime that consists of three
+phases. Between the the subscribe and unsubscribe phases, the subscription is
+considered to be in the "active" state.
+
+### Subscribe
```
wincent: We should probably add new subheadings to this section:

```
## Executing Operations

...

### Executing Queries

...

### Executing Mutations

...

### Executing Subscriptions

...

#### Subscribe

(etc)
```

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+The result of executing a subscription operation is a subscription object with
+the following capabilities:
+
+  * **must** support observation of the associated publish stream (for example,
+  via iteration, callbacks, or reactive semantics).
+  * **must** support cancellation of the subscription (aka unsubscribe).
+  * **must** support a way to identify the subscriber/subscription pair (for
+    example, a GUID/callback table or closure over the callback).
+  * **may** include an initial response associated with executing the selection
```
wincent: Capitalize "m" in "must"/"may" here for consistency with the other lists (that all seem to start with capitals)?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+considered to be in the "active" state.
+
+### Subscribe
+
+The result of executing a subscription operation is a subscription object with
+the following capabilities:
+
+  * **must** support observation of the associated publish stream (for example,
+  via iteration, callbacks, or reactive semantics).
+  * **must** support cancellation of the subscription (aka unsubscribe).
```
wincent: "aka" is a bit informal: I'd suggest, "that is," instead.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+phases. Between the the subscribe and unsubscribe phases, the subscription is
+considered to be in the "active" state.
+
+### Subscribe
+
+The result of executing a subscription operation is a subscription object with
+the following capabilities:
+
+  * **must** support observation of the associated publish stream (for example,
+  via iteration, callbacks, or reactive semantics).
```
wincent: Not sure what "iteration" means in the context of observation: do you specifically mean async iteration? (and if so, any concern that that is allowing JS-specific terminology to leak into the spec?)

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * **must** support cancellation of the subscription (aka unsubscribe).
+  * **must** support a way to identify the subscriber/subscription pair (for
+    example, a GUID/callback table or closure over the callback).
+  * **may** include an initial response associated with executing the selection
+  set defined on the subscription operation.
+
+Subscribe(subscription, schema, variableValues, initialValue):
```
wincent: `initialValue` is not referenced anywhere in this section.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+MapSubscriptionEvents(rootField, variableValues):
+
+  * *Application-specific logic to map from root field/variables to events*
+
+### Publish
+
+Publish(subscription, schema, variableValues, payload, publishStream)
+
+  * For each {eventStream} as {event} and {payload}:
```
wincent: Where does `payload` come from? On line 169 above it says that `eventStream` is "the result" of running `MapSubscriptionEvents`, but the "result" isn't defined anywhere.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+    * Let {data} be the result of running
+      {ExecuteSelectionSet(selectionSet, subscriptionType, payload, variableValues)}
+      *normally* (allowing parallelization).
+    * Let {errors} be any *field errors* produced while executing the
+      selection set.
+    * If {errors} is not empty, optionally terminate the subscription.
```
wincent: Should we define what termination means?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+### Unsubscribe
+
+The unsubscribe operation can be implemented in a number of ways. For example,
+by using a dedicated subscription manager, defining it as a method on the
+subscription object, or cancelling the iterator.
+
+Unsubscribe()
+
+  * Let {publishStream} be a mapping of {null}
```
wincent: I'm not sure what this means.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+Supporting subscriptions is a large change for any GraphQL server. Query and
+mutation operations are stateless, allowing scaling via cloning GraphQL server
+instances. Subscriptions, by contrast, are stateful. The pieces of state for a
+subscription are:
+
+  * Subscriber/client channel
+  * Subscription document
+  * Variables
+  * Execution context (for example, current logged in user, locale, etc.)
```
wincent: s/logged in/logged-in/

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +143,84 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+We recommend thinking about the behavior of your system when this state is lost
+due to single-node failures. We can improve durability and availability by using
+dedicated sub-systems to manage this state. For example, event streams can be
+built using modern pub-sub systems, and client channels can be handled with a
+dedicated client gateway.
+
+Rather than mixing stateless (queries and mutations) and stateful
+(subscriptions), we recommend keeping the GraphQL server stateless and
+delegating all state persistence these sub-systems.
```
wincent: these sub-systems → to these sub-systems

spec/Section 5 -- Validation.md
robzhu: Added a description above.

spec/Section 6 -- Execution.md
robzhu: I was referring to the general iterator pattern (async iterators also count). For example, Unity handles async operations via iterators/generators, so I don't think it's JS-specific.

spec/Section 6 -- Execution.md
robzhu: good catch, this is a good place to talk about returning an initial response.

spec/Section 6 -- Execution.md
robzhu: changed "Unsubscribe"

spec/Section 6 -- Execution.md
robzhu: Rewrote this section to make it more clear.

spec/Section 6 -- Execution.md
robzhu: reworded this to "Terminate {publishStream}" with some examples.

spec/Section 5 -- Validation.md
```diff
@@ -198,6 +198,67 @@ query getName {
...
+
+**Formal Specification**
+
+  * For each subscription operation definition {subscription} in the document
+  * Let {rootFields} be the top level selection set on {subscription}.
+    * {rootFields} must be a set of one.
+
+**Explanatory Text**
+
+Subscription operations must have exactly one root field.
```
DxCx: @robzhu what is the reason for that? why can't i subscribe to more then one subscription with one call?
so far i thought this is implementation limitation..

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,88 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+#### Subscribe
+
+The result of executing a subscription operation is a subscription object with
+the following capabilities:
+
+  * **Must** support observation of the associated publish stream (for example,
+  via iteration, callbacks, or reactive semantics).
+  * **Must** support cancellation of the subscription, see {Unsubscribe}.
+  * **Must** support a way to identify the subscriber/subscription pair (for
```
laneyk: I'm confused by what this means even with the additional explanation. What do you mean by "identify the subscriber/subscription pair"?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,88 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+  * *Application-specific logic to map from root field/variables to events*
+
+#### Publish
+
+Once a subscription is active, we listen for events on its event stream. Each
+event carries an optional payload, which we combine with the arguments from
+Subscribe() to resolve the selection set.
+
+  * For each {event} and {payload} on {eventStream}:
```
laneyk: Weren't we going to stop calling this "payload"?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,88 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  set defined on the subscription operation.
+
+Subscribe(schema, subscription, operationName, variableValues, initialValue):
+
+  * Let {subscriptionType} be the root Subscription type in {schema}.
+  * Assert: {subscriptionType} is an Object type.
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {rootField} be the first top level field in {selectionSet}.
+  * Let {eventStream} be the result of running {MapSubscriptionEvents(rootField, variableValues)}.
+  * Optionally: let {data} be the result of running {ExecuteRequest(schema, document, operationName, variableValues, initialValue)}, and return {data} on the subscription object.
```
laneyk: What is initialValue here? What's the value of specifying here and above (lines 166-7) that we can have an initial result? This leads to a lot of questions about how the execution works without any event that we don't address anywhere. Or is there discussion of this elsewhere?

spec/Section 6 -- Execution.md
robzhu: Updated to say "**Must** support a way to send data to the subscriber."

spec/Section 6 -- Execution.md
robzhu: Renamed this "eventData"

spec/Section 6 -- Execution.md
robzhu: Based on this discussion https://github.com/facebook/graphql/issues/283, if someone builds subscriptions that *can* be evaluated without event data, we would not necessarily want that to be in violation of the spec. I've added more text above the clarify the reasoning.

spec/Section 5 -- Validation.md
stubailo: I agree with Rob's text that starting with just one root field is reasonable. A later version of the spec could support multiple fields, but that requires figuring out a lot of edge cases around errors, when subfields are re-executed, and more.

spec/Section 6 -- Execution.md
```diff
@@ -103,8 +103,10 @@ Note: This algorithm is very similar to {CoerceArgumentValues()}.
 ## Executing Operations
 
 The type system, as described in the “Type System” section of the spec, must
-provide a query root object type. If mutations are supported, it must also
-provide a mutation root object type.
+provide a query root object type. If mutations/subscriptions are supported, it must also
+provide a mutation/subscription root object type, respectively.
```
stubailo: Probably should be plural:

> provide mutation/subscription root object types, respectively

However, I'd be in favor of splitting into two sentences, one about mutations and one about subscriptions.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,96 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
     selection set.
   * Return an unordered map containing {data} and {errors}.
 
+### Subscription
+
+Unlike queries and mutations, subscriptions have a lifetime that consists of three
```
stubailo: If this sentence is going to start with "unlike", it should probably provide the thing we are comparing to, like:

> Unlike queries and mutations, which deal with one request and response, subscriptions have a lifetime that consists of three

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,96 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+#### Subscribe
+
+The result of executing a subscription operation is a subscription object with
+the following capabilities:
+
+  * **Must** support observation of the associated publish stream (for example,
+  via iteration, callbacks, or reactive semantics).
+  * **Must** support cancellation of the subscription, see {Unsubscribe}.
+  * **Must** support a way to send data to the subscriber.
+  * **May** include an initial response associated with executing the selection
```
stubailo: Is the initial response supposed to be distinguishable from a regular response sent immediately after the subscription is started, from the point of view of the client?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,96 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+considered to be in the "active" state.
+
+#### Subscribe
+
+The result of executing a subscription operation is a subscription object with
+the following capabilities:
+
+  * **Must** support observation of the associated publish stream (for example,
+  via iteration, callbacks, or reactive semantics).
+  * **Must** support cancellation of the subscription, see {Unsubscribe}.
```
stubailo: I think using the word "cancel" here comes out of nowhere - maybe "must support unsubscribing" or "must have a way to stop the subscription"?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,96 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * **Must** support a way to send data to the subscriber.
```
stubailo: Is this referring to the event stream? I think this point could be more specific, since it's not clear what "data" is, or which entity is the "subscriber".

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,96 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+event carries an optional payload, which we combine with the arguments from
+Subscribe() to resolve the selection set.
+
+  * For each {event} and {eventData} on {eventStream}:
+    * Let {data} be the result of running
+      {ExecuteSelectionSet(selectionSet, subscriptionType, eventData, variableValues)}
+      *normally* (allowing parallelization).
+    * Let {errors} be any *field errors* produced while executing the
+      selection set.
+    * If {errors} is not empty, optionally, Unsubscribe() the subscription.
```
stubailo: Is this basically just saying the subscription can be terminated at any time by the backend? Or is there something special about terminating in response to a particular kind of error?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,96 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+We recommend thinking about the behavior of your system when this state is lost
+due to single-node failures. We can improve durability and availability by using
+dedicated sub-systems to manage this state. For example, event streams can be
+built using modern pub-sub systems, and client channels can be handled with a
+dedicated client gateway.
+
+Rather than mixing stateless (queries and mutations) and stateful
+(subscriptions), we recommend keeping the GraphQL server stateless and
+delegating all state persistence to these sub-systems.
```
stubailo: This almost sounds like it's recommending two schemas. I think what we are trying to say is that one thing should have all of the resolvers (for queries, mutations, _and_ subscriptions), and another thing should have all of the state stuff. Perhaps:

> Rather than mixing stateless and stateful logic in the same infrastructure, we recommend keeping your GraphQL execution system stageless, and delegating handling state about event streams and client connections to some other sub-system. In this design, the pub/sub system and client connection gateway call into the stateless execution layer when necessary to resolve GraphQL results. Production concerns like throttling and routing should be handled at the gateway layer to keep that complexity out of your API definition.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,96 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
     selection set.
   * Return an unordered map containing {data} and {errors}.
 
+### Subscription
```
stubailo: Should we include some example responses here, and perhaps an example of a lifecycle for a simple subscription?

spec/Section 6 -- Execution.md
asiandrummer: or "cancellation of the observation"

wincent:
> Do you think it would be helpful to include a diagram? I think the RFC content was very helpful to give people a sense of what the text was trying to say, and I don't know if people can really construct an image of what is going on from this spec. ***Perhaps that's the job of the documentation, though.***

@stubailo: I think you're right in what you say in that last sentence (emphasis added).

spec/Section 5 -- Validation.md
OlegIlyenko: I also would like to know the reason for this rule. I find that it is quite hard limitation, and in a way it goes against GraphQL principles where clients have the power to decide what it want to get.

>  that requires figuring out a lot of edge cases around errors

I think it would be helpful to list and discuss these edge cases. From my experience, there are definitely things to consider when merging different event streams from different GraphQL subscription fields, but I find it quite manageable.

OlegIlyenko: I listed most of the points I found as I was implement it in sangria here:

https://github.com/facebook/graphql/issues/282#issuecomment-284584160

I think that it boils down to points `3.i`, `5` and `6`. My take on it is here: https://github.com/facebook/graphql/issues/282#issuecomment-284586368

IMHO, if we need to make a trade off, I would rather disallow not-null root subscription field types than allow only a single subscription field in a query.

DxCx: @stubailo if one wants to experiment that's fine, but i don't think it should be in the official spec unless there is a reason (which is not implementation) behind it.

stubailo: IMO the reason is that allowing more than one root field requires that partial results are sent for the document, in the case when the fields are triggered via different events.

DxCx: all that is "required" is that the requested streams will have initial data before sending initial response. and in some cases the user might want them togather.
also, in the future one can use @defer on subscription so he wont have to wait for initial response for certain streams.

stubailo: Subscriptions are not even guaranteed to have any initial response, so you are definitely eventually going to end up in a situation where you receive a response that only includes one of the fields you asked for.

DxCx: @stubailo i was agreeing that the original limitation of 1 subscription field makes sense at this phase until things will become more stable, just saying that it shouldn't be part of the spec since there is no real reason for that other then implementation.
(I'm already experimenting with subscriptions that has more then 1 event)

if that's not the case please enlighten me, i'm waiting for an answer for a week now..

stubailo: The limitation isn't the implementation, it's about what the client is going to receive, as I mentioned above. The spec should specify everything about what the client should expect to receive, and I don't think there is a very reasonable answer for what the client would receive with multiple root fields.

OlegIlyenko: @stubailo can you demonstrate it on small example? I think it would be helpful.

DxCx:
there is still something which is not clear to me which is subfields updates.
for example, i have subscription of latestPost, and it has comments.
am i expected to update comments of latest post every time comment is updated/added?
or the client is expected to get this behavior only when applying `@live` directive? and the latestPost subscription should treat only posts event?

spec/Section 5 -- Validation.md
stubailo: well, what about the following subscription:

```
subscription {
  post1Comments: newComments(postId: 1) { ... }
  post2Comments: newComments(postId: 2) { ... }
}
```

When post 1 gets a comment, you could probably get a payload like:

```
{ post1Comments: { ... } }
```

And when post 2 gets one, you'd get:

```
{ post2Comments: { ... } }
```

I can't think of any situation where you would get both.

So you're sending two subscription fields in the request, but you're never going to get that entire response. Plus, you can now only unsubscribe to the whole thing at once (maybe a benefit) and it's not clear what to do when one of them has a fatal error - do both get unsubscribed?

From the client's point of view, IMO, having multiple root fields just adds complexity (on a conceptual level) without a lot of benefits.

stubailo:
> or the client is expected to get this behavior only when applying @live directive? and the latestPost subscription should treat only posts event?

That's correct. You only get updates when the `latestPost` field wants to send them, it doesn't depend on the subfields.

DxCx:
cool @stubailo thanks 👍

spec/Section 5 -- Validation.md
OlegIlyenko: @stubailo thanks a lot for describing in on an example! 

> So you're sending two subscription fields in the request, but you're never going to get that entire response.

I don't see it as a problem but rather as a natural behavior. This is inherent property of event-based interaction. This is also the reason why I suggested to disallow not-null root subscription field types. 

> you can now only unsubscribe to the whole thing at once (maybe a benefit) 

I also don't see it as a issue either. Can you describe in more detail why this behavior can be disadvantageous? (considering that you still can make 2 separate and isolated subscription queries if it suits better for the use-case at hand)

> it's not clear what to do when one of them has a fatal error - do both get unsubscribed?

I feel that either behavior is fine as long as it is defined in the spec. Though in this case I would suggest draw inspiration from streaming libraries: if 2 event streams are joined/merged together in a single result stream, then an error in either of these will also case the result stream to fail. If one of event steams naturally completes (because of the exhaustion), then the result steam still continue to emit events until all of the source streams are exhausted. I think this behavior is quite intuitive and widespread.

spec/Section 6 -- Execution.md
robzhu: Yes, I think the initial response should be distinguishable from a publish.

spec/Section 6 -- Execution.md
robzhu: Yes, from the client's perspective, the subscription could be terminated at any point by the server. Nothing specific here about what sorts of errors result in termination.

spec/Section 6 -- Execution.md
robzhu: Agreed. I've reworded this section to try and highlight the main message better.

spec/Section 6 -- Execution.md
stubailo: OK I completely misinterpreted this. So "subscriber" refers to the client consuming GraphQL, and "send data" means "send GraphQL results".

spec/Section 6 -- Execution.md
stubailo: That seems to make this bullet point unnecessary, since it refers to a thing that can happen at this time or any other time, or may not happen at all.

spec/Section 6 -- Execution.md
stubailo: > Rather than mixing stateless (queries and mutations) and stateful (subscriptions)

I think this still sounds like the above issue, because it is implying that _everything_ about subscriptions should be separate. When in fact, I think subscription resolvers should probably be defined next to the regular schema, and just the event management and gateway needs to be separate.

spec/Section 5 -- Validation.md
leebyron: Just to add a meta-point to this conversation:

In my view, there's nothing stopping us from working through these details and figuring out the edge and corner cases of allowing multiple subscription fields in a single request, however the choice we do have is to address those concerns now, or allow for more time to do so. In previous conversations @robzhu has had over the last few months about subscriptions, he has convinced many that this is far more complicated than we originally thought and may not have clear answers. This limitation is added mostly in a desire to expedite the addition of subscriptions to the spec, while reserving the ability to continue to work out how or if multi-field subscriptions should be allowed.

Had this limitations been omitted while also not making mention of how to address multi-field subscriptions in the spec, then we would see divergence of behavior and that could tie our hands in the future for deciding how to address these edge cases.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,127 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+If the operation is a subscription, the result of the operation is the creation
+of a persistent object on the server that exists for the lifetime of the
```
leebyron: "object" is a strange word to use in this context, IMHO. Perhaps the creation of a persistent *connection*?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,127 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+The result of executing a subscription operation is a subscription object with
+the following capabilities:
+
+  * **Must** support observation of the associated publish stream (for example,
+  via iteration, callbacks, or reactive semantics).
+  * **Must** support cancellation of the observation, see {Unsubscribe}.
+  * **Must** support a way to send data to the subscriber.
+  * **May** include an initial response associated with executing the selection
+  set defined on the subscription operation.
```
leebyron: I find this list of 4 things confusing and am having trouble understanding what exactly it is that it's describing. Specifically, these things seem to be describing an implementation detail that a client of GraphQL doesn't necessarily interact with or care about. For example, as a client of your GraphQL server, how would I test that you're adhering to these? For point 1, how would I know that the payloads I'm receiving are the associated to a "publish stream"? For point 2, why should I care? I can always just disconnect from the network. For point 3, this is obvious - nowhere else in the spec do we state a requirement that GraphQL must be used along with a way to send data to the requester. Point 4 is out of place as a "may" - it likely belongs elsewhere, probably as part of the "publish" section below.

What about writing lines 157-165 as:

The result of a subscription operation is a sequence of events over time, each event containing the result of executing the provided GraphQL selection on the data associated with that event.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,127 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+Most subscriptions can only be evaluated when event data is available. For
+example, a subscription tells us when users log on and off would require event
+data that tells us *who* logged on or off. Without this event data, the
+subscription cannot be evaluated. However, it is possible to define
+subscriptions that can be evaluated without event data (for example, the current
+time on the server, synthetically triggered every second). Subscriptions that do
```
leebyron: A lot of this paragraph is describing implementation details, and IMHO is muddying the term "event data" used in context elsewhere. The example about an event triggered every second *does* contain event data - that event data is the current time on the server.

spec/Section 6 -- Execution.md
leebyron: I find this very confusing and probably not what we want. If I understand this correctly, sometimes this algorithm results in a value `{data}` which is returned, sometimes it does not - in which case it returns nothing?

Also, this algorithm does not describe what should be done with `{eventStream}` once created - I would have expected `{eventStream}` to be returned.

As a concrete example - what should I expect to see differently between a subscription which has an initial response vs one that does not, say in terms of websocket payloads? Do I actually get a `{}` or `{data: null}` payload as an "ACK" for subscriptions without an initial value? Or is "ACK" a detail that belongs to the transport mechanism?

Perhaps another way to frame this is that "initial value" could be a first published event which simply occurs instantaneously after the creation of a subscription?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,127 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * Let {subscriptionType} be the root Subscription type in {schema}.
+  * Assert: {subscriptionType} is an Object type.
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {rootField} be the first top level field in {selectionSet}.
+  * Let {eventStream} be the result of running {MapSubscriptionEvents(rootField, variableValues)}.
+  * Optionally: let {data} be the result of running {ExecuteRequest(schema, document, operationName, variableValues, initialValue)}, and return {data} on the subscription object.
+
+MapSubscriptionEvents(rootField, variableValues):
+
+  * *Application-specific logic to map from root field/variables to events*
```
leebyron: This should probably be filled in. In particular, nothing is referencing the algorithm in Publish yet, which should probably be referenced here.

Or perhaps there's something missing? By name, I would expect this algorithm to map one set of events to another. Maybe this one should be renamed `CreateSubscriptionEvents` or something along those lines which should actually just be a delegation to provided logic, and then separately include `MapSubscriptionEvents` which calls into the `Publish` algo.

Also - this section on mapping event streams is a great opportunity to mention that while described as an inline algorithm, this step is often implemented *across* services, with the original event stream belonging to a subscription management service, and the execution happening on an API middleware service.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,127 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+#### Publish
+
+Once a subscription is created, we listen for events on its event stream. Each
+event carries an optional payload, which we combine with the arguments from
+Subscribe() to resolve the selection set.
+
+  * For each {event} and {eventData} on {eventStream}:
```
leebyron: This algo needs a name and arguments

leebyron: Also, the `For each {event} ...` line likely belongs in the `MapSubscriptionEvents` such that Publish can be written in a way that is independent to each publish action

leebyron: Is there value to describing `{event}` and `{eventData}` separately? For example, `{event}` is never referenced below - what would be its value?

spec/Section 6 -- Execution.md
leebyron: Agree w/ @stubailo - since this could also optionally unsubscribe the subscription if `{errors}` *is* empty, this line could be confusing.

I agree this bullet could just be removed here, but somewhere above mention that a subscription's stream of events may end at any moment, perhaps but not always in response to an error on the server

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,127 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+#### Unsubscribe
+
+The unsubscribe operation can be implemented in a number of ways. For example,
+by using a dedicated subscription manager, defining it as a method on the
+subscription object, or cancelling the iterator.
+
+Unsubscribe()
+
+  * Terminate {publishStream} (For example, cancel iteration, detach mapping function)
+  * Terminate and clean up {eventStream}
```
leebyron: This whole section seems like implementation detail and not really relevant to what is or isn't GraphQL, could it be collapsed into a footnote in the top section on subscriptions?

spec/Section 5 -- Validation.md
OlegIlyenko: Thanks a lot @robzhu and @leebyron for the insightful comments! I think now I got a better understanding of the issue. I was also in two minds on this. On one hand, I wanted to get a better understanding about motivation behind inclusion of this rule. But on the other hand, I don't want to delay the progress on subscriptions incision in the spec. I tend to agree that it is a good idea to disallow multiple fields for now and start the separate discussion. I think it is a discussion worth having. I am actually very glad that this point is considered in the spec since I was also quite concerned about the semantics of multiple subscriptions fields. 

> Can someone help me understand a compelling use case that is served by multi-root subscription operations that would not be equally served by separate individual subscriptions?

In general, I found it very valuable to have as much information from client as possible in single query. For example, the fact that a client can express its requirements for a view or particular part of the application in a single query allowed us to make very interesting optimizations which would be quite hard to do otherwise (it is quite hard to correlate seemingly independent requests/queries). So by allowing client to better express it's requirements with several subscription fields in a single query, we open a door for potential server-side optimizations.

Now that I'm equipped with new insights, I will give it another thought. This thread was definitely helpful in this respect.

Before we will introduce this rule though, I think it is important to consider the nullability of the subscription fields, as i mentioned above. It is possible to make a nullable field not-null later on in a backwards-compatible way. If we allow subscription fields to be not-null now, it might become a challenge in future to allow multiple subscription fields, if we decide to do so.

spec/Section 6 -- Execution.md
laneyk: I had (and still have) the same reaction as @leebyron to this part of the spec. I find it confusing and I don't really understand the value we're adding by explicitly mentioning it. As I mentioned before, this option raises a lot of questions for me that we don't really answer elsewhere. By calling out this "atomic hybrid query/subscribe" operation here, we're making it seem like a central concept for subscriptions (which I don't think it is). What if we leave this out of the spec until we can get more data on the demand for it and the best way to do it?

spec/Section 5 -- Validation.md
leebyron: Nullability may mean two different things in this context - this is a great point we should address.

One thing it may mean is that a subscription may not exist given some inputs in a way that isn't considered an error. I think this interpretation is both not what you were referring to, and also probably confusing to think about. The schema talks about the type of the *payload result* - so we're talking about the types of responses. We should probably make this point in the spec to clarify.

Secondly the nullability of the responses. This is one of the concerns with multi-field subscriptions to address later. For example, should it be legal to have a subscription field `streamThings: String?` where it is legal for any payload in the event sequence to in fact be `null`? I don't see a compelling reason to explicitly disallow this - though it is an edge case.

I think handling the payloads of multi-field subscriptions will need to account for this

spec/Section 6 -- Execution.md
leebyron: @robzhu and I talked about this offline, but I just wanted to point out that this initialValue addition improves but doesn't guarantee a solution to race conditions. Events can still potentially be repeated or dropped between the generation of the initial value payload and subsequent payloads, depending on the system design and the latency of events flowing through it.

I think the kicker here is that most examples of wanting initial data before subscription payloads desire a *different* query shape - in the example above a list of comments vs an individual comment. The spec as written wouldn't allow for this distinction, so I actually think your example is more powerful while requiring less from the spec.

robzhu: Agreed. Upon further discussion with @laneyk and @leebyron, initial value behavior can be supported out-of-the-box using two sub-fields in a subscription, as in @rmosolgo's example. However, the model of an "atomic query/subscribe" is probably better supported via batching. 

To be clear, the resulting edit is that we will simply not mention server behavior with respect to initial value. If you choose to implement/use it, that does not make your server violate the spec.

spec/Section 6 -- Execution.md
```diff
@@ -103,8 +103,9 @@ Note: This algorithm is very similar to {CoerceArgumentValues()}.
 ## Executing Operations
 
 The type system, as described in the “Type System” section of the spec, must
-provide a query root object type. If mutations are supported, it must also
-provide a mutation root object type.
+provide a query root object type. If mutations or subscriptions are supported, it must also provide a mutation and subscription root object type, respectively.
```
leebyron: nit: line break to maintain roughly 80c?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
   * Return an unordered map containing {data} and {errors}.
 
+### Subscription
+
+We define an event stream as a sequence of discrete events over time that can be
+observed. An observer of an event stream may cancel observation to avoid stop
+future events.
+
+If the operation is a subscription, the result is an event stream called the
+"Publish Stream" where each event in the stream is called a "Publish Payload".
```
leebyron: Flip the order of these two paragraphs.

Also for the para on event streams, some additional detail is necessary around event stream completion. Maybe also some examples for color. Perhaps:

> An event stream represents a sequence of discrete events over time which can be observed. As an example, a "Pub-Sub" system may produce an event stream when "subscribing to a topic", with an event occurring on that event stream for each "publish" to that topic. Event streams may produce an infinite sequence of events or may complete at any point. Event streams may complete in response to an error or simply because no more events will occur. An observer may at any point decide to stop observing an event stream, after which it must receive no more events from that event stream. 
> 
> Note: If an event stream's observer has stopped observing, that may be a good opportunity to clean up any associated resources such as closing any connections which are no longer necessary.

leebyron: For the other paragraph, I'm not sure the events in the stream need a name so much as they need a brief description:

> If the operation is a subscription, the result is an event stream called the "Response Stream" where each event in the event stream is the result of executing the operation for each new event on an underlying "Source Stream".

leebyron: I fear "Publish Stream" may be too ambiguous - it could be too easy to confuse it with the underlying stream. How about: "Source Stream" and "Response Stream"?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * Let {subscriptionType} be the root Subscription type in {schema}.
+  * Assert: {subscriptionType} is an Object type.
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {rootField} be the first top level field in {selectionSet}.
+  * Let {eventStream} be the result of running {CreateUnderlyingEventStream(rootField, variableValues)}.
+  * Let {publishStream} be the result of running {MapEventToPayload(eventStream)}
+
+CreateUnderlyingEventStream(rootField, variableValues):
+
+  * *Application-specific logic to map from root field/variables to events*
```
leebyron: As a reflection of the feedback on https://github.com/graphql/graphql-js/pull/846/ - let's break out the top portion of `Subscribe` and name this second algorithm "Resolve" since it's resolving from a field.

`CreateSourceEventStream(schema, subscription, operationName, variableValues, initialValue)`

`ResolveFieldEventStream(subscriptionType, rootValue, fieldName, argumentValues)` <- should mirror ResolveFieldValue - look there for an example of getting these values

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+#### Publish Stream
+
+Each event in the underlying event stream triggers execution of the subscription
+selection set.
+
+MapEventToPayload(eventStream):
```
leebyron: Missing the arguments referred to below

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+  * For each {event} on {eventStream}:
+    * Let {publishPayload} be the result of running
+      {ExecuteSelectionSet(selectionSet, subscriptionType, event, variableValues)}
+      *normally* (allowing parallelization).
+    * Let {errors} be any *field errors* produced while executing the
+      selection set.
+    * Yield an unordered map containing {publishPayload} and, optionally,
+      {errors} on {publishStream}.
+  * At any time while the publish stream is active, the client or server may
```
leebyron: This is not really a step in the algorithm, it can be removed. However, we should have a step here which clarifies what to do when the source event stream runs out:

`* Once {sourceEventStream} has completed, this Event Stream also completes.

spec/Section 6 -- Execution.md
leebyron: Perhaps `MapSourceToResponseEventStream`?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+variables as inputs.
+
+Subscribe(schema, subscription, operationName, variableValues, initialValue):
+
+  * Let {subscriptionType} be the root Subscription type in {schema}.
+  * Assert: {subscriptionType} is an Object type.
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {rootField} be the first top level field in {selectionSet}.
+  * Let {eventStream} be the result of running {CreateUnderlyingEventStream(rootField, variableValues)}.
+  * Let {publishStream} be the result of running {MapEventToPayload(eventStream)}
```
leebyron: Add a step which returns `{publishStream}`

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * *Application-specific logic to map from root field/variables to events*
+
+#### Publish Stream
+
+Each event in the underlying event stream triggers execution of the subscription
+selection set.
+
+MapEventToPayload(eventStream):
+
+  * For each {event} on {eventStream}:
```
leebyron: We need to return something, so perhaps we can add a step before this:

`* Return a new event stream which yields events as follows:`

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+    * Let {publishPayload} be the result of running
+      {ExecuteSelectionSet(selectionSet, subscriptionType, event, variableValues)}
+      *normally* (allowing parallelization).
+    * Let {errors} be any *field errors* produced while executing the
+      selection set.
+    * Yield an unordered map containing {publishPayload} and, optionally,
```
leebyron: Break this into two such that:

* Let {response} be an unordered map containing {data} and {errors}.
* Yield an event containing {response}.

leebyron: Perhaps we should define `ExecuteSubscription` to mirror `ExecuteQuery`? It should be identical except for the first step. Then we could just have `Let {response} be the result of `{ExecuteSubscription(...)}`.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+      {errors} on {publishStream}.
+  * At any time while the publish stream is active, the client or server may
+  Unsubscribe().
+
+Common reasons for Unsubscribe() include:
+  * client no longer wants to receive payloads for a subscription.
+  * The underlying event stream has produced an error or has naturally ended.
```
leebyron: Move this paragraph below to be within the `Unsubscribe` section. Also I think this would read easier as prose instead of bullet points.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+      "sender": "Hagrid",
+      "text": "You're a wizard!"
+    }
+  }
+}
+```
+
+#### Recommendations and Considerations for Supporting Subscriptions
+
+Supporting subscriptions is a large change for any GraphQL server. Query and
```
leebyron: "significant" instead of "large"?

spec/Section 6 -- Execution.md
leebyron: Once the `Subscribe()` step is only these two steps, I think you could make a reference to your bottom most section by adding a `Note: ` to clarify that in many real services, this `Subscribe` algorithm is run on a separate service than the execute step, and to refer to the section below on Considerations when supporting subscriptions

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  "data": {
+    "newMessage": {
+      "sender": "Hagrid",
+      "text": "You're a wizard!"
+    }
+  }
+}
+```
+
+#### Recommendations and Considerations for Supporting Subscriptions
```
leebyron: This is a long section title and it's specifically about scale. So perhaps just `Supporting Subscriptions at Scale`?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+Supporting subscriptions is a large change for any GraphQL server. Query and
+mutation operations are stateless, allowing scaling via cloning GraphQL server
+instances. Subscriptions, by contrast, are stateful. The pieces of state for a
+subscription are:
+
+  * Subscriber/client channel
+  * Subscription document
+  * Variables
+  * Execution context (for example, current logged-in user, locale, etc.)
+  * Event stream resulting from Subscribe step (above)
```
leebyron: I find this list of things a bit confusing and overly detailed for the point you're trying to make. Perhaps just:

Subscriptions, by contrast, are long-lived and stateful and require maintaining the GraphQL document, variables, and other context over the lifetime of the request.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+We recommend thinking about the behavior of your system when this state is lost
```
leebyron: Spec shouldn't speak in the third person. Instead:

Consider the behavior of your system when state is lost due to the failure of a single machine in a service. Durability and availability may be improved by having a separate dedicated service for managing this state.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +146,120 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+due to single-node failures. We can improve durability and availability by using
+dedicated sub-systems to manage this state. For example, event streams can be
+built using modern pub-sub systems, and payload delivery to clients can be
+handled by a dedicated client gateway tier.
+
+For systems with high capacity, availability, and durability requirements, we
+recommend keeping the GraphQL server stateless and delegating all state
+persistence to sub-systems that are designed for stateful scaling. Note that
+subscription types are still defined in the original schema along with queries
+and mutations.
```
leebyron: This note is a bit out of place. You can define this stuff wherever the heck you want, that's implementation detail. Also, this paragraph is really just repeating the point made in the previous one. I think you can probably just remove it entirely if not move a few of the more fine-pointed words into the prior paragraph

spec/Section 6 -- Execution.md
robzhu: It's pretty close after the accumulated edits.

spec/Section 6 -- Execution.md
robzhu: I think this is clearly from the "for each" iteration of {sourceStream}.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+more events from that event stream.
+
+Note: If an event stream's observer has stopped observing, that may be a good
+opportunity to clean up any associated resources such as closing any connections
+which are no longer necessary.
+
+#### Subscribe
+
+Executing a subscription creates a persistent function on the server that
+maps an underlying event stream to the Publish Stream. The logic to create the
```
leebyron: maps an underlying Source Stream to a returned Response Stream

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+underlying event stream is application-specific and takes the root field and
```
leebyron: to create the Source Stream is application-specific

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+query variables as inputs.
+
+Subscribe(schema, subscription, operationName, variableValues, initialValue):
```
leebyron: `operationName` isn't used here. Other related algos like `ExecuteQuery` look like:

ExecuteQuery(query, schema, variableValues, initialValue):

So perhaps:

ExecuteSubscription(subscription, schema, variableValues, initialValue):

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+  * Let {subscriptionType} be the root Subscription type in {schema}.
+  * Assert: {subscriptionType} is an Object type.
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {sourceStream} be the result of running {CreateSourceEventStream(schema, subscription, selectionSet, variableValues, initialValue)}.
```
leebyron: The 3 steps above this one should probably be a step in `CreateSourceEventStream`. That algo's arguments should mirror that of this one.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * Let {responseStream} be the result of running
+    {MapSourceToResponseEvent(sourceStream)}
```
leebyron: one line for this should be fine

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+    * Let {publishPayload} be the result of running
+      {ExecuteSelectionSet(selectionSet, subscriptionType, event, variableValues)}
+      *normally* (allowing parallelization).
+    * Let {errors} be any *field errors* produced while executing the
+      selection set.
+    * Let {response} be an unordered map containing {publishPayload} and, optionally,
+      {errors}.
+    * Yield an event containing {response}.
+
+Note: in large scale subscription systems, ExecuteSelectionSet and Subscribe
```
leebyron: nit: `{}` wrapped around algo names so they link in the published html

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+MapSourceToResponseEvent(sourceStream, subscriptionType, selectionSet, variableValues):
+
+  * Return a new event stream {publishStream} which yields events as follows:
+  * For each {event} on {sourceStream}:
+    * Let {publishPayload} be the result of running
+      {ExecuteSelectionSet(selectionSet, subscriptionType, event, variableValues)}
+      *normally* (allowing parallelization).
+    * Let {errors} be any *field errors* produced while executing the
+      selection set.
+    * Let {response} be an unordered map containing {publishPayload} and, optionally,
```
leebyron: this is a bit ambiguous since we expect the unordered map to use the key `data` but this implies using a key `publishPayload` - maybe just rename the variable

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * Let {resolver} be the internal function provided by {subscriptionType} for
+    determining the resolved value of a field named {fieldName}.
+  * Return the result of calling {resolver}, providing {rootValue} and {argumentValues}.
+
+#### Publish Stream
+
+Each event in the underlying event stream triggers execution of the subscription
+selection set.
+
+MapSourceToResponseEvent(sourceStream, subscriptionType, selectionSet, variableValues):
```
leebyron: This should probably operate on the whole `subscription` instead of `selectionSet` and `schema` instead of `subscriptionType` - and pass those through to another algorithm which is a mirror of `ExecuteQuery` - that becomes the seam for passing the command to another system. Maybe call it `ExecuteSubscriptionEvent`?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,131 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+      {errors}.
+    * Yield an event containing {response}.
+
+Note: in large scale subscription systems, ExecuteSelectionSet and Subscribe
+algorithm may be run on separate services to maintain predictable scaling
+properties. See the section below on Supporting Subscriptions at Scale.
+
+#### Unsubscribe
+
+Unsubscribe cancels the Publish Stream. This is also a good opportunity for the
```
leebyron: cancels the Response Stream?

robzhu:
I took another pass and made changes per your suggestions. Reading everything top-to-bottom, it feels a lot more dense/verbose. We also have some duplication, for example with ExecuteSubscriptionEvent. Do you feel this restructuring buys us consistency/clarity?

spec/Section 6 -- Execution.md
leebyron: Also noticed that a step should be added to `{ExecuteRequest}` above to handle subscription requests.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+which are no longer necessary.
+
+#### Subscribe
+
+Executing a subscription creates a persistent function on the server that
+maps an underlying Source stream to the Publish Stream. The logic to create the
+Source stream is application-specific and takes the root field and query
+variables as inputs.
+
+Subscribe(subscription, schema, variableValues, initialValue):
```
leebyron: Should this be called `ExecuteSubscription` to mirror the other algos called by `ExecuteRequest`?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {rootField} be the first top level field in {selectionSet}.
+  * Let {argumentValues} be the result of {CoerceArgumentValues(subscriptionType, rootField, variableValues)}.
+  * Let {fieldStream} be the result of running {ResolveFieldEventStream(subscriptionType, initialValue, rootField, argumentValues)}.
+  * Return {fieldStream}.
+
+ResolveFieldEventStream(subscriptionType, rootValue, fieldName, argumentValues):
+  * Let {resolver} be the internal function provided by {subscriptionType} for
+    determining the resolved value of a field named {fieldName}.
+  * Return the result of calling {resolver}, providing {rootValue} and {argumentValues}.
```
leebyron: It would be nice to add a `Note: ` below this algorithm mentioning its intentional similarity to `{ResolveFieldValue}`

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+#### Publish Stream
```
leebyron: "Response Stream"?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+  * Let {subscriptionType} be the root Subscription type in {schema}.
+  * Assert: {subscriptionType} is an Object type.
+  * Let {selectionSet} be the top level Selection Set in {subscription}.
+  * Let {data} be the result of running
+    {ExecuteSelectionSet(selectionSet, subscriptionType, initialValue, variableValues)}
+    *normally* (allowing parallelization).
+  * Let {errors} be any *field errors* produced while executing the
+    selection set.
+  * Return an unordered map containing {data} and {errors}.
```
leebyron: Similarly, it would be nice to have a `Note: ` which mentions the intentional similarity to `{ExecuteQuery}`

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+Unsubscribe cancels the Response Stream. This is also a good opportunity for the
+server to clean up the underlying event stream and any other resources used by
+the subscription. Here are some example cases in which to Unsubscribe: client
+no longer wishes to receive payloads for a subscription; the source event stream
+produced an error or naturally ended; the server encountered an error during
+ExecuteSubscriptionEvent.
+
+Unsubscribe()
+
+  * Cancel {responseStream}
```
leebyron: Above we used the term "stop observing" - we should probably be consistent here.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+scaling properties. See the section below on Supporting Subscriptions at Scale.
+
+#### Unsubscribe
+
+Unsubscribe cancels the Response Stream. This is also a good opportunity for the
+server to clean up the underlying event stream and any other resources used by
+the subscription. Here are some example cases in which to Unsubscribe: client
+no longer wishes to receive payloads for a subscription; the source event stream
+produced an error or naturally ended; the server encountered an error during
+ExecuteSubscriptionEvent.
```
leebyron: I agree with you that this is redundant with the note above. I think you should merge the two together, either relying on the Note above and leaving this section with a much simpler explanation of the algorithm below, or removing the Note above and leaving those details here.

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+
+Unsubscribe()
+
+  * Cancel {responseStream}
+
+#### Example
```
leebyron: Elsewhere we use `**Example**` to avoid creating sections in the TOC for examples. Also, what do you think about moving the Example directly above the algorithms as the last part of the top level "Subscription" section?

spec/Section 6 -- Execution.md
```diff
@@ -143,6 +147,136 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
+  "data": {
+    "newMessage": {
+      "sender": "Hagrid",
+      "text": "You're a wizard!"
+    }
+  }
+}
+```
+
+#### Supporting Subscriptions at Scale
```
leebyron: You might consider moving this section above the algorithms and making this a `**Title**` instead so it's more directly linked to the discussion of subscriptions.

leebyron:
> We also have some duplication, for example with ExecuteSubscriptionEvent. Do you feel this restructuring buys us consistency/clarity?

Yes, I think that it helps to illustrate the seam between creating a subscription and executing each event, which will help those building this across two services.

spec/Section 6 -- Execution.md
robzhu: Since the server is the producer of the response stream, if we use "stop observing", we probably give the impression that we are describing client behavior, whereas the previous sections clearly describe server behavior.

spec/Section 6 -- Execution.md
robzhu: Removed the note in the "### Subscription" section and left this paragraph in place. I think this reads more clearly after moving the example.

spec/Section 6 -- Execution.md
robzhu: Great suggestion!

spec/Section 6 -- Execution.md
```diff
@@ -34,12 +34,18 @@ ExecuteRequest(schema, document, operationName, variableValues, initialValue):
...
 GetOperation(document, operationName):
 
   * If {operationName} is {null}:
     * If {document} contains exactly one operation.
-      * Return the Operation contained in the {document}.
+      * Let {operation} be the Operation contained in the {document}.
+      * If {operation} is a subscription operation:
+        * If {operation} contains more than one root field, produce a query error.
+        * Return {operation}.
+      * Otherwise return {operation}.
```
leebyron: These new steps aren't necessary. The validation rule will check for this, and the `CreateSourceEventStream` already does the right thing of using the first (only) field in the set

spec/Section 2 -- Language.md
```diff
@@ -193,12 +193,14 @@ OperationDefinition :
...
-OperationType : one of `query` `mutation`
+OperationType : one of `query` `mutation` `subscription`
 
-There are two types of operations that GraphQL models:
+There are three types of operations that GraphQL models:
 
   * query - a read-only fetch.
   * mutation - a write followed by a fetch.
+  * subscription - a long-lived request that returns data whenever a domain
+    event triggers.
```
leebyron: to recycle terms, how about: "a long-lived request that fetches data in response to source events."

leebyron:
This looks awesome. Imminently mergeable.

leebyron:
This turned out great.

josephsavona:
Nice job with this, @robzhu!!!

Issue #288, opened 2017-03-16
Clarification of non-null arguments wording 


gtod:
The spec currently has (my italics)

#### 5.3.3.2 Required Non-Null Arguments

**Explanatory Text**

> Arguments can be required. Arguments are required if the type of the argument is non‐null. If it is not non‐null, the argument is optional. When an argument type is non‐null, *and is required*, the explicit value null may also not be provided.

I think that italicised phrase should be *and thus required*.  In fact, it should probably just be omitted.  Or even:

> Arguments can be required.  If the argument type is non null the argument is required and furthermore the explicit value null may not be provided. If the argument type is not non‐null, the argument is optional.

Either that or I have misunderstood...

robzhu:
Thanks for spotting this. Agreed that this wording could be more clear. Please go ahead and submit a PR.

wincent:
Closed via https://github.com/facebook/graphql/pull/306

Thanks for the report, and for the fix @gtod.

PR #306, opened 2017-05-04
Clairfy required non null arguments wording (#288)

gtod:
Per #288


wincent:
Superb! Thanks a lot @gtod.

PR #307, opened 2017-05-17
Update README.md

iamaldi:
Minor typo fix.


wincent:
Thanks @iamaldi. FWIW I think the original wording was intentional and not a typo (it's grammatically valid) but your wording is nicer so let's go with it!

iamaldi:
@wincent That's my first pull request so I'm happy with that :+1:  !!!

PR #317, opened 2017-06-05
Clarify handling of null handling within lists

TOTBWF:
As it stands, the spec is unclear on what to do when an Non-Null element of a list resolves to null. These changes clarify that the entire list should resolve to null, and to propagate the error upwards


leebyron:
Thank you for this clarification. I agree this is a helpful addition!

PR #319, opened 2017-06-13
Add '@ignoreIfUnknown' to allow forward compatibility for clients

IvanGoncharov:
GraphQL is an excellent technology for backward compatibility of APIs. However, sometimes you need to provide forward compatibility for the clients. 

The big example is introspection mechanism of GraphQL itself. Up to this moment, there were two changes in it:
  - `onOperation`, `onFragment` and `onField` fields were deprecated in favor of `locations`
  - new `subscriptionType` field was added

In both cases, backward compatibility was preserved, so old clients can still work with newer GraphQL APIs. However, forward compatibility is broken since you can't use the same introspection query for both old and new servers. For example, GraphiQL solves this by having fallback query without `subscriptionType` field: 
https://github.com/graphql/graphiql/blob/master/src/utility/introspectionQueries.js#L11-L13

This problem is even trickier since GraphQL libraries for many languages (including `graphql-js`) expose standard introspection query. Should this "standard" query incorporate features from the last standard or should it stick to the query supported by the first public release of GraphQL?

To address forward compatibility I propose to add `@ignoreIfUnknown` directive that will silently ignore fields that weren't defined inside appropriate type. Comparing to `skip`/include` it will not only omit value from response but also prevent the query failing during validation.

This issue is not only limited to introspection. I think this feature can solve similar problems for IoT (different firmware versions), distributed systems, etc.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -969,7 +969,8 @@ a non-null input type as invalid.
 A GraphQL schema includes a list of the directives the execution
 engine supports.
 
-GraphQL implementations should provide the `@skip` and `@include` directives.
+GraphQL implementations should provide the `@skip`, `@include` and
+`ignoreIfUnknown` directives.
 
 
 ### @skip
@@ -1010,6 +1011,20 @@ field or fragment, it *must* be queried only if the `@skip` condition is false
 must *not* be queried if either the `@skip` condition is true *or* the
 `@include` condition is false.
 
+### @ignoreIfUnknown
+
+The `@ignoreIfUnknown` directive may be provided for fields to safely ignore
+them during validation and execution of the query in case they are not defined.
+
+In this example `experimentalField` will be queried only if Query type defines
+this field. However, this query will always pass validation even if 
+`experimentalField` is not defined inside Query type.
+
+```graphql
+{
+  experimentalField @ignoreIfUnknown
+}
+```
 
 ## Initial types
 


leebyron:
I'm not sure we actually need this. Validation and Execution are separated and Execution actually follows this behavior exactly. If a field is not defined, it's skipped over during execution.

Validation, on the other hand, is what looks for field definitions. I'm not sure we want to provide mechanisms for opting out of validation, and it's not the only way to supply forward-compatibility.

While many servers receive queries and validate and execute in one request, I think it's becoming best practice to persist queries to be used at production time. At Facebook we do this by persisting queries against the latest branch of our web services which have defined the newest fields at which point we validate them. Later, these queries can be executed on any production machine, even if the newest version has not been deployed yet - and they will execute with the exact behavior described in this proposal.

stubailo:
> While many servers receive queries and validate and execute in one request, I think it's becoming best practice to persist queries to be used at production time.

I don't think that the community has actually adopted this pattern yet. I think that in a practical sense 99.9% of GraphQL users are running validation on every query on every request.

IvanGoncharov:
> If a field is not defined, it's skipped over during execution.

IMHO, this behavior breaks API contract. Here is a short IDL to illustrate this:
```graphql
type Person {
  username: String!
  isOnline: Boolean
}
```
And matching JS code: 
```js
if (!person) throw Error('Can\'t get profile info');
const {username, isOnline} = person;
const statusImage = unknownImage;
if (isOnline !== null)
  statusImage = isOnline? onlineImage : offlineImage;

return <div>{username}<a src={statusImage}/></div>
```
In that code I never expect `username` or `isOnline` to be `undefined` so result HTML will be incorrect. Moreover if you generate Flow & TS typings from Query document they will also reasure you in this assumption.

> While many servers receive queries and validate and execute in one request, I think it's becoming best practice to persist queries to be used at production time.

What about tooling that is intended to be used in development like GraphiQL or our own GraphQL Voyager and GraphQL Faker. They need the ability to send introspection query to servers implemented against all releases of GraphQL specification and not only the newest one. Should we have separate introspection query for every spec release that adds new fields into introspection? (as it's already [done in GraphiQL](https://github.com/graphql/graphiql/blob/master/src/utility/introspectionQueries.js#L11-L13)).

> Later, these queries can be executed on any production machine, even if the newest version has not been deployed yet - and they will execute with the exact behavior described in this proposal.

I agree that mechanism is very similar. However, proposed directive do this explicitly and prevents surprise for the clients and tooling. For example generator of Flow & TS will be able to mark this field as optional because it is marked with `@ignoreIfUnknown` directive.

leebyron:
While this proposal may help GraphiQL's introspection query, it is not the only way to do so. Another approach that would not require any spec changes is to perform a two-pass introspection which first learns of the capabilities of the introspection system, and second learns of the capabilities of the schema. I'm not concerned about the introspection system relative to evaluating this particular change, so I think it's safe to leave it out as an example.

Also, the more generalized version of this problem isn't actually solved by this proposal because even if this was added to the spec, not all graphql libraries will use the latest version of the spec and not all servers will deploy the latest versions of those libraries even if they do. This means that while you could add `@ignoreIfUnknown` to your query for some tools like GraphQL Faker, there's no guarantee that a server will have support for this directive, and thus you cannot rely on using them from tools like this.

I'm concerned that this proposal is based on an arrived solution and not grounded in a real problem people using GraphQL regularly encounter. I would love to move discussion for this proposal towards some of these problems so that we can investigate a breadth of solutions for those problems which this particular solution can be evaluated against.

IvanGoncharov:
> I'm concerned that this proposal is based on an arrived solution and not grounded in a real problem people using GraphQL regularly encounter.

Agree 👍 I have only introspection use case in mind so I can't say for sure if it solves real life problems in any other domains. 

> for example @skip and @include only "should" exist.

So `@skip` and `@include` always pass validation but there is no guarantee that the server actually implements them. Or am I missing something? (this is not so clear from the spec)

> Another approach that would not require any spec changes is to perform a two-pass introspection which first learns of the capabilities of the introspection system, and second learns of the capabilities of the schema.

This is definitely possible however it complicates client a lot since the second query should be at least partly dynamic. 

I think about something like `__graphqlVersion` which will return the latest spec release that server supports, e.g. `October 2016`.  In such case, if the latest introspection query fails, the client can send:
``` graphql
{
  __graphqlVersion
}
```
and use appropriate query.

@leebyron What do think about adding `__graphqlVersion`?
I can prepare PRs both on the spec and `graphql-js`.

leebyron:
This discussion was quite stale, and it seems we decided against this particular approach, though I agree we still need a solution for feature discovery via introspection. My suggestion to explore would be to first introspect on the introspection types in a first query, then use the results to write a second query which can introspect on the schema's types.

PR #320, opened 2017-06-13
Add support for leading vertical bar in union types

alexandrebodin:
This spec change is matched with a PR on the language repository 

PR on the parser https://github.com/graphql/graphql-js/pull/907

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -757,19 +757,21 @@ Union types have the potential to be invalid if incorrectly defined.
 
 #### Union type syntax
 
-Union types are a list of types joined with a vertical bar `|`
+Union types are defined by delimiting one or more types with a single vertical bar character `|`:
 
 ```graphql
 union SearchResult = Photo | Person
 ```
 
-You may also use a leading vertical bar
+You may also use a leading vertical bar.
 
 ```graphql
 union SearchResult =
   | Photo
   | Person
 ```
+
+Note Trailing vertical bars are not supported.
 
 ### Enums
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -755,6 +755,22 @@ Union types have the potential to be invalid if incorrectly defined.
    Similarly, wrapping types may not be member types of a Union.
 2. A Union type must define one or more unique member types.
 
+#### Union type syntax
+
+Union types are a list of types joined with a vertical bar `|`
+
+```graphql
+union SearchResult = Photo | Person
+```
+
+You may also use a leading vertical bar
+
+```graphql
+union SearchResult =
+  | Photo
+  | Person
+```
+
 ### Enums
 
 GraphQL Enums are a variant on the Scalar type, which represents one of a
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -763,7 +763,7 @@ Union types are defined by delimiting one or more types with a single vertical b
 union SearchResult = Photo | Person
 ```
 
-You may also use a leading vertical bar.
+You may also use a leading vertical bar:
 
 ```graphql
 union SearchResult =
@@ -771,7 +771,11 @@ union SearchResult =
   | Person
 ```
 
-Note Trailing vertical bars are not supported.
+Trailing delimiters are not supported:
+
+```!graphql
+union SearchResult = Photo | Person |
+```
 
 ### Enums
 


spec/Section 3 -- Type System.md
```diff
@@ -755,6 +755,22 @@ Union types have the potential to be invalid if incorrectly defined.
    Similarly, wrapping types may not be member types of a Union.
 2. A Union type must define one or more unique member types.
 
+#### Union type syntax
+
+Union types are a list of types joined with a vertical bar `|`
```
robzhu: How about: "Union types are defined by delimiting one or more types with a single vertical bar character `|`:"

spec/Section 3 -- Type System.md
```diff
@@ -755,6 +755,22 @@ Union types have the potential to be invalid if incorrectly defined.
...
+
+```graphql
+union SearchResult = Photo | Person
+```
+
+You may also use a leading vertical bar
```
robzhu: We should probably mention the treatment of a trailing vertical bar.

alexandrebodin:
@robzhu changes made

spec/Section 3 -- Type System.md
```diff
@@ -757,20 +757,22 @@ Union types have the potential to be invalid if incorrectly defined.
...
-You may also use a leading vertical bar
+You may also use a leading vertical bar.
 
 ```graphql
 union SearchResult =
   | Photo
   | Person
 ```
 
+Note Trailing vertical bars are not supported.
```
robzhu: I would update this to say "Trailing delimiters are not supported:" followed by an (invalid) example. You can "```!graphql" to designate an invalid example.

spec/Section 3 -- Type System.md
```diff
@@ -757,20 +757,22 @@ Union types have the potential to be invalid if incorrectly defined.
...
 
-Union types are a list of types joined with a vertical bar `|`
+Union types are defined by delimiting one or more types with a single vertical bar character `|`:
 
 ```graphql
 union SearchResult = Photo | Person
 ```
 
-You may also use a leading vertical bar
+You may also use a leading vertical bar.
```
robzhu: Change this to `You may also use a leading vertical bar:` to be consistent with the opening clause.

IvanGoncharov:
@alexandrebodin @robzhu But IDL is not merged yet 😱 , see #90.
This PR should have been merged into `rfc-idl` branch not into the `master`.

Moreover, this is a syntax change in IDL and it has nothing to do with `Section 3 -- Type System.md`. It should go to `Section 2 -- Language.md` [here](https://github.com/facebook/graphql/pull/90/files#diff-fe406b08746616e2f5f00909488cce66R1098) and [here](https://github.com/facebook/graphql/pull/90/files#diff-c9d195396161bb690c5fd8899af11d78R208).

robzhu:
@IvanGoncharov good point. Just checked with @leebyron as well. My mistake, I wasn't fully caught up on the status of the rfc-idl branch. @alexandrebodin, can you please re-submit the PR against the rfc-idl branch?

IvanGoncharov:
@leebyron @robzhu I think this PR gives a good example of what happens if you don't give enough time for the community to provide feedback before merging. With implementations in a few dozen of languages and a huge number of tools/libraries around GraphQL there are many developers who can provide valuable feedback. Moreover, they are subscribed for updates in this repo:
image

Can we make a rule that PRs with changes beyond grammar/formatting fixes will not be merged in less than 48 hours? That way everyone will have a chance to review it and provide feedback.

robzhu:
@IvanGoncharov that sounds very reasonable. We should probably add a `CONTRIBUTING.md` doc to the spec repo. Do you have a similar concerns about the change being merged into the reference implementation? https://github.com/graphql/graphql-js/pull/907

stubailo:
It's also not clear when these changes take effect? Would this be only in the next "version" of GraphQL, or is merging the PR an immediately binding decision that needs to be supported by all implementations ASAP?

robzhu:
@stubailo In retrospect, it's clear this PR should have been made against the rfc-idl repo. It makes sense to me that we would align these spec updates with batches to coincide with the working group schedule (TBD). Do you have any ideas you think would work well here?

stubailo:
I think having a clear distinction between "changes" and "clarifications", and making sure that changes are merged into a non-deployed "next version" branch, would be enough!

Then, having a clear schedule around releases or a warning process when a new release is about to come out, will also help a lot. So that all server/client authors can have some time to update things if they want to.

IvanGoncharov:
> We should probably add a CONTRIBUTING.md doc to the spec repo.

Great idea 👍 It would make sense to add other unwritten rules, e.g. reference implementation should be updated to match PR before it will be merged. Probably it would make sense to describe it in form of typical workflow of PR in this repo.

> Do you have a similar concerns about the change being merged into the reference implementation?

Less so, because it more like a playground and accepting something there doesn't mean it will be merged into spec itself. 

My idea of ideal workflow is like that:
1)  Submit PR to the spec
2) Wait 1-2 days to get feedback
3) Start to work on PR for `graphql-js`
4) Submit PR on `graphql-js` and keep it in sync with PR on spec
5) Wait for PR in `graphql-js` to be  merged and new version of `graphql-js` released.
6) Wait at least a couple days to get feedback from users in field.
7) Merge PR into the spec.

But maybe it's too much of bureaucracy

stubailo:
I do think merging public-facing stuff into the reference implementation which is not spec compliant is an issue, although a minor one, since there is a very real risk of people coding their API to GraphQL.js, and using things that don't ever end up in the spec.

I don't think this issue is related though because the whole SDL isn't in the spec, but if the SDL were in the spec I'd be against adding random changes to it in GraphQL.js.

PR #326, opened 2017-06-19
RFC: Scalar serialize as built-in scalar type

leebyron:
Currently, Custom Scalars only describe their name and no other behavior about them can be programmatically determined. However in practice most custom scalar types are specializations of one of the existing built-in scalar types (String, Int, Float, etc.).

This proposes adding one additional piece of metadata to custom scalars, a "serializes as" type which must be a built-in scalar type. This serialize as type describes what kinds of values are allowed to be provided as input to that type and what kind of value is expected to serialize from that type. This metadata is useful for code-generation systems which expect to generate code in well typed environments. It can also improve the validation accuracy and IDE ergonomics by providing a similar level of information about custom scalars as exist for the built-in scalars.

To provide this metadata to those kinds of tools, this also extends the introspection system to return the serialize as type in the `ofType` field.

This spec text has a reference implementation here: https://github.com/graphql/graphql-js/pull/914/

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -76,13 +76,7 @@ response format being used, there may be a more appropriate primitive for the
 given scalar type, and server should use those types when appropriate.
 
 GraphQL provides a number of built-in scalars, but type systems can add
-additional scalars with semantic meaning. For example, a GraphQL system could
-define a scalar called `Time` which, while serialized as a string, promises to
-conform to ISO-8601. When querying a field of type `Time`, you can then rely on
-the ability to parse the result with an ISO-8601 parser and use a
-client-specific primitive for time. Another example of a potentially useful
-custom scalar is `Url`, which serializes as a string, but is guaranteed by
-the server to be a valid URL.
+additional custom scalars with specific semantic meaning.
 
 A server may omit any of the built-in scalars from its schema, for example if a
 schema does not refer to a floating-point number, then it will not include the
@@ -132,8 +126,8 @@ For all types below, with the exception of Non-Null, if the explicit value
 **Built-in Scalars**
 
 GraphQL provides a basic set of well-defined Scalar types. A GraphQL server
-should support all of these types, and a GraphQL server which provide a type by
-these names must adhere to the behavior described below.
+should support all of these types, and a GraphQL server which provides a type by
+these names must adhere to the behavior described for that type.
 
 
 #### Int
@@ -246,6 +240,63 @@ When expected as an input type, any string (such as `"4"`) or integer (such
 as `4`) input value should be coerced to ID as appropriate for the ID formats
 a given GraphQL server expects. Any other input value, including float input
 values (such as `4.0`), must raise a query error indicating an incorrect type.
+
+
+### Custom Scalars
+
+In addition to GraphQL's built-in scalars, a type system may add additional
+custom scalars which provide specific semantic meaning.
+
+For example, a type system might define a scalar called `Time` which, while
+serialized as `String`, promises to conform to ISO-8601. When querying a field
+of type `Time`, you can then rely on the ability to parse the result with an
+ISO-8601 parser and use a client-specific primitive for time.
+
+Another example of a potentially useful custom scalar is `Url`, which serializes
+as `String`, but is guaranteed by the server to be a valid URL.
+
+**Serialize as type**
+
+A custom scalar may provide a built-in scalar type which describes how the
+custom scalar is serialized. This provides more information for how execution
+responses are query variable inputs are coerced. If provided, this built-in
+scalar can be used both for improved validation results and stricter guarantees
+during execution.
+
+While a custom scalar type is not required to provide a serialize as type, if
+one is provided it must be one of the built-in scalar types (`Int`, `Float`,
+`String`, `Boolean`, `ID`).
+
+**Result Coercion**
+
+If a custom scalar provides a built-in scalar type which it serializes as, then
+that custom scalar must coerce using the same rules as that built-in scalar
+type. Using the example above, if a custom scalar `Time` declares it serializes
+as a `String`, then the result of a field of type `Time` must produce a
+`String`, which may be subject to additional coercion rules specific to the
+custom scalar type.
+
+If a custom scalar does not provide a built-in scalar type which it serializes
+as, then since this coercion behavior is not observable to clients of the
+GraphQL server, the precise rules of coercion are left to the implementation,
+and the serialized form will not be known ahead of time.
+
+**Input Coercion**
+
+If a custom scalar provides a built-in scalar type which it serializes as, then
+that custom scalar should accept the same kinds of input which that built-in
+scalar type accepts and must produce an error for the kinds of input for which
+that built-in scalar type would produce an error. Using the example above, if a
+custom scalar `Time` declares it serializes as a `String`, then it should accept
+`String` values and must not accept `Int` or `Float` values as input.
+
+A custom scalar may provide additional rules for coercing input values as
+implementation details, producing a query error when it deems necessary.
+
+If a custom scalar does not provide a built-in scalar type which it serializes
+as, then since this input coercion behavior is not observable to clients of the
+GraphQL server, the precise rules of coercion are left to the implementation,
+and the allowed input values will not be known ahead of time.
 
 
 ### Objects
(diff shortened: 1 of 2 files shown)


stubailo:
> a "serializes as" type which must be a built-in scalar type.

What if I'm using a transport which supports additional types, like ordered maps or something, which aren't one of the built in types? Also, this wouldn't support the currently widely used `JSON` scalar type.

I think this is a great proposal, but it's too restrictive if all scalars have to be serializable in this way. It would be awesome if it were optional, so you could get default type generation but only for the types that have this specified.

OlegIlyenko:
in general i like this addition, Though I have concerns about this part:

> if one is provided it must be one of the built-in scalar types (`Int`, `Float`, `String`, `Boolean`, `ID`).

Do you think it would make sence to relax this rule and allow custom scalar values to be defined in terms of other custom scalar values? I do see that it can simplify this feature from client perspective, but there is still some value to be gained when server is able to describe nested constraints.

That said, i just checked how I defined `ScalarAlias` in sangria, and it turs out that it does not allow nested dependencies either (so you can't define a scalar alias in terms of another scalar alias) :) Though scalar aliases serve a bit different purpose.

On the other hand, if we don't allow nested definitions, then i'm a buit lost. From [this description](https://github.com/facebook/graphql/pull/308#issuecomment-308913177):

> as the real value extracted by doing this is to allow clients to understand the serialized response. In that sense, we do actually want to consider this in terms of the semantic scalar values. 

I undestood that the whole point of this feature is to focus on the sematic meaning of the scalar value. By restricting it to only built-in scalars, we greatly deminish it's value, in my opinion (built-in scalars do not provide much of veriety). In this case I feel that it actually would be better to expose `literal` information instead. This has an a few advantages:

* we would be able to describe all built-in types as well as any custom scalar type
* base type question is not relevant anymore sice we only have a limited set of literals

Still not 100% about soundness of my argument. Would love to hear your thoughts on this :)

stubailo:
Sorry, I totally misread! Thanks for the clarification @IvanGoncharov

leebyron:
@OlegIlyenko 

> Do you think it would make sence to relax this rule and allow custom scalar values to be defined in terms of other custom scalar values? I do see that it can simplify this feature from client perspective, but there is still some value to be gained when server is able to describe nested constraints.

I think @IvanGoncharov made a good point above about this. Allowing custom scalars to be defined in terms of other custom scalars essentially produces a chain, where the end of that chain is either something that client tooling understands or something it does not. Ultimately that tooling needs to resolve the chain and determine if it has the knowledge to do anything more specific.

The motivation for restricting to the built-in scalar types is that any other type a client tool won't know what to do with, and biasing towards a simpler solution protects us from incidental complexity in the future. But I'm definitely open to discussion on this. I would be curious to hear of an example where the nested constraints would be useful for solving a problem which would weigh in favor of a slightly more complicated solution.

> I undestood that the whole point of this feature is to focus on the sematic meaning of the scalar value. By restricting it to only built-in scalars, we greatly deminish it's value, in my opinion (built-in scalars do not provide much of veriety). In this case I feel that it actually would be better to expose literal information instead.

I do think there are two related but different problems here. 

The one that drove this proposal is client-side tooling which seek to understand how to treat custom scalars. What should client tooling do with a `URL` or `DateTime` or the like? Explaining how they serialize in terms of existing scalars allows client tools to only need to understand the built-in scalars and via this addition understand how to treat nearly all other custom scalars as well. This can also handle literals as well, that anything which serializes as String should accept String literals.

A different but related problem is how to define custom scalars which actually expose *new* serialization behavior. For example a `Long` 64-bit int value. It would not be correct to say that this serializes as Int, however it would be useful to say that it accepts INT literals to incrementally improve validation.

My opinion is that the first problem is more important to solve since client code-gen has become important for nearly all popular GraphQL client frameworks (Apollo, Relay, and FB's native apps), where defining custom scalars in terms of the literals they accept rather than how they serialize would not help. 

The second problem is interesting in its ability to allow GraphQL to grow to support more kinds of serialization behavior, but is not as much a problem encountered by clients now and to solve would likely require more than just describing the literals that type accepts.

Is that a fair characterization?

OlegIlyenko:
Yes, this sounds good! After thinking about it fo a while, I also think that it is a good idea to approach this from a practical perspective. Approaching it from a client and codegen perspective is quite helpful. At least for numbers, these tools need to know the actual size of a number, the literal information is not very helpful here. So even limited subset of semantic information is already quite helpful.

This gives me another idea: should we distinguish between input and output base type? for a `Long` it would be quite helpful - it can declare itself as an `Int`-compatible on the input side, but undefined on the output side. the same is for `BigInt` and `BigDecimal`. Not sure whether this information is worth added complexity. if a codegen tool encounters such type, it probably will need a dedicated mapping anyways to fully support such scalar type.

wincent:
I know this is an early draft, but from the word-smithery perspective some of the wording here is a bit awkward. In a few places we have this structure:

> Custom scalars may provide a built-in scalar type it serializes as in `ofType`, which describe primary data format and scalar coercion rules.

The "which it serializes as" pattern is a bit of a mouthful and I suspect may be harder to read for English-as-a-second-language readers. I'd suggest reordering these to avoid the trailing "as". In the example above, for instance, we could say:

> Custom scalars may indicate how they serialize by returning a built-in scalar type from `ofType`.

(Or similar.)

Likewise:

> If a custom scalar provides a built-in scalar type which it serializes as, ...

could become:

> If a custom scalar serializes as a built-in scalar type, ...

IvanGoncharov:
As we discussed during the last WG meetup there was no major objection to this proposal.

However, I want to raise my concern about SDL syntax. With the proposed change you will be able to define custom scalars in two ways:
1) `scalar SomeScalarType` -  to define scalars with new serialization behavior e.g. 64-bit integers. 
2) `scalar SomeScalarType as String` - to define domain-specific scalars as `HTML`, `CreditCardNumber`, etc. as serializable to one of the standard scalars.

I'm concerned that developers may use the first form (because it shorter) for the cases where the second form is the correct choice.
I think it's bad to create syntax that encourages developers to define scalars with custom serialization since in most of the cases it's not what developer intend. Again, I think first form would be used because it short and because you need to know that the second form exists.

Also, I think the majority of APIs either don't need scalar with custom serialization behavior or they use scalars provided by GraphQL implementations (e.g. Sangria's `Long`).

So there is an issue: the short form (`scalar SomeScalarType`) is used for more complex and infrequent case and the long form (`scalar SomeScalarType as String`) is used for the simple and very common case.

I expect the following example scenario:
1) a developer defines API with scalars like `scalar Quantity` and assumes it's integer >= 0. But he doesn't know about `as Int` part and everything works just fine without it.
2) runs some codegen tool that assumes that either:
  *`Quantity` is a string (e.g. [apollo-ios](https://github.com/apollographql/apollo-ios/issues/23)) 
  * produces error 
3) developer either decides that defining scalars was a bad idea and just uses `Int` instead or
spends time googling just to find out about the second form.

I strongly believe that the syntax for defining scalars with the new serialization behavior should signalize (like `dangerouslySetInnerHTML` in React) that by doing so, you create problems to code generation, some of the tools, etc

 I'm still not sure about how alternative syntax may look like but here are a few of the possible variants:

```graphql
scalar SomeScalarType asCustom
scalar SomeScalarType as Custom
scalar SomeScalarType as Any
custom scalar SomeScalarType
```
Would be great if someone can come up with a better one.

I think it is still possible to introduce such changes to the SDL because it's not merged into the spec yet and because [Descriptions as strings PR](https://github.com/graphql/graphql-js/pull/927) also introduces breaking change.

What do you think about it?

P.S. @leebyron Would be great to discuss it before `Descriptions as strings` PR is merged.

schickling:
Is there any further movement on this @IvanGoncharov? I remember you championing this. 👑

IvanGoncharov:
@schickling Thanks for ping 👍 
I pretty happy with PR and it's current state.
It's look like next step would be implementing it inside graphql-js.
I rebased https://github.com/graphql/graphql-js/pull/914 as https://github.com/graphql/graphql-js/pull/1173 and will work on adding more tests.

But I'm still concern about SDL problem I described in my previous comment and think it should be disscussed before SDL syntax will be set in stone.
What do you think about it and proposed solution?

IvanGoncharov:
As suggested by @leebyron during last WG I'm adopting this PR as #521

PR #328, opened 2017-07-03
Update copyright year

kevinSuttle:
Because someone has to

### DIFF
--- a/spec/GraphQL.md
+++ b/spec/GraphQL.md
@@ -13,7 +13,7 @@ enhancement will continue in future editions of this specification.
 
 **Copyright notice**
 
-Copyright (c) 2015-2016, Facebook, Inc. All rights reserved.
+Copyright (c) 2015-2017, Facebook, Inc. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
 are permitted provided that the following conditions are met:


wincent:
Sure, although we should probably just change the end of the range to "present", once and for all. We do this in a lot of our projects ([example](https://github.com/facebook/relay/blob/ef5ab894fc5279e78bb36b077803f7c5b196a5e5/LICENSE#L5)). I'll merge this in any case as it is still better than having a stale year in there.

kevinSuttle:
Good point!

eriknstr:
Regarding the practice of setting the end of the range to "present", is this something that Facebook lawyers have okayed, or was that something developers came up with? IMO it's unfortunate for a project to be using "present" because it makes it more difficult to determine in the future what years apply to a full or partial snapshot of some project, e.g. if a file from one project is copied into another project in accordance with the terms of the license of the originating project.

wincent:
@eriknstr:

> Regarding the practice of setting the end of the range to "present", is this something that Facebook lawyers have okayed, or was that something developers came up with?

It's all over our internal codebases, so is an established practice here.

> IMO it's unfortunate for a project to be using "present" because it makes it more difficult to determine in the future what years apply to a full or partial snapshot of some project, e.g. if a file from one project is copied into another project in accordance with the terms of the license of the originating project.

In general we try to avoid copying files from one project to another, but in any case, version control is probably the most reliable source of information here (compared to a possibly stale, human-maintained code comment): you can see when a file entered the project, and when it was modified, pretty trivially with both `git` on the command-line and in the UI, along with the generally useful context that you'll get from whatever commit messages apply.

PR #330, opened 2017-07-12
Master

wanghaisheng:


wanghaisheng:
done

wanghaisheng:
i  am keenly  looking for others contributing to translation and disseminate graphQL

wincent:
Thanks for this @wanghaisheng. I think the approach that we want to take to internationalizing the docs here is going to be similar to the one taken by other FB projects like [Jest](https://github.com/facebook/jest). They store only the English docs in the source branch of the repo, and then use [Crowdin](https://crowdin.com/) as a platform to manage the translations (which at build-time get written out to the `gh-pages` branch).

So I don't think we'd want to proceed with this PR as-is (and even if we did, we'd want to squash the history to remove the extraneous commits and unwanted changes such as setting the executable bit on image files etc) but rather do this the "right" way and set up Crowdin. Let me create another issue for that and we can continue the discussion there. Hopefully we can leverage the work that you've already done.

PR #332, opened 2017-07-20
Fix grammatical error

RichardLitt:
Verbal number must agree with subject.


wincent:
Good catch. Thanks @RichardLitt.

PR #345, opened 2017-08-18
@skip Directive example sentence wrong

burner:
https://github.com/facebook/graphql/pull/344 was against the wrong branch


wincent:
Thanks @burner. That reads quite a bit more nicely.

burner:
@wincent nice spec, glad to help

PR #346, opened 2017-08-18
@include Directive example sentence wrong

burner:
fix copy paste error in the same way as https://github.com/facebook/graphql/pull/345


wincent:
Thanks again @burner.

PR #384, opened 2017-11-21
Change order of fields inside Response

IvanGoncharov:
Currently, the spec requires `data` to be the first field in a response.
But after [my PR](https://github.com/graphql/graphql-js/pull/838) was merged into `graphql-js`, `errors` always precede `data`:
image

I still think that having `errors` as the first property is a big DX improvement so I updated the spec to reflect the change in `graphql-js`.

### DIFF
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -85,15 +85,15 @@ objects in the response as unordered Maps and arrive at a valid value.
 
 A response to a GraphQL operation must be a map.
 
-If the operation included execution, the response map must contain a first entry
-with key `data`. The value of this entry is described in the "Data" section. If
-the operation failed before execution, due to a syntax error, missing
-information, or validation error, this entry must not be present.
-
-If the operation encountered any errors, the response map must contain a next
+If the operation encountered any errors, the response map must contain a first
 entry with key `errors`. The value of this entry is described in the "Errors"
 section. If the operation completed without encountering any errors, this entry
 must not be present.
+
+If the operation included execution, the response map must contain a next entry
+with key `data`. The value of this entry is described in the "Data" section. If
+the operation failed before execution, due to a syntax error, missing
+information, or validation error, this entry must not be present.
 
 The response map may also contain an entry with key `extensions`. This entry,
 if set, must have a map as its value. This entry is reserved for implementors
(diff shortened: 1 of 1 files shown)


acjay:
Why should the order matter? I like having `errors` first, but it seems arbitrary. And what of JSON libraries that leave order undefined?

jlouis:
I've always found the ordering requirements odd.

At most, the specification can put a *should* on them, because JSON encoders are free to do whatever they want. If the internal representation of a JSON dict is a large hash table, then the rendering order is going to be arbitrary. Furthermore, some languages (Go is one example) randomizes the order in which you visit a hash table in order to make it explicit to the programmer that they cannot rely on the ordering.

In short, requiring order here is rather brittle and is bound to create problems down the line.

IvanGoncharov:
@acjay @jlouis GraphQL spec only recommends order of properties, for more details see 
https://github.com/facebook/graphql/commit/b22c73c875807e0cbb2c1d51804311e42cba0a36

This PR just makes `graphql-js` (reference implementation) in line with spec recommendation :)

stubailo:
I think order requirements in the spec are for parsers which rely on the order to more efficiently parse the result. IMO that makes errors at the front even more valuable, since it is super useful to know about errors first in the case of a huge result. If it's framed as a recommendation it shouldn't be an issue since things are still valid GraphQL even if they return data first.

IvanGoncharov:
> If it's framed as a recommendation it shouldn't be an issue since things are still valid GraphQL even if they return data first.

You right. But I think the spec should recommend only best practices and error first approach provide better UX. Moreover spec recommendation should be implemented at least in the reference implementation.

stubailo:
Yep, I just think there's a big difference between "recommend" vs. "require" in a spec! I totally agree errors first is the best approach.

leebyron:
"A response to a GraphQL operation must be a map." does not imply order (otherwise it would say "ordered map") - I agree that the "first" and "next" terminology is more ambiguous than helpful in this context, though I like declaring the order via the order of paragraphs.

I think we can add a more explicit phrase about ordering errors first in a formatted output as a convenience

IvanGoncharov:
@leebyron Great improvment 👍 
Hope it will be merged soon.

PR #388, opened 2017-12-02
Rewrite the Input Object input coercion subsection

### LINKED ISSUE #272: Input Objects Example is Invalid
Hello,

Great document! Thank you!

I am parsing my way through the [Input Object, Input Coercion](http://facebook.github.io/graphql/#sec-Input-Objects) section.  It states input definition of 
```graphql
input ExampleInputObject {
  a: String
  b: Int!
}
```
and says an input of 
```graphql
{ 
  a: 123, 
  b: "123"
}
```
can be coerced into 
```graphql
{ 
  a: "123", 
  b: 123
}
```
.

I find this example to be impossible as the [String Input Coercion](http://facebook.github.io/graphql/#sec-String) does not allow for anything other than string inputs and the [Int Input Coercion](http://facebook.github.io/graphql/#sec-Int) does not allow for anything other than Int inputs.

This type change would make sense if it was a return value, but it is an input, which is very rigid.

Also, if the [Input Object Type Validation](http://facebook.github.io/graphql/#sec-Input-Object-type-validation) could be updated to include a comment about nullish values, that would be great.

Thank you,
Barret

leebyron:
This section had some redundancy and confusing statements in addition to some outdated examples for coercing field values.

This rewrite expands on the examples and consolidates the coercion rules.

Fixes #272
Closes #311

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -805,29 +805,28 @@ An input object is never a valid result.
 **Input Coercion**
 
 The value for an input object should be an input object literal or an unordered
-map, otherwise an error should be thrown. This unordered map should not contain
-any entries with names not defined by a field of this input object type,
-otherwise an error should be thrown.
+map supplied by a variable, otherwise an error should be thrown. In either
+case, the input object literal or unordered map should not contain any entries
+with names not defined by a field of this input object type, otherwise an error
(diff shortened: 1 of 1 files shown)


leebyron:
@andimarek please let me know if this resolves your original concern

PR #402, opened 2018-01-28
Clarify expected behavior for field arguments not defined in interface

### LINKED ISSUE #401: Question: expected behavior for field arguments not defined in interface
I have a question about the following clause in the draft spec (under 3.1.2.3).

> The object field may include additional arguments not defined in the interface field, but any additional argument must not be required.

http://facebook.github.io/graphql/draft/#sec-Object-type-validation

What is the expected behavior of additional arguments that have a default value? For example:
```graphql
interface Namer {
  name: String!
}

type User implements Namer {
  id: ID!
  name(type: NameType = FULL): String!
  email: String!
}

enum NameType {
  FULL
  FIRST
  LAST
}
```

In this example, does `User` implement `Namer`?

IvanGoncharov:
Fixes #401

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -564,7 +564,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          1. The object field argument must accept the same type (invariant) as
             the interface field argument.
       3. The object field may include additional arguments not defined in the
-         interface field, but any additional argument must not be required.
+         interface field, but any additional argument should be of a nullable
+         type.
 
 
 ### Interfaces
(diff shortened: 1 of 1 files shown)


tonyghita:
💯 thank you for the clarification.

spec/Section 3 -- Type System.md
```diff
@@ -564,7 +564,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          1. The object field argument must accept the same type (invariant) as
             the interface field argument.
       3. The object field may include additional arguments not defined in the
-         interface field, but any additional argument must not be required.
+         interface field, but any additional argument must not be required, e.g.
```
IvanGoncharov: @leebyron Does `e.g.` mean that there are other situations beyond argument having `non-null` type that will make this argument required?

PR #403, opened 2018-01-29
Remove number of excessive whitespaces

IvanGoncharov:
I was wondering if it's possible to use `prettier` on markdown files in this repo and [found a few bugs in prettier](https://github.com/prettier/prettier/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+author%3AIvanGoncharov) that makes it impossible to use.
But reviewing the diff I found multiple places with excessive whitespaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ this query would then be:
 ```
 
 Specifying the `query` keyword and an operation name is only required when a
-GraphQL document defines multiple operations.  We therefore could have written
+GraphQL document defines multiple operations. We therefore could have written
 the previous query with the query shorthand:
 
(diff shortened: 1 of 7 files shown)


leebyron:
Yay! We don't use typewriters anymore and modern fonts provide correct kerning pairs, so I'm happy to see this fixed.

Glad to see you take a crack and improving prettier markdown support. Once support works we should use it.

N.B. http://www.slate.com/articles/technology/technology/2011/01/space_invaders.html

benjie:
The GraphQL spec now uses prettier :raised_hands:

PR #406, opened 2018-02-09
Fix Markdown links

klzns:
This PR fixes the links navigation at Github.

Before | After
---|---
image | image


klzns:
After checking the build step, I think there's no way to fix the Github links without changing `spec-md`.

leebyron:
Thanks!

PR #418, opened 2018-03-06
RFC: Fix ambiguity with null variable values and default values

leebyron:
(Rendering Preview: https://out-nmocjoypvp.now.sh)

This is a *NON-BREAKING* **validation change** which allows some previously invalid queries. It is also a behavioral **execution change** which changes how values interact with variable and argument default values.

There is currently ambiguity and inconsistency in how `null` values are coerced and resolved as part of variable values, default values, and argument values. This inconsistency and ambiguity can allow for `null` values to appear at non-null arguments, which might result in unforseen null-pointer-errors.

The version of this proposal to be merged includes the following:

* *NEW*: Non-null arguments and input object fields are no longer required to be provided a value/variable if that argument also supplies a default value.
* *NEW*: Similarly, non-null input object fields can be omitted if they supply a default value. (Note that previously this validation clause was missing from the spec, however has long been included in reference implementations)
* *NEW*: All variables can now have default values by way of removing a validation rule. Previously it was invalid to supply a default value with a non-null variable.
* *BUG FIX*: Variables with a nullable type and a `null` default value cannot be provided to a non-null argument or input field.
* *NEW*: Optional (nullable) variables can now be used for arguments or input object fields which supply default values.
* *CLARITY*: Redefine `CoerceVariableValues()` to make it more clear how to treat a lack of runtime value vs the explicit value `null` with respect to default values and type coercion.
* *NEW*: Redefine `CoerceArgumentValues()` to ensure providing a `null` variable value to a Non-Null argument type causes a field error, and to add clarity to how `null` values and default values should be treated.
* *NEW*: Redefine Input object coercion rules to ensure providing a `null` value or runtime variable value to a non-null type causes a field error, and improve clarity for these coercion rules. This mirrors the changes to `CoerceArgumentValues()`.

---

Previous version of this proposal:

<strike>

This appears in three distinct but related issues:

**Validation: All Variable Usages are Allowed**

The explicit value `null` may be used as a default value for a variable with a nullable type, however this rule asks to treat a variable's type as non-null if it has a default value. Instead this rule should specifically only treat the variable's type as non-null if the default value is not `null`.

Additionally, the `AreTypesCompatible` algorithm is underspecificied, which could lead to further misinterpretation of this validation rule.

**Coercing Variable Values**

`CoerceVariableValues()` allows the explicit `null` value to be used instead of a default value. This can result in a null value flowing to a non-null argument due to the validation rule mentioned above. Instead a default value must be used even when an explicit `null` value is provided. This is also more consistent with the explanation for validation rule "Variable Default Value Is Allowed"

Also, how to treat an explicit `null` value is currently underspecified. While an input object explains that a `null` value should result in an explicit `null` value at the input object field, there is no similar explaination for typical scalar input types. Instead, `CoerceVariableValues()` should explicitly handle the `null` value to make it clear a `null` is the resulting value in the `coercedValues` Map.

**Coercing Argument Values**

The `CoerceArgumentValues()` algorithm is intentionally similar to `CoerceVariableValues()` and suffers from the same inconsistency. Explicit `null` values should not take precedence over default values, and should also be explicitly handled rather than left to underspecified input scalar coercion.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1825,11 +1825,12 @@ an extraneous variable.
   * For each {operation} in {document}:
   * Let {variableUsages} be all usages transitively included in the {operation}.
   * For each {variableUsage} in {variableUsages}:
-    * {IsVariableUsageAllowed(variableUsage)} must be {true}.
+    * Let {variableName} be the name of {variableUsage}.
+    * Let {variableDefinition} be the {VariableDefinition} named {variableName}
+      defined within {operation}.
+    * {IsVariableUsageAllowed(variableDefinition, variableUsage)} must be {true}.
 
-IsVariableUsageAllowed(variableUsage):
-  * Let {variableName} be the name of {variableUsage}.
-  * Let {variableDefinition} be the {VariableDefinition} in {operation}.
+IsVariableUsageAllowed(variableDefinition, variableUsage):
   * Let {variableType} be the expected type of {variableDefinition}.
   * Let {locationType} be the expected type of the {Argument}, {ObjectField},
     or {ListValue} entry where {variableUsage} is located.
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -676,6 +676,7 @@ type Arguments {
   intArgField(intArg: Int): Int
   nonNullBooleanArgField(nonNullBooleanArg: Boolean!): Boolean!
   booleanListArgField(booleanListArg: [Boolean]!): [Boolean]
+  optionalNonNullBooleanArgField(optionalBooleanArg: Boolean! = false): Boolean!
 }
 
 extend type Query {
@@ -1824,28 +1825,41 @@ an extraneous variable.
   * For each {operation} in {document}:
   * Let {variableUsages} be all usages transitively included in the {operation}.
   * For each {variableUsage} in {variableUsages}:
-    * Let {variableName} be the name of {variableUsage}.
-    * Let {variableDefinition} be the {VariableDefinition} in {operation}.
-    * Let {variableType} be {EffectiveVariableType(variableDefinition)}.
-    * Let {argumentType} be the type of the argument {variableUsage} is passed to.
-    * AreTypesCompatible({argumentType}, {variableType}) must be {true}.
+    * {IsVariableUsageAllowed(variableUsage)} must be {true}.
 
-AreTypesCompatible(argumentType, variableType):
-  * If {argumentType} is a non-null type:
+IsVariableUsageAllowed(variableUsage):
+  * Let {variableName} be the name of {variableUsage}.
+  * Let {variableDefinition} be the {VariableDefinition} in {operation}.
+  * Let {variableType} be the expected type of {variableDefinition}.
+  * Let {locationType} be the expected type of the {Argument}, {ObjectField},
+    or {ListValue} entry where {variableUsage} is located.
+  * If {locationType} is a non-null type AND {variableType} is NOT a non-null type:
+    * Let {hasNonNullVariableDefaultValue} be {true} if a default value exists
+      for {variableDefinition} and is not the value {null}.
+    * Let {hasLocationDefaultValue} be {true} if a default value exists for
+      the {Argument} or {ObjectField} where {variableUsage} is located.
+    * If {hasNonNullVariableDefaultValue} is NOT {true} AND
+      {hasLocationDefaultValue} is NOT {true}, return {false}.
+    * Let {nullableLocationType} be the unwrapped nullable type of {locationType}.
+    * Return {AreTypesCompatible(variableType, nullableLocationType)}.
+  * Return {AreTypesCompatible(variableType, locationType)}.
+
+AreTypesCompatible(variableType, locationType):
+  * If {locationType} is a non-null type:
     * If {variableType} is NOT a non-null type, return {false}.
-    * Let {nullableArgumentType} be the nullable type of {argumentType}.
-    * Let {nullableVariableType} be the nullable type of {variableType}.
-    * Return {AreTypesCompatible(nullableArgumentType, nullableVariableType)}.
+    * Let {nullableLocationType} be the unwrapped nullable type of {locationType}.
+    * Let {nullableVariableType} be the unwrapped nullable type of {variableType}.
+    * Return {AreTypesCompatible(nullableVariableType, nullableLocationType)}.
   * Otherwise, if {variableType} is a non-null type:
     * Let {nullableVariableType} be the nullable type of {variableType}.
-    * Return {AreTypesCompatible(argumentType, nullableVariableType)}.
-  * Otherwise, if {argumentType} is a list type:
+    * Return {AreTypesCompatible(nullableVariableType, locationType)}.
+  * Otherwise, if {locationType} is a list type:
     * If {variableType} is NOT a list type, return {false}.
-    * Let {itemArgumentType} be the item type of {argumentType}.
-    * Let {itemVariableType} be the item type of {variableType}.
-    * Return {AreTypesCompatible(itemArgumentType, itemVariableType)}.
+    * Let {itemLocationType} be the unwrapped item type of {locationType}.
+    * Let {itemVariableType} be the unwrapped item type of {variableType}.
+    * Return {AreTypesCompatible(itemVariableType, itemLocationType)}.
   * Otherwise, if {variableType} is a list type, return {false}.
-  * Return {true} if {variableType} and {argumentType} are identical, otherwise {false}.
+  * Return {true} if {variableType} and {locationType} are identical, otherwise {false}.
 
 **Explanatory Text**
 
@@ -1915,21 +1929,21 @@ query listToNonNullList($booleanList: [Boolean]) {
 This would fail validation because a `[T]` cannot be passed to a `[T]!`.
 Similarly a `[T]` cannot be passed to a `[T!]`.
 
-**Supporting legacy operations**
+**Allowing optional variables when default values exist**
 
-EffectiveVariableType(variableDefinition):
-  * Let {variableType} be the expected type of {variableDefinition}.
-  * If service supports operations written before this specification edition:
-    * If {variableType} is not a non-null type:
-      * Let {defaultValue} be the default value of {variableDefinition}.
-      * If {defaultValue} is provided and not the value {null}:
-        * Return the non-null of {variableType}.
-  * Otherwise, return {variableType}.
+A notable exception to typical variable type compatibility is allowing a
+variable definition with a nullable type to be provided to a non-null location
+as long as either that variable or that location provides a default value.
 
-A notable exception to type compatibility a GraphQL service may support is
-allowing a variable definition with a nullable type and provided default value
-to be provided to a non-null position as long as that variable's default value
-is also not the value {null}.
+```graphql example
+query booleanArgQueryWithDefault($booleanArg: Boolean) {
+  arguments {
+    optionalNonNullBooleanArgField(optionalBooleanArg: $booleanArg)
+  }
+}
+```
+
+In the example above, an optional variable is allowed to be used in an non-null argument which provides a default value.
 
 ```graphql example
 query booleanArgQueryWithDefault($booleanArg: Boolean = true) {
@@ -1939,11 +1953,10 @@ query booleanArgQueryWithDefault($booleanArg: Boolean = true) {
 }
 ```
 
-Earlier editions of this specification explicitly allowed such operations due to
-a slightly different interpretation of default values and {null} values. GraphQL
-services accepting operations written before this edition of the specification
-may continue to allow such operations, however GraphQL services established
-after this edition should not allow such operations.
+In the example above, a variable provides a default value and can be used in a
+non-null argument. This behavior is explicitly supported for compatibility with
+earlier editions of this specification. GraphQL authoring tools may wish to
+report this is a warning with the suggestion to replace `Boolean` with `Boolean!`.
 
 Note: The value {null} could still be provided to a such a variable at runtime.
 A non-null argument must produce a field error if provided a {null} value.
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1836,17 +1836,33 @@ an extraneous variable.
   * For each {operation} in {document}
   * Let {variableUsages} be all usages transitively included in the {operation}
   * For each {variableUsage} in {variableUsages}
-    * Let {variableType} be the type of variable definition in the {operation}
-    * Let {argumentType} be the type of the argument the variable is passed to.
-    * Let {hasDefault} be true if the variable definition defines a default.
-    * AreTypesCompatible({argumentType}, {variableType}, {hasDefault}) must be true
+    * Let {variableName} be the name of {variableUsage}
+    * Let {variableDefinition} be the {VariableDefinition} in {operation}.
+    * Let {variableType} be the type of {variableDefinition}.
+    * Let {defaultValue} be the default value of {variableDefinition}.
+    * If {variableType} is not a non-null and {defaultValue} is provided and not {null}:
+      * Let {variableType} be the non-null of {variableType}.
+    * Let {argumentType} be the type of the argument {variableUsage} is passed to.
+    * AreTypesCompatible({argumentType}, {variableType}) must be {true}.
 
-  * AreTypesCompatible({argumentType}, {variableType}, {hasDefault}):
-    * If {hasDefault} is true, treat the {variableType} as non-null.
-    * If inner type of {argumentType} and {variableType} are different, return false
-    * If {argumentType} and {variableType} have different list dimensions, return false
-    * If any list level of {variableType} is not non-null, and the corresponding level
-      in {argument} is non-null, the types are not compatible.
+AreTypesCompatible(argumentType, variableType):
+  * If {argumentType} is a non-null type:
+    * If {variableType} is a non-null type:
+      * Let {nullableArgumentType} be the nullable type of {argumentType}.
+      * Let {nullableVariableType} be the nullable type of {variableType}.
+      * Return {AreTypesCompatible(nullableArgumentType, nullableVariableType)}.
+    * Otherwise return {false}.
+  * If {variableType} is a non-null type:
+    * Let {nullableVariableType} be the nullable type of {variableType}.
+    * Return {AreTypesCompatible(argumentType, nullableVariableType)}.
+  * If {argumentType} is a list type:
+    * If {variableType} is a list type:
+      * Let {itemArgumentType} be the item type of {argumentType}.
+      * Let {itemVariableType} be the item type of {variableType}.
+      * Return {AreTypesCompatible(itemArgumentType, itemVariableType)}.
+    * Otherwise return {false}.
+  * If {variableType} is a list type return {false}.
+  * Return if {variableType} and {argumentType} are identical.
 
 **Explanatory Text**
 
@@ -1890,8 +1906,8 @@ query booleanArgQuery($booleanArg: Boolean) {
 }
 ```
 
-A notable exception is when default arguments are provided. They are, in effect,
-treated as non-nulls.
+A notable exception is when a default value are provided. They are, in effect,
+treated as non-nulls as long as the default value is not {null}.
 
 ```graphql example
 query booleanArgQueryWithDefault($booleanArg: Boolean = true) {
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -727,9 +727,8 @@ and invalid.
 
 **Explanatory Text**
 
-Arguments can be required. If the argument type is non-null and does not have a
-default value, the argument is required and furthermore the explicit value
-{null} may not be provided. Otherwise, the argument is optional.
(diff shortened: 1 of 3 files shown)


leebyron:
Since this changes behavior, I'd love your feedback @IvanGoncharov, @OlegIlyenko, et al

andimarek:
cc @bbakerman

IvanGoncharov:
@leebyron I fully agree with the changes from `Validation: All Variable Usages are Allowed` 👍

> Coercing Variable Values
> Coercing Argument Values

> Explicit null values should not take precedence over default values.

This is very confusing at the first glance, especially for argument values. Here is the example from JS:
```js
function f(arg = 'foo') {
  return arg;
}

f()      // 'foo'
f(null)  // null
f('bar') // 'bar'
```
And same in Python and probably all other langs so I would expect the same from GraphQL.
But I fully understand technical challenge here so I need to think more about it.

leebyron:
Absolutely correct, @IvanGoncharov. There are two paths forward to solve for the ambiguity, each of which have some tradeoffs.

The path I'm proposing here preserves the ability to treat `$var: String = "default"` as non-nullable and usable in an argument expecting `String!`. The tradeoff is that a default value would take precedence over an explicit `null` value.

The alternative is to *no longer* treat `$var: String = "default"` as non-nullable while preserving explicit `null` over default values. This would mean queries which pass variables with default values to arguments expecting non-null types would no longer be valid.

My concern is that the change cost of this alternative is too high since passing a variable with a default value to a non-null argument is a fairly common pattern. However I'm less certain how common overriding a default value with an explicit `null` value is. I haven't found an example of it within queries at Facebook yet, so I would love to hear use cases to better understand change cost.

OlegIlyenko:
It's an interesting change. I remember the time when I finally correctly implemented the support for the `null` in sangria. As far as I remember, my initial implementation was doing the same as was suggested in this PR: "Explicit null values should not take precedence over default values.". Later on I refactored it in order to comply with the reference implementation.

Along the way, I also faced these issues where null value can potentially appear in nullable the argument with a default value (which might be treated as not-null since it has a default). As a solution to this problem I introduced following 2 things:

* I delay the value resolution to the last moment. This led to the introduction of a new [`Trinary`](https://github.com/sangria-graphql/sangria/blob/cb92d9b510690609de4f8fd65e9a4bfa1712dc93/src/main/scala/sangria/execution/ValueCoercionHelper.scala#L478-L494) container which able to preserve information about the default value.
* By default nullable arguments with a default value are treated as non-null on a type level. But since I have a `Trinary` value around I can force the usage of a default value. I also provide an extra function to get a nullable variation of the value (`Option[Value]`) - this one behaves in the same way as reference implementation at the moment. 

I find this solution quite robust - by default, it behaves similarly to the proposal in this PR, but it also provides a way to propagate `null` when this information is significant.

Though it is still possible to trick the system by setting the default value to `null`, like here:

```graphql
type Query {
  article(id: ID = null): Atricle
}
```

Do you think it would make sense to disallow this? (for arguments, input fields, and variables)

I personally don't have practical insights on API that takes advantage of the `null` value. I know that some people use it to explicitly "unset" fields in the database.  But in this scenario, I don't really see the need for a default value. So it's hard for me to assess the impact of this change, but I find the use case quite niche.

IvanGoncharov:
@leebyron It's not an ideal solution since we are trying to fix validation issue by changing execution behavior. Ideally, we would allow defaults to be specified on nonull arguments:
```graphql
query foo($arg: String! = "foo")

{} => $arg === "foo"
{ arg: null } => Exception

query bar($arg: String = "bar")

{} => $arg === "bar"
{ arg: null } => $arg === null
```
But it's not a viable alternative since it will break huge number of existing queries 😢 

So practically speaking if there is no GraphQL APIs that distinguish between `null` and default value as field argument then I don't see a point to fight for theoretical purity. 

That said I think it's very important question so we need to be extremely careful and do some research beyond the Facebook use case.
I volunteer to do research on a major public APIs(Shopify, Yelp, ...).

GitHub was my first victim and tried to pass nulls in every field with a default argument. A few of them treat `null` the same way as absent value but in one case it actually broke query:
image
Moreover, in GitHub schema `repository` defined as nullable so absence of this field inside `data` suggests that it wasn't simple exception inside resolver.

eapache:
IIUC this is ~ the same issue as https://github.com/facebook/graphql/issues/359.

I also find the "explicit nulls don't take precedence over defaults" confusing. Per the issue I originally filed, I think this can be solved by by allowing non-null variables/arguments/input-fields to have default values. This is I think what you mean by
> The alternative is to no longer treat $var: String = "default" as non-nullable while preserving explicit null over default values. This would mean queries which pass variables with default values to arguments expecting non-null types would no longer be valid.

The change cost of that approach is real, but the migration is fairly trivial and I think important: anywhere that this ambiguity currently occurs the author is forced to resolve it by adding a `!` to the variable to make it clear that explicit nulls are forbidden. If they did intend to support nulls for that variable then their query is invalid no matter what and that's something they should be aware of.

spec/Section 5 -- Validation.md
```diff
@@ -1358,6 +1359,31 @@ For example the following query will not pass validation.
 ```
 
 
+### Input Object Required Fields
```
leebyron: This rule exists as part of GraphQL.js's `ValuesOfCorrectType` but was missing from the spec

spec/Section 5 -- Validation.md
```diff
@@ -1494,44 +1520,6 @@ fragment HouseTrainedFragment {
 ```
 
 
-### Variable Default Value Is Allowed
```
leebyron: This rule is completely removed since the queries it prohibited we explicitly wish to support.

spec/Section 5 -- Validation.md
```diff
@@ -1836,17 +1824,33 @@ an extraneous variable.
...
-    * If any list level of {variableType} is not non-null, and the corresponding level
-      in {argument} is non-null, the types are not compatible.
+    * Let {variableName} be the name of {variableUsage}
+    * Let {variableDefinition} be the {VariableDefinition} in {operation}.
+    * Let {variableType} be the type of {variableDefinition}.
+    * Let {defaultValue} be the default value of {variableDefinition}.
+    * If {variableType} is not a non-null and {defaultValue} is provided and not {null}:
+      * Let {variableType} be the non-null of {variableType}.
+    * Let {argumentType} be the type of the argument {variableUsage} is passed to.
+    * AreTypesCompatible({argumentType}, {variableType}) must be {true}.
```
leebyron: Most of this change is for clarity, however the default value is now checked for {null} first

spec/Section 5 -- Validation.md
```diff
@@ -1836,17 +1824,33 @@ an extraneous variable.
...
+    * Let {nullableVariableType} be the nullable type of {variableType}.
+    * Return {AreTypesCompatible(argumentType, nullableVariableType)}.
+  * If {argumentType} is a list type:
+    * If {variableType} is a list type:
+      * Let {itemArgumentType} be the item type of {argumentType}.
+      * Let {itemVariableType} be the item type of {variableType}.
+      * Return {AreTypesCompatible(itemArgumentType, itemVariableType)}.
+    * Otherwise return {false}.
+  * If {variableType} is a list type return {false}.
+  * Return if {variableType} and {argumentType} are identical.
```
leebyron: Other than the checking default values being removed from this algorithm, this is all for clarity and doesn't signify any other behavioral change

leebyron:
Excellent feedback everyone. I've made serious revisions which I hope both protect against allowing null values to non-null arguments while preserving existing behavior. The result is a bit broader in scope, however no longer breaks existing queries in either validation or execution.

The reference PR has also been updated in graphql/graphql-js#1274

This update better defines the difference between a "required" and "non-null" argument / input-field as a "non-null" type which includes a default value is no longer required. As such the validation rule which prohibited queries from using non-null variables and default values has been removed completely. This change also adds clarity to the input field validation - this rule has existed in the GraphQL.js reference implementation however was found missing within the spec. It essentially mirrors the change of the required argument rule.

This also updates the `CoerceVariableValues()` and `CoerceArgumentValues()` algorithms to retain the behavior of explicit null values overriding a default value (minimizing breaking changes), however critically adding additional protection to `CoerceArgumentValues()` to explicitly block null variable values from passing into a non-null argument. This retains the existing common pattern of passing a nullable variable with a default value into a non-null argument while removing the problematic case of an explicit null value at runtime.

I'd love another look at this and the accompanying PR. I feel confident in this approach but would appreciate your feedback.

eapache:
I'm a bit torn - on the one hand I think this approach is probably the best we can do in terms of a clarification that is strictly non-breaking. On the other hand I'm not sure that strictly non-breaking is something we should be striving for in this case:
- The last published spec is still marked as a draft, with no implicit or explicit compatibility guarantees.
- Even Golang (whose [explicit compatibility guarantee](https://golang.org/doc/go1compat) is one I use as a good model) reserves the right to clarify undefined or inconsistent behaviour in a way that may break programs.
- The "from-scratch" design (if we weren't concerned about compatibility at all) is I think fairly uncontroversially to decouple default values from the nullability of arguments entirely. For anyone new to the spec, that behaviour seems the least surprising, and much simpler than the behaviour proposed here.

I realize that putting a breaking change like that in the spec causes its own huge set of pains for implementations that have to migrate, but I guess I'm not convinced that isn't still the best long-term solution.

All that said, I do believe this current approach is the best we can do while being non-breaking, so if that's the way that we want to go then I won't object further.

leebyron:
@eapache - great points. I actually think what I now have here is very close to what I would propose in a "from scratch" design.

The execution behavior is almost exactly as I would expect it in a from-scratch design as it respects explicit values over default values in the same way as other programming languages. The only difference I might have expected in a from-scratch design is to be even more trusting of variable values in `CoerceArgumentValues()`, where this proposal explicitly checks for null values and throws field errors. This isn't trading on correctness however, so I'm not concerned at all about that extra check. This is still *technically* a "breaking change" by the definition that execution behavior of existing queries will now be different. Previously null values could be provided to non-null typed arguments; after this proposal this is an error. I think it's pretty uncontroversial to explain this as a bug fix.

The validation behavior is really where we've made specific considerations to avoid breaking existing queries, however this proposal is still very close to what I would propose in a "from scratch" design. 

This proposal makes 3 important changes to query validation:

* Removal of the "Variable Default Value Is Allowed" rule. I argue this should have been removed when we introduced explicit null values with the expectation that they supersede default values. This rule originally sought to prevent "dead code" style misunderstandings, however in the present context it does not succeed at that and instead it prevents real possibilities. In a "from scratch" design I would include this change exactly as is.

* Changes the definition of "required" arguments and input fields. Previously a non-null type meant that an argument or input field was required to be provided. However because arguments may have default values, this limits their expressivity in the same was the previous rule. This proposal changes this such that non-null typed arguments/input-fields are only considered required if they *also* do not have a default value. This, again, I would propose unchanged in a "from scratch" design.

* Change the "Variable Usages are Allowed" rule tweaking the rules for treating nullable variables as non-null for the sake of determining where they're allowed to be used. This is *not* what I would do in a from-scratch design. Instead I would propose removing the ability for nullable variables to flow into non-null arguments *at all*. Where the previous two changes expanded the possible set of queries that pass validation, this change would have restricted that possible set. This is also, unfortunately, a very common pattern so that restriction could be a very costly breaking change. However, I definitely agree that enforcing that strictness for newly produced GraphQL services would be valuable! 

I will look at editing this to both make it more clear where the caveat for existing systems is while making it possible (or even encouraged) for new systems not to adopt that caveat.

@eapache are there other differences from the current state of this PR that you would consider approaching differently in a from-scratch design?

eapache:
> Change the "Variable Usages are Allowed" rule tweaking the rules for treating nullable variables as non-null for the sake of determining where they're allowed to be used. This is not what I would do in a from-scratch design. Instead I would propose removing the ability for nullable variables to flow into non-null arguments at all.

Yup, this is the one that is bugging me :)

> I will look at editing this to both make it more clear where the caveat for existing systems is while making it possible (or even encouraged) for new systems not to adopt that caveat.

Awesome. I was going to suggest that we include an RFC 2119 header in the spec so we can use SHOULD/MAY appropriately here, but I note that you already did that, it just hasn't made it into http://facebook.github.io/graphql/ yet. I think that servers `SHOULD` prevent nullable variables from flowing into non-null arguments, but they `MAY` permit it in accordance with old versions of the spec in order to maintain compatibility. Thoughts?

> are there other differences from the current state of this PR that you would consider approaching differently in a from-scratch design?

I'll go through again but I think that was the only one.

spec/Section 5 -- Validation.md
```diff
@@ -710,25 +710,26 @@ and invalid.
...
       * {value} must not be the {null} literal.
 
 **Explanatory Text**
 
-Arguments can be required. If the argument type is non-null the argument is
-required and furthermore the explicit value {null} may not be provided.
-Otherwise, the argument is optional.
+Arguments can be required. If the argument type is non-null and does not have a
+default value, the argument is required and furthermore the explicit value
+{null} may not be provided. Otherwise, the argument is optional.
```
eapache: I read this to imply (probably unintentionally) that a non-null argument with a default value is allowed to be passed an explicit `null`?

leebyron:
@eapache - I just updated this with your feedback incorporated. I'm glad we agree on the ideal outcome of this proposal.

My last update is one step further towards the idealized "from scratch" proposal, this makes it more explicitly clear that changing the effective type of a variable definition is only relevant when supporting legacy clients and suggests that new services should not use this behavior. I like that this balances a clear description of how this rule should work for existing services along with a stricter and therefore safer future path for new services.

I also just updated https://github.com/graphql/graphql-js/pull/1274 to include this and make this a breaking change by default while providing an API for using this legacy behavior.

I'd love a last look by anyone else interested

leebyron:
Updating this a bit. I'm still trying to convince myself that the breaking change is worth doing. I realized that it also has a side effect of not allowing an optional variable to be provided to a optional non-null argument.

```graphql
query Example($var: Int!) {
  field(arg: $var)
}

type Query {
  field(arg: Int! = 0): Int
}
```

While the above query is valid, it is invalid to not provide a runtime value for `$var`, even though arg supplies a default value. Perhaps this is just an acceptable cost.

An alternative, (which I'm also not sure about) is allowing nullable (optional) variables to flow into non-null optional positions, while still throwing at runtime for explicitly provided null values (already included in this proposal).

For example, this is currently invalid but would become valid in the alternative proposal:

```graphql
query Example($var: Int) { # nullable
  field(arg: $var) # non-null but optional
}

type Query {
  field(arg: Int! = 0): Int
}
```

In that case, omitting `$var` at runtime could just use `arg`'s default value.

Thoughts on this?

leebyron:
After discussing this in depth with @dschafer at a whiteboard - we've talked ourselves out of any breaking change - I think that having a fork in the expected behavior of validation rules is just too easy to get wrong, liable to cause issues during deployment, and could create a bifurcation of tooling environments. We came up with mitigation strategies for Facebook's codebases and APIs, however also agreed that public APIs like Github would be much much more challenging to solve for.

Even though the behavior is not an ideal "from scratch" - I think that it's worth avoiding the breaking change. I still think the spec should call out this potential issue directly in a non-normative note for clarity.

leebyron:
I've just updated the description to mirror the current change. I believe this is ready to merge and will first wait for a bit for others to review and comment.

IvanGoncharov:
> However, I don't think we should be terribly concerned with a bifurcation of tooling.

I agree with @mike-marcacci I think most of the tooling doesn't need to be backward compatible. There is only one scenario we shouldn't break: existing queries compiled inside clients that were deployed in production should be still supported by GraphQL server.
 
That means that all client libraries, ESlint plugin, client code generation, GraphiQL and other libraries/tools that are used only in development or during build process can be easily updated.

The only part of the ecosystem that should be backward compatible are GraphQL servers and maybe some proxy servers (e.g. Apollo Engine).

That's why I think both specification and `graphql-js` should use "from-scratch" behavior by default. Compatibility section of the specification can provide an alternative version of `IsVariableUsageAllowed` function and `graphql-js` can have `VariablesInAllowedPositionCompatible` validation rule.

> Because we're already altering the spec for query validation to expand the definition of validity, we're going to see a mismatch in the wild: some servers won't update their software, while new clients will expect looser queries to work. 

This is not a problem since if a new client stops working with old servers it would be caught during development. So I think we should distinguish between changes that will break already deployed clients and changes that will result in temporary inconvenience during development.
However, I fully agree that breaking changes are unavoidable in the future and we should prepare for them:

> This is probably unavoidable, and makes me think that there should be a "graphql version" or programmatically readable "feature set" (as discussed in the WG meeting) supplied in the result of an introspection query.

I think any mechanism involving introspection query is not a solution since the majority of GraphQL clients doesn't do any handshake; they just send queries to the server. So the only solution I see is for sever to detect what was the expected behavior during client development and fallback to it.
So I think having some kind of "graphql version" delivered together with GraphQL query is the right call.

> After discussing this in depth with @dschafer at a whiteboard - we've talked ourselves out of any breaking change - I think that having a fork in the expected behavior of validation rules is just too easy to get wrong, liable to cause issues during deployment, and could create a bifurcation of tooling environments. We came up with mitigation strategies for Facebook's codebases and APIs, however also agreed that public APIs like Github would be much much more challenging to solve for.

@leebyron I want to propose a solution. Currently, all GraphQL clients send GraphQL request document inside `query` field. But `ExecuteRequest` names the same parameter `document`:
```
ExecuteRequest(schema, document, operationName, variableValues, initialValue)
```
So it makes sense to rename this field to `document` inside GraphQL request.
In addition to that, we can make `version` a required field for all requests that use `document`. And both specification and reference implementation should provide the algorithm for converting request documents to a new version.

So a server can implement the following algorithm:
* Request contains `document`:
  * if the request also contains `query`, return a fatal error
  * if the request doesn't contain `version`, return a fatal error
  * use `version` from the request
* Request contains `query` then assume `version` is equal to legacy one(e.g. `0`)
* if `version` is newer than the last version supported by server return `fatal error`.

Pros:
* Don't break any deployed client.
* Add versioning to queries that allow bugfixes similar to this one.
* Allows clear error messages
* Rename `query` to `document` that makes sense on its own and solve a lot of terminology problems.

Cons:
* One additional field in every request.
* New clients stop working with old servers.

I also think we should use an integer for `version` field and increment it only when breaking change is made.

spec/Section 6 -- Execution.md
```diff
@@ -543,30 +551,36 @@ CoerceArgumentValues(objectType, field, variableValues):
...
+    * Otherwise if {argumentType} is a Non-Nullable type, and either {hasValue}
+      is not {true} or {value} is {null}, throw a field error.
+    * Otherwise if {hasValue} is true:
+      * If {value} is {null}:
+        * Add an entry to {coercedValues} named {argumentName} with the
+          value {null}.
+      * Otherwise, if {argumentValue} is a {Variable}:
+        * Add an entry to {coercedValues} named {argumentName} with the
+          value {value}.
+      * Otherwise:
```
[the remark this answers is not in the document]
leebyron: It is intentional. Variables are coerced during `CoerceVariableValues` before execution of a query at which point and resulting failures are also reported before execution, so a second coercion is not necessary.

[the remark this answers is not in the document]
mike-marcacci: Ah I see – I missed that note below. 👍

leebyron:
Thanks for the reviews, @mike-marcacci and @IvanGoncharov.

@IvanGoncharov I think there are too many critical issues with the versioning approach you're suggesting. Beyond being fundamentally inconsistent with one of GraphQL's core values to not need versioning, the introduction of versions causes even more breaking changes that we're trying to avoid.

>  Currently, all GraphQL clients send GraphQL request document inside query field. But ExecuteRequest names the same parameter document

This seems like it is an issue not for the spec but for HTTP services, and is just a naming convention so probably not a viable mechanism for specification algorithms. The spec document should be pretty consistent in referring to "document" and "operation" terms already.

> I also think we should use an integer for version field and increment it only when breaking change is made.

I would much prefer to not make breaking changes at all - I still believe our goal for evolving the GraphQL specification should be to arrive at a stable base and project confidence and stability. An incrementing integer version implies that there will be a continuous stream of breaking changes. Also, having worked on API surfaces (including public facing) that have dealt with breaking changes in the past, incrementing integer versions do not adequately capture scenarios where a client has adopted support for one breaking change while it has not yet adopted support for another - the fixed ordering of the version numbers forces an adoption curve that does not align with client expectation or capability. I'd like to strongly avoid making the same mistake with GraphQL.

@mike-marcacci I know you had similar ideas:

> This is probably unavoidable, and makes me think that there should be a "graphql version" or programmatically readable "feature set" (as discussed in the WG meeting) supplied in the result of an introspection query.

I think these are interesting ideas, but I'd love to avoid expanding the surface area of this proposal since I believe it is in a final reviewable state. I believe that finding consensus on programmable feature set support in a client-server communication is going to be quite challenging and require a lot of exploration. Just the two proposals you two offered are very different.

My hope is that we can arrive at consensus around this proposal in its current form. I'm not comfortable proposing a change which breaks the validation of existing queries.

mike-marcacci:
> I think these are interesting ideas, but I'd love to avoid expanding the surface area of this proposal since I believe it is in a final reviewable state.

Agreed; apologies for nudging the discussion off-topic. To reiterate what I intended to be the core point of my review, the current proposal successfully avoids breaking things that are already deployed while adding very little mess. This ended up a much cleaner solution than initially seemed possible 👍

leebyron:
Thanks for the reviews and offline conversations to everyone involves. Super pleased with how this ended up.

PR #422, opened 2018-03-08
Fix formatting for the Note section

IvanGoncharov:
image
http://facebook.github.io/graphql/draft/#example-a4647


leebyron:
Thanks for spotting this

PR #424, opened 2018-03-09
 Require Interface to implement at least 1 Object

mattkrick:
Matches graphql-js PR: https://github.com/graphql/graphql-js/pull/1280
For background, see https://github.com/graphql/graphql-js/issues/1277


leebyron:
Thanks for accounting for this schema validation addition!

mjmahone:
I think we should change the spec and revert this changes: there is no reason a schema should be required to have objects implementing each interface. A very common pattern is to create an interface first, start programming against that, then weeks later actually implement that interface. This is also a really nasty breaking change for large, existing schemas.

PR #428, opened 2018-04-18
RFC: SchemaExtension

leebyron:
This adds new grammar to the GraphQL SDL:

```graphql
extend schema {
  mutation: MutationType
}
```

This feels like the missing piece of the type system extension framework. In addition to extending types in a type system, the top level schema should be extendable as well.

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -187,10 +187,18 @@ Directive[Const] : @ Name Arguments[?Const]?
 TypeSystemDefinition :
   - SchemaDefinition
   - TypeDefinition
-  - TypeExtension
   - DirectiveDefinition
+  - TypeSystemExtension
+
+TypeSystemExtension :
+  - SchemaExtension
+  - TypeExtension
 
 SchemaDefinition : schema Directives[Const]? { OperationTypeDefinition+ }
+
+SchemaExtension :
+  - extend schema Directives[Const]? { OperationTypeDefinition+ }
+  - extend schema Directives[Const]
 
 OperationTypeDefinition : OperationType : NamedType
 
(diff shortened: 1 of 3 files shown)


OlegIlyenko:
:+1: I also got the same feeling after working with other extensions for a while.

On a side note (but I think it is a bit related), I wanted to ask about another limitation that was added at some point:

[[RFC] Add Validation rule for unique directives per location.](https://github.com/facebook/graphql/pull/229)

From the moment this validation was introduced, I felt that it was quite limiting. Considering that now there is a good support for splitting the schema across multple files, I think that it might be a good idea to reassess this validation. 

Do you think it make sense to reconsider it? (I personally would vote in favor of removing it or, at least, limiting the validation to a subset of directives)

Just to give a small example where it might be helpful. Consiudering the schema delegation case where multiple GraphQL schemas are merged into one. I can define the config like this:

https://github.com/OlegIlyenko/graphql-gateway/blob/master/testSchema.graphql#L6-L16

With the new extension I can potentially split the schema in 2 files:

```graphql
# file 1
schema
  @includeGraphQL(
    name: "starWars", 
    url: "http://try.sangria-graphql.org/graphql") {
  query: Query
}

# file 2
schema
  @includeGraphQL(
    name: "universe", 
    url: "https://www.universe.com/graphql/beta")
```

But even if directives are not split across different extensions, I would still find helpful to be able to use same directive multiple times (with different arguments) in the same location.

leebyron:
That's certainly something we could reconsider. I remember when applying that rule we did so intentionally to carve out the space for future exploration. Would you mind creating a new issue about that topic where we can discuss further?

OlegIlyenko:
Sure thing, I'll create a new issue :+1:

OlegIlyenko:
I created a separate issue for this:

[Limiting the scope of "Directives Are Unique Per Location" validation](https://github.com/facebook/graphql/issues/429)

leebyron:
Leaving this open for review for a shorter period of time - I hope this is an easy win :)

leebyron:
Thanks for the online and offline feedback everyone!

Issue #189, opened 2016-06-24
Spec and implementation inconsistent about circular references in input types 


kaqqao:
The Type System spec, under [Input objects](https://github.com/facebook/graphql/blob/master/spec/Section%203%20--%20Type%20System.md#input-objects), states (emphasis mine):

> The `Object` type defined above is inappropriate for re-use here, because `Object`s can contain fields that express **circular references or references to interfaces and unions, neither of which is appropriate for use as an input argument**. For this reason, input objects have a separate type in the system.

The reference implementation indeed disallows interfaces and unions in inputs, implying the above sentence should be understood as a rule forbidding them and not just a recommendation against using them.
On the other hand, the implementation allows circular references in `GraphQLInputObjectType`s.
E.g. the following is a legal declaration (`ArticleInputType` referring to `ArticleInputType`):

```
const ArticleInputType = new GraphQLInputObjectType({
  name: 'ArticleInput',
  fields: () => ({
    id:          { type: GraphQLInt },
    title:       { type: new GraphQLNonNull(GraphQLString) },
    body:        { type: GraphQLString },
    relatedArticles: { type: new GraphQLList(ArticleInputType) }
  })
});
```

This implies the sentence from the spec should be understood as a recommendation.

I suggest the wording in the corresponding spec section is changed to be consistent with the reference implementation or vice versa.
The current state leads to inconsistencies between the reference and other implementations, depending on their interpretation of the spec.

leebyron:
Great feedback, thanks! I'll work on clarifying this.

leebyron:
I think that the existing phrasing is being unduly strict. Circular types should be allowed, but circular non-null types probably should not.

For example, this input type seems like it should be valid:

```graphql
input Something {
  name: String
  somethingElse: Something
}
```

But this one is not valid since you cannot possibly provide a legal value:

```graphql
input Something {
  name: String
  somethingElse: Something!
}
```

dschafer:
> Circular types should probably be allowed, but circular non-null types should not.

Yeah, agreed. As it stands, the explanatory text is right (objects can have circular references or references to interfaces and unions, neither of which is appropriate for use as an input argument), but then we don't correctly disallow that in input objects. I would probably update the explanatory text to say (emphasis mine):

> objects can have **non-null** circular references or references to interfaces and unions, neither of which is appropriate for use as an input argument

and then add a 4th item to "Input Object type validation" that clarifies that input objects cannot contain circular references that are not broken up by a nullable field.

----

[...]

PR #445, opened 2018-05-16
Disallow non-breakable chains of circular references in Input Objects

spawnia:
This was already discussed in Issue #189 

Input Objects are allowed to reference other Input Objects. A circular reference occurs
when an Input Object references itself either directly or through subordinated Input Objects.

Circular references are generally allowed, however they may not be defined as an
unbroken chain of Non-Null fields. Such Input Objects are invalid, because there
is no way to provide a legal value for them.

The following examples are allowed:

```graphql example
input Example {
  self: Example
  value: String
}
```

This is fine because a value for `self` may simply be omitted from the arguments.

```graphql example
input Example {
  self: [Example!]!
  value: String
}
```

This also works as `self` can just contain an empty list.

The following examples are invalid:

```graphql counter-example
input Example {
  value: String
  self: Example!
}
```

```graphql counter-example
input First {
  second: Second!
  value: String
}

input Second {
  first: First!
  value: String
}
```

The following example shows why no possible value can be provided:

```graphql
{
  someField(input: {
    value: "val"
    # self is required
    self: {
      value: "nextval"
      # self is still required
      self: {
        # We would have to recurse down infinitely
        ...
      }
    }
  })
}
```

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -445,6 +445,8 @@ type Mutation {
 
 ## 💡 2. Explicit configurable Discriminator field
 
+**Champion:** @binaryseed
+
 A configurable discriminator field enables schema authors to model type discrimination into their schema more naturally.
 
 A schema author may choose to add their chosen type discriminator field to output types as well to completely mirror the structure in a way that enables sending data back and forth through input & output with no transformations.
@@ -755,6 +757,8 @@ input DogInput {
   * ❔ Not evaluated
 
 ## 💡 5. One Of (Tagged Union)
+
+**Champion:** @benjie
 
 This solution was presented in:
 * https://github.com/graphql/graphql-spec/pull/395#issuecomment-361373097


IvanGoncharov:
> This Input definition does not really make much sense, as the only use might be somehow conveying a certain nesting depth?

It's hard to write something like this by accident so I don't think it's worth to complicate spec and all implementations to prevent such constructs.
Plus technically this construct is valid it just happens to be useless in 99.999% of cases.
So I think it may be worth to add this check in 3rd-party GraphQL linters but not in the spec itself.

For me it's simular to how `a = a` is valid in JS, Flow and TS but you can configure ESLint to reject it:
https://eslint.org/docs/rules/no-self-assign

spawnia:
Good point. I also just thought of a use case where something like this might occur. When evolving a schema, there may be a period of time inbetween changes where one field is removed but later added.

Having such a rule in place would require major structural changes, so while it does not make sense often - it definitely should be allowed. Thank you for helping me to figure this out.


What is your opinion on the PR? Anything you would change?

spec/Section 3 -- Type System.md
```diff
@@ -1360,11 +1360,57 @@ Literal Value            | Variables               | Coerced Value
 
 **Type Validation**
 
-1. An Input Object type must define one or more input fields.
-2. The fields of an Input Object type must have unique names within that
-   Input Object type; no two fields may share the same name.
-3. The return types of each defined field must be an Input type.
+* An Input Object type must define one or more input fields.
```
IvanGoncharov: It should be consistent with other sections `Type Validation` that have numbered lists.

IvanGoncharov:
> What is your opinion on the PR? Anything you would change?

I think you need to split your text in two parts explaining "self-referencing" and an additional step in `Type Validation` section.
For example here is how it's done for field types:
* Explaination: https://github.com/facebook/graphql/blame/master/spec/Section%203%20--%20Type%20System.md#L1274-L1291
* Validation step: https://github.com/facebook/graphql/blame/master/spec/Section%203%20--%20Type%20System.md#L1366

Given the number of examples maybe it worth to have separate 'self-referencing' section.

leebyron:
Just to make sure I understand correctly, the example you posted should be possible:

```
input Hello {
  world: Hello
}
```

Since by default values can be null. But this would become invalid:

```
input Hello {
  world: Hello!
}
```

Since there cannot exist a value which satisfies it?

I think this is an interesting and potentially useful validation rule, but I sort of agree with Ivan that part of what we should weigh here is if this solves enough of a salient pain point for developers that it is worth the additional complexity in the spec.

Perhaps a good way to understand this question is to gauge how difficult actually implementing this validation rule in graphql-js would be. Is that something you're willing to try out, @spawnia ?

IvanGoncharov:
Merged in `graphql-js`: https://github.com/graphql/graphql-js/pull/1359 and planned to be released in `14.4.0`.

spawnia:
PR for the PHP implementation is on the way https://github.com/webonyx/graphql-php/pull/492.

spawnia:
Both major JS and PHP implementations have implemented this. @leebyron is there anything left for me to do?

IvanGoncharov:
@spawnia I think the last required step would be to discuss it on the WG.
We can have it tomorrow on WG: https://github.com/graphql/graphql-wg/blob/master/agendas/2019-10-10.md#agenda
Would you be able to join?

jbaxleyiii:
@IvanGoncharov @leebyron from what I can tell there is nothing technically blocking circular input types from being possible? Circular references are easily acomplished in many languages and may be the best representation of an actual input. Even serilization and transport is possible using different methods or transports (i.e. https://www.npmjs.com/package/arson)

spawnia:
@jbaxleyiii i can understand the logic behind your use case, it makes sense.

As of now, there is no way to provide a legal value for the input `User` you defined, using just a literal GraphQL query. The GraphQL query language lacks support for reference types.

It is certainly possible to represent such a data structure in most programming languages and possible some transport formats. That would *force* you to use variables whenever using that query. That might become problematic when, for example, you want to write a unit test or test the query through GraphiQL.

We could discuss adding reference types to GraphQL in a seperate RFC. This RFC merely clarifies what is possible with GraphQL as it is now, without actually changing the language.

jbaxleyiii:
> As of now, there is no way to provide a legal value for the input User you defined, using just a literal GraphQL query. 

This is also true for
```
input User {
  id: ID
  friend: User
}
```
GraphQL can't model recursive types (required or possible) in its current form. Due to that, I think this change is half in the door of making recursion forbidden and half in the door of making it possible but actually solving neither of the two.

jbaxleyiii:
Right, but you can't model all possible values there?

IvanGoncharov:
@spawnia Based on discussion during last WG the last point we need to address is the complexity of implementation: https://github.com/graphql/graphql-wg/blob/master/notes/2019-10-10.md#disallow-non-breakable-chains-of-circular-references-in-input-objects-benedikt-5m
Would you be able to make a presentation of JS or PHP implementation on WG tomorrow?

spawnia:
Hey @IvanGoncharov thanks for reminding me. I plan to participate, yes.

Here is a quick objective analysis in terms of metrics:

### LOC

Raw lines of code, including comments and whitespace

|-|PHP|JS|
|---|---|---|
|Implementation|104|50|
|Tests|138|114|

### Cyclomatic Complexity

In the JS implementation, the validation function has a cyclomatic complexity of **6** and is called in exactly one place.

spawnia:
I want to add a few thoughts i developed during the last month.

### Implementors are not forced to add this validation

Just because it is in the spec does not mean that a validation that enforces this rule must be implemented. Implementors can choose to ignore it, if they don't want the extra complexity or spend time on implementing the check.

### Assumptions

There are a few assumptions underlying this rule, which naturally lead to it being a requirement.

#### Schemas must be queryable

GraphQL schemas are meant to be used for querying and must be functional for such. GraphQL is not meant solely to express a type system and offer introspection capabilities.

The sole purpose of a non-breakable self-referencing input object would be to be introspected upon. It could not ever be used in an actual query.

#### GraphQL literals are first class

All GraphQL operations must be able to be expressed by using literal GraphQL query language. Features that work only when variables are involved are not valid.

While there may be ways to express non-breakable self-referencing input object in most programming languages and some transport formats, it is not possible to do so using only the GraphQL query language itself and the literals it provides.

spawnia:
The pull bot nuked my `master` branch, this continues in https://github.com/graphql/graphql-spec/pull/701

PR #451, opened 2018-05-29
Fix `SameResponseShape` definition

MSleepyPanda:
For motivation, see discussion in https://github.com/facebook/graphql/issues/450

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -430,7 +430,7 @@ SameResponseShape(fieldA, fieldB) :
     * Let {typeB} be the item type of {typeB}
     * Repeat from step 3.
   * If {typeA} or {typeB} is Scalar or Enum.
-    * if {typeA} and {typeB} are the same type return true, otherwise return
+    * If {typeA} and {typeB} are the same type return true, otherwise return
       false.
   * If {typeA} and {typeB} are both composite types:
     * Let {mergedSet} be the result of adding the selection set of {fieldA} and
@@ -438,7 +438,7 @@ SameResponseShape(fieldA, fieldB) :
     * Let {fieldsForName} be the set of selections with a given response name in
       {mergedSet} including visiting fragments and inline fragments.
     * Given each pair of members {subfieldA} and {subfieldB} in {fieldsForName}:
-      * if SameResponseShape(subfieldA, subfieldB)} is false, return false,
+      * If SameResponseShape(subfieldA, subfieldB)} is false, return false,
         otherwise continue.
     * Return true.
   * Return false.
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -430,14 +430,18 @@ SameResponseShape(fieldA, fieldB) :
     * Let {typeB} be the item type of {typeB}
     * Repeat from step 3.
   * If {typeA} or {typeB} is Scalar or Enum.
-    * {typeA} and {typeB} must be the same type.
-  * Assert: {typeA} and {typeB} are both composite types.
-  * Let {mergedSet} be the result of adding the selection set of {fieldA} and
-    the selection set of {fieldB}.
-  * Let {fieldsForName} be the set of selections with a given response name in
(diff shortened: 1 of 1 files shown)


spec/Section 5 -- Validation.md
```diff
@@ -430,14 +430,18 @@ SameResponseShape(fieldA, fieldB) :
...
+    * If {typeA} and {typeB} are the same type return true, otherwise return
+      false.
+  * If {typeA} and {typeB} are both composite types:
+    * Let {mergedSet} be the result of adding the selection set of {fieldA} and
+      the selection set of {fieldB}.
+    * Let {fieldsForName} be the set of selections with a given response name in
+      {mergedSet} including visiting fragments and inline fragments.
+    * Given each pair of members {subfieldA} and {subfieldB} in {fieldsForName}:
+      * If SameResponseShape(subfieldA, subfieldB)} is false, return false,
+        otherwise continue.
```
IvanGoncharov: you don't need to write `otherwise continue`.
Example:
http://facebook.github.io/graphql/draft/#AreTypesCompatible()
> If variableType is NOT a non‐null type, return false

spec/Section 5 -- Validation.md
```diff
@@ -430,14 +430,18 @@ SameResponseShape(fieldA, fieldB) :
...
-  * Assert: {typeA} and {typeB} are both composite types.
-  * Let {mergedSet} be the result of adding the selection set of {fieldA} and
-    the selection set of {fieldB}.
-  * Let {fieldsForName} be the set of selections with a given response name in
-    {mergedSet} including visiting fragments and inline fragments.
-  * Given each pair of members {subfieldA} and {subfieldB} in {fieldsForName}:
-    * {SameResponseShape(subfieldA, subfieldB)} must be true.
+    * If {typeA} and {typeB} are the same type return true, otherwise return
+      false.
+  * If {typeA} and {typeB} are both composite types:
```
IvanGoncharov: I suggest changing it to:
>  If {typeA} or {typeB} is not a composite type, return false

It allows moving subsequent steps on the top level and also removing final:
> Return false

IvanGoncharov:
To achieve full consistency I would also change:
> {typeA} and {typeB} must both be Non-Null.

to

> If {typeA} or {typeB}  is not a Non-Null, return false.

Same here:
> {typeA} and {typeB} must both be List.

MSleepyPanda:
Addressed review!

spec/Section 5 -- Validation.md
```diff
@@ -421,23 +421,25 @@ SameResponseShape(fieldA, fieldB) :
...
+    * If {typeA} or {typeB} is not List, return false.
     * Let {typeA} be the item type of {typeA}
     * Let {typeB} be the item type of {typeB}
     * Repeat from step 3.
   * If {typeA} or {typeB} is Scalar or Enum.
-    * {typeA} and {typeB} must be the same type.
-  * Assert: {typeA} and {typeB} are both composite types.
+    * If {typeA} and {typeB} are the same type return true, otherwise return
+      false.
+  * If {typeA} and {typeB} is not a composite type, return false.
```
IvanGoncharov: should be `or`

spec/Section 5 -- Validation.md
```diff
@@ -421,23 +421,25 @@ SameResponseShape(fieldA, fieldB) :
...
   * Let {mergedSet} be the result of adding the selection set of {fieldA} and
     the selection set of {fieldB}.
   * Let {fieldsForName} be the set of selections with a given response name in
     {mergedSet} including visiting fragments and inline fragments.
   * Given each pair of members {subfieldA} and {subfieldB} in {fieldsForName}:
-    * {SameResponseShape(subfieldA, subfieldB)} must be true.
+    * If SameResponseShape(subfieldA, subfieldB)} is false, return false.
```
IvanGoncharov: You missed the curly brace here

MSleepyPanda: Whoops, fixed!

leebyron:
Great catch and fix! Thanks for your help

PR #452, opened 2018-05-29
Fix: Leaf Field Selections for enums must also be empty

MSleepyPanda:


IvanGoncharov:
@MSleepyPanda Thanks for quick PR 👍 
Would be great if you also fix `Explanatory Text` in the same section, to keep it in sync.

MSleepyPanda:
While i'm at it, re Example 114:
`fragment scalarSelectionsNotAllowedOnBoolean on Dog`: Why Boolean? barkVolume is an Int. Also the intent is to convey that no further selections are allowed on primitives.

I'd propose changing the name to `scalarSelectionsNotAllowed`. I'd expand the example by `enumSelectionsNotAllowed`.

IvanGoncharov:
> While i'm at it, re Example 114:
fragment scalarSelectionsNotAllowedOnBoolean on Dog: Why Boolean? barkVolume is an Int. Also the intent is to convey that no further selections are allowed on primitives.

@MSleepyPanda Good catch 👍 

> I'd propose changing the name to scalarSelectionsNotAllowed. I'd expand the example by enumSelectionsNotAllowed.

Makes total sense to me, but I suggest keeping `onInt` part to prevent possible confusion.

spec/Section 5 -- Validation.md
```diff
@@ -568,7 +568,7 @@ fragment conflictingDifferingResponses on Pet {
 
 **Explanatory Text**
 
-Field selections on scalars are never allowed: scalars
+Field selections on scalars or enums are never allowed: scalars and enums
```
IvanGoncharov: nit: maybe instead of `: scalars and enums` write `because they are the leaf nodes of any GraphQL query.`

IvanGoncharov:
@MSleepyPanda This and #451 both look good 👍 
Next step is the review by  @leebyron

If you find any other places where clarification needed to feel free to open PR or create an issue.

leebyron:
Great edit, thank you!

PR #454, opened 2018-06-04
Add validation steps for schema extensions

IvanGoncharov:
Reported by @OlegIlyenko in https://github.com/facebook/graphql/issues/429#issuecomment-392990065
I think it should be added in upcomming release for consistency with other definition extensions.


leebyron:
Agreed

PR #456, opened 2018-06-04
Use "type system definition language" consistently

IvanGoncharov:
All other places reference it as "type system definition language".


IvanGoncharov:
@leebyron Just to clarify "schema definition language" was renamed to "type system definition language", right?
So what abbreviation should be used instead of SDL?
TSDL or TSD?

leebyron:
schema definition language is synonymous and fine to continue to use (as is SDL) when discussing - but type system definition language was more specifically correct relative to the rest of the document's use of the terminology

I don't think we need an acronym / abbreviation since we don't use abbreviations within the spec - I think talking about it as SDL in discussion is still clear

PR #460, opened 2018-06-08
RFC: Require Interfaces to have 1+ Objects implementing

mjmahone:
Re-applies #424, but as an RFC. This change is a major spec breaking change, and if it is desire-able, we should provide an upgrade path for existing schemas.


leebyron:
Thanks for putting this up - I'll add it as an agenda item for next week's WG session since it's notable

leebyron:
We forgot to address this at the WG meeting, but I think the concerns encountered warrant rejecting this proposal. Thanks for the careful revert and re-RFC, @mjmahone!

PR #470, opened 2018-06-23
[RFC] "Directive order is significant" section

OlegIlyenko:
I would like to add a note on the directive ordering to make it more explicit. It is similar to the notes:

* [Argument are unordered. ](http://facebook.github.io/graphql/June2018/#example-7eba7)
* [Input object fields are unordered.](http://facebook.github.io/graphql/June2018/#example-09646)

But it instead states that directive order is significant and may have specific semantic interpretation.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -1094,3 +1094,21 @@ and operations.
 
 As future versions of GraphQL adopt new configurable execution capabilities,
 they may be exposed via directives.
+
+**Directive order is significant**
+
+Directives may be provided in a specific syntactic order which may have semantic interpretation.
+
+These two type definitions may have different semantic meaning:
+
+```graphql example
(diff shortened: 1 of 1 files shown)


OlegIlyenko:
Forgot to mention that this RFC is the result of recent discussions in https://github.com/facebook/graphql/issues/429 and at the WG meetings.

leebyron:
Are any changes necessary to GraphQL.js or Sangria to support this?

OlegIlyenko:
@leebyron I believe no changes are necessary. RFC permits implementations to have a special treatment for directive order, but does not require it. Also, for all built-in directives, specific order is not important.

spawnia:
This is a very reasonable clarification to the spec. In a project of ours, we actually rely on the order of directive definitions being preserved - which https://github.com/webonyx/graphql-php thankfully does already.

Would be great to have this in to make this an explicitely defined rule that we can depend on.

PR #471, opened 2018-06-24
[RFC] Limit uniqueness to `@skip`, `@include` and `@deprecated` directives

### LINKED ISSUE #429: Limiting the scope of "Directives Are Unique Per Location" validation
According to the spec, ["Directives Are Unique Per Location"](http://facebook.github.io/graphql/October2016/#sec-Directives-Are-Unique-Per-Location) validation is defined like this:

> Directives are used to describe some metadata or behavioral change on the definition they apply to. When more than one directive of the same name is used, the expected metadata or behavior becomes ambiguous, therefore only one of each directive is allowed per location.

In the light of recent GraphQL spec changes (extensions on all types and the schema itself) and other developments in the community, I would like to suggest to reconsider the inclusions of this validation rule or limiting it's scope to a subset of directives.

I would like to show 2 cases where it might be beneficial to allow non-unique directives.

### Example 1: adding more information

In this scenario, directives might add more information. The order of directives is insignificant.

Considering the schema delegation case where multiple GraphQL schemas are merged into one. I can define the config like this (with the current limitation):

```
schema
  @includeGraphQL(schemas: [{
    name: "schema1"
    url: "http://some.server/graphql"
  }, {
    name: "schema2"
    url: "http://another.server/graphql"
  }]) {

  query: Query
}
```

If the limitation is lifted, I can potentially split the schema in 2 files like this:

```graphql
# file 1
schema
  @includeGraphQL(
    name: "schema1", 
    url: "http://some.server/graphql") {
  query: Query
}

# file 2
extend schema
  @includeGraphQL(
    name: "schema2", 
    url: "http://another.server/graphql")
```

This provides more modularity (and, I would argue, simplicity), even in absence of an extension:

```graphql
schema
  @includeGraphQL(
    name: "schema1", 
    url: "http://some.server/graphql") 
  @includeGraphQL(
    name: "schema2", 
    url: "http://another.server/graphql") {
  query: Query
}
```

### Example 2: chaining directives/data transformation

In this scenario, I would like to transform the data in a chain. A small subset of directives might represent reusable logic for a data retrieval and transformation. In this case the ordering the directives is significant.

For example, if I have a set of directives to retrieve data from an external HTTP service, I can define a field like this:

```graphql
type Query {
  users: [User]
    @httpGet(url: "...")
    @jsonPath(path: "$.results")
    @enrichData
    @jsonPath(path: "$.profile")
}
```

### Possible solutions

We can tackle this in a number of ways:

1. Remove the validation altogether. In this case, we need to reconsider what to do with ambiguous `@skip`, `@include` and potentially other custom directives
2. Limit the validation to only `@skip` and `@include` directives
3. Limit the validation to only these directives that are explicitly marked as "unique". This implies that we need to introduce a new option in the directive definition.
4. Leave "Directives Are Unique Per Location" validation as-is

I would really appreciate your opinions and comments on this suggestion.

OlegIlyenko:
This proposal directly relates to a discussion in https://github.com/facebook/graphql/issues/429. As was discussed at the latest WG meeting, I'm creating several alternative proposals. This one implements [proposed solution](https://github.com/facebook/graphql/issues/429#issuecomment-392946579) `3. Limit the validation to only @skip and @include directives`.

It limits the scope of "Directives Are Unique Per Location" to `@skip` and `@include` directives only.

This proposal is mutually exclusive with other alternative proposals:

* [RFC] Limit uniqueness to `@skip`, `@include` and `@deprecated` directives (#471, this one)
* [RFC] Limit directive uniqueness to explicitly marked directives (#472)

@leebyron @IvanGoncharov @jjergus I would appreciate your reviews.

Closes #429

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -165,7 +165,6 @@ adds additional operation types, or additional directives to an existing schema.
 Schema extensions have the potential to be invalid if incorrectly defined.
 
 1. The Schema must already be defined.
-2. Any directives provided must not already apply to the original Schema.
 
 
 ## Descriptions
@@ -544,7 +543,6 @@ GraphQL tool or service which adds directives to an existing scalar.
 Scalar type extensions have the potential to be invalid if incorrectly defined.
 
 1. The named type must already be defined and must be a Scalar type.
-2. Any directives provided must not already apply to the original Scalar type.
 
 
 ## Objects
@@ -934,10 +932,9 @@ Object type extensions have the potential to be invalid if incorrectly defined.
    may share the same name.
 3. Any fields of an Object type extension must not be already defined on the
    original Object type.
-4. Any directives provided must not already apply to the original Object type.
-5. Any interfaces provided must not be already implemented by the original
+4. Any interfaces provided must not be already implemented by the original
    Object type.
-6. The resulting extended object type must be a super-set of all interfaces it
+5. The resulting extended object type must be a super-set of all interfaces it
    implements.
 
 
@@ -1116,7 +1113,6 @@ Interface type extensions have the potential to be invalid if incorrectly define
 4. Any Object type which implemented the original Interface type must also be a
    super-set of the fields of the Interface type extension (which may be due to
    Object type extension).
-5. Any directives provided must not already apply to the original Interface type.
 
 
 ## Unions
@@ -1239,7 +1235,6 @@ Union type extensions have the potential to be invalid if incorrectly defined.
 3. All member types of a Union type extension must be unique.
 4. All member types of a Union type extension must not already be a member of
    the original Union type.
-5. Any directives provided must not already apply to the original Union type.
 
 ## Enums
 
@@ -1308,7 +1303,6 @@ Enum type extensions have the potential to be invalid if incorrectly defined.
 2. All values of an Enum type extension must be unique.
 3. All values of an Enum type extension must not already be a value of
    the original Enum.
-4. Any directives provided must not already apply to the original Enum type.
 
 
 ## Input Objects
@@ -1437,7 +1431,6 @@ Input object type extensions have the potential to be invalid if incorrectly def
 3. All fields of an Input Object type extension must have unique names.
 4. All fields of an Input Object type extension must not already be a field of
    the original Input Object.
-5. Any directives provided must not already apply to the original Input Object type.
 
 
 ## List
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@OlegIlyenko I will try to do detail review in the next couple of days. But from top of my head, it also should include `@deprecate`.

OlegIlyenko:
@IvanGoncharov thanks! Good point, I focused on `@skip` and `@include` and completely forgot about `@deprecated`. I updated the PR.

IvanGoncharov:
Partly alternative proposal based on https://github.com/facebook/graphql/issues/429#issuecomment-392972131: I think we can define the behavior of multiple `@skip`/`@include`.
We could change this:
http://facebook.github.io/graphql/June2018/#sec--include
> Neither @skip nor @include has precedence over the other. In the case that both the @skip and  @include directives are provided in on the same the field or fragment, it must be queried only if the  @skip condition is false and the @include condition is true

into this:
> Neither @skip nor @include has precedence over the other or other instance of the same directive. In the case that multiple @skip or @include directives are provided in on the same the field or fragment, it must be queried only if all @skip conditions are false and all @include conditions are true

This also require to change this algorithm:
http://facebook.github.io/graphql/June2018/#CollectFields()

Benefits: Currently you can model only OR for includes:
```graphql
{
   field @include(if: $foo)
   field @include(if: $bar)
}
```
These proposal allows to do the opossite operation:
```graphql
{
   field @include(if: $foo) @include(if: $bar)
}
```

OlegIlyenko:
I like this suggestion. This lifts the constraint and defines intuitive disambiguation rules.

I still struggle with the question of why such a strong line is drawn between:

1. `field @skip(if: false) @skip(if: true)` (disallowed by validation)
2. `field @skip(if: false) @include(if: false)` (allowed)

both variations are semantically identical (and raise the same ambiguity issues). In my opinion, if we keep that validation around, then it probably should be reason other than ambiguity between `@skip` and `@include`. Though it makes sense for `@deprecated` since there is no straightforward way to disambiguate 2 of these.

Maybe I should prepare the third PR that is a combination of #472 + non-unique `@skip` and `include` + the change in the text that you have suggested. What does everybody think about it?

leebyron:
To follow up from the WG meeting: I think we should extract the skip/include rules from the deprecated to ensure we're leading to the right solution for each. I'm excited about the direction of clearing up ambiguity when using skip and include together

OlegIlyenko:
@leebyron sure thing. i think I would prefer to first finish #470 and #472 to not mix thing too much. Later on I think this PR needs to be closed and I can open another one which is dedicated to skip/include interaction.

PR #480, opened 2018-07-03
Add example of negative integer coercion to ID

IvanGoncharov:
Changed from:
> or integer (such as `4`) input value

to
> or non-negative integer (such as `0` or `4`) input value

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -524,10 +524,11 @@ When coercion is not possible they must raise a field error.
 
 **Input Coercion**
 
-When expected as an input type, any string (such as `"4"`) or integer (such
-as `4`) input value should be coerced to ID as appropriate for the ID formats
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@OlegIlyenko Looks like negative IDs is a thing:
https://dba.stackexchange.com/questions/2895/what-are-negative-keys-used-for

In this case, I think the specification should have an example of such conversion to show that this is intentional behavior.
I changed PR to:
> or integer (such as `4` or `-4`) input value

leebyron:
A good next step might be an analysis of what GraphQL libraries already have this restriction in place, to better understand the cost of a breaking change. I believe GraphQL js has had this for a long time and I know many libraries start as a port of that library

IvanGoncharov:
@leebyron After @OlegIlyenko suggestion I changed this PR to just add example of negative ID value so that it more explicit in the spec.
So no breaking changes in this PR.

PR #509, opened 2018-08-29
[RFC] Allow singular variables in list locations

taion:
This is a follow-up to discussion in https://github.com/graphql/graphql-relay-js/issues/20#issuecomment-416809871.

The current coercion rules, for convenience, allow passing in singular literals in positions expecting lists. However, they do not allow passing in singular variables.

This proposal would allow passing in singular variables as well. The coercion rules already ensure reasonable behavior when this happens. Enabling this allows easier evolution of APIs, as it would enable converting singular filtering/ordering fields per the linked issues to plural ones without breaking clients.

cc @dschafer

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1857,7 +1857,10 @@ AreTypesCompatible(variableType, locationType):
     * Let {nullableVariableType} be the nullable type of {variableType}.
     * Return {AreTypesCompatible(nullableVariableType, locationType)}.
   * Otherwise, if {locationType} is a list type:
-    * If {variableType} is NOT a list type, return {false}.
+    * If {variableType} is NOT a list type:
+      * Let {itemLocationType} be the unwrapped item type of {locationType}.
+      * Let {nullableItemLocationType} be the unwrapped nullable type of {itemLocationType} if {itemLocationType} is non-null, or {itemLocationType} if {itemLocationType} is nullable.
+      * Return {AreTypesCompatible(variableType, nullableItemLocationType)}.
     * Let {itemLocationType} be the unwrapped item type of {locationType}.
     * Let {itemVariableType} be the unwrapped item type of {variableType}.
     * Return {AreTypesCompatible(itemVariableType, itemLocationType)}.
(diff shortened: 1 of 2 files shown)


dschafer:
CC @leebyron — I can't recall a reason not to do this offhand. Maybe a good discussion for the next WG?

taion:
@IvanGoncharov Interesting! So the behavior in graphql-js currently doesn't follow the spec? And I imagine the same applies for Ruby, too, since that's what GitHub are using. I'll amend, thanks.

taion:
Oh, never mind, I misunderstood what `CoerceVariableValues` was doing.

IvanGoncharov:
@taion You can't remove variable coercion it's extremely important section.
\+ coercing variables on every argument would affect perfomance.

What I meant in my comment is that with your proposal coercing variable value before executing operation is still requered but not sufficient on it's own. And that's why you need to add specific step to `CoerceVariableValues` that will wrap variable value in List if needed.

IvanGoncharov:
@taion For example by spec ID also supports Int as an input and it's coerced to String. So in my example, if I pass `{ id: 1 }` there are following steps:
* `query ($id: ID){` coercion  so it's become `{ id: "1" }` and pass to [ExecuteQuery](http://facebook.github.io/graphql/June2018/#ExecuteQuery())
* `getPeople(ids: $id)` coercion so it's become `{ id: ["1"]}` and pass to [ResolveFieldValue](http://facebook.github.io/graphql/June2018/#ResolveFieldValue())

So one coercion is per operation to match type in variable definition and second is per arg to match type in argument definition.

IvanGoncharov:
This dialog helped me to understand why such change is problematic.
Arguments from @taion can be applied to any other type of coercion beyond scalar to list. For example, integers are coerced to floats so you can write
```graphql
{
  divide(dividend: 10, divisor: 3)
}
```
Does this mean we should support:
```graphql
query ($value: Int) {
  divide(dividend: 10, divisor: $value)
}
```

> Enabling this allows easier evolution of APIs, as it would enable converting singular filtering/ordering fields per the linked issues to plural ones without breaking clients.

You can apply exact same argument saying that it would enable converting Int to Float without breaking clients. Same goes for other conversions including custom scalar.

So I would argue it pretty big change. More over it's change semantic of coercion. because right now coercion applied only on user-provided data when it's crossing GraphQL boundaries and become typed.
On the other hand, proposed change basically means that we process client value at operation boundary and it's already strong typed but we also allow to convert it into another type inside the query.

IMHO, this PR is not about extending coercion of user-provided data to typed GraphQL values but instead the proposal to add the first entry to completely new class of "conversion" rules beetween already typed values inside query.

Thinking about it, I'm completely ok with changing validation rules but I'm not sure that it's woth to change Execute algorithms to enable this use case.

taion:
This is bigger than I thought it would be.

The motivation here really is around concerns like https://github.com/graphql/graphql-relay-js/issues/20#issuecomment-135887330.

Especially when using static queries that use variables for these things, such as when using Relay Modern or when using Apollo on iOS, it's pretty difficult to make these sorts of fixes to input types.

There's some real inconvenience here, and it all seems so unnecessary, as it wouldn't be a problem if we were using literals.

Perhaps given the scope, it would be better to wait to discuss this at the next WG meeting.

taion:
As discussed at the WG, the next step here is to spike an implementation in graphql-js to allow some benchmarking, and to evaluate the incremental change in code complexity.

8860730 (approved):

23tux:
Is there any news on this?

PR #520, opened 2018-10-02
Fix ambiguity about optional variables in non-null args

### LINKED ISSUE #507: Incoherent description for default value in an argument
The caption for [example nr. 174](http://facebook.github.io/graphql/June2018/#example-0877c) reads:

> In the example above, an optional variable is allowed to be used in an non‐null argument which provides a default value.

but the code snippet doesn't show an example with a default value:

```graphql
query booleanArgQueryWithDefault($booleanArg: Boolean) {
  arguments {
    optionalNonNullBooleanArgField(optionalBooleanArg: $booleanArg)
  }
}
```

leebyron:
"Allowing optional variables when default values exist" wasn't very clear about what elements of the example were being referred to as optional/required or non-nullable. This minor rewrite attempts to resolve that ambiguity

Fixes #507

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1938,9 +1938,9 @@ A notable exception to typical variable type compatibility is allowing a
 variable definition with a nullable type to be provided to a non-null location
 as long as either that variable or that location provides a default value.
 
-In the example below, an optional variable is allowed to be used in an 
-non-null optional argument (`optionalBooleanArg`) because it provides a default 
-value (`false`).
+In the example below, an optional variable `$booleanArg` is allowed to be used
+in the non-null argument `optionalBooleanArg` because the field argument is
+optional since it provides a default value in the schema.
 
 ```graphql example
 query booleanArgQueryWithDefault($booleanArg: Boolean) {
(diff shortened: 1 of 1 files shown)


spec/Section 5 -- Validation.md
```diff
@@ -1938,6 +1938,10 @@ A notable exception to typical variable type compatibility is allowing a
 variable definition with a nullable type to be provided to a non-null location
 as long as either that variable or that location provides a default value.
 
+In the example below, an optional variable is allowed to be used in an 
+non-null optional argument (`optionalBooleanArg`) because it provides a default
```
mjmahone: `it` is ambiguous (optional variable or non-null optional argument?) and further muddied with the naming `optionalBooleanArg` (how is an optional argument non-nullable?)

Proposal:
```
In the example below, an optional variable is allowed to be used in an 
non-null optional argument (`booleanArgWithDefault`) because the
field argument provides a default value in the schema.
```

spec/Section 5 -- Validation.md
```diff
@@ -1946,7 +1950,11 @@ query booleanArgQueryWithDefault($booleanArg: Boolean) {
 }
 ```
 
-In the example above, an optional variable is allowed to be used in an non-null argument which provides a default value.
+In the example below, an optional variable (`$booleanArg`) provides a default 
+value (`true`) and can be used in a non-null requied argument. This behavior is
```
mjmahone: typo: required

Issue #197, opened 2016-07-19
Deprecation of input fields


dylanahsmith:
It is possible to deprecate fields and enum values, but not input values or arguments.  I think it should be possible to deprecate optional arguments and input fields.

A natural way of evolving the schema in this case would be to make an input field optional if it is required, deprecate it, then remove it once clients stop explicitly setting the value.  We can add support for that into the server to detect when the input field is still used, but we would like deprecation warnings on the client so we can easily remove usage of these deprecated input fields.

One possible workaround is to create a new field with different input, but that is a more disruptive way of evolving the schema.

Note that if __InputValue was just given isDeprecated and deprecationReason fields, then that would mean that an input object type could be deprecated, would that be undesirable?  Right now an __Type (object, interface or union) also can't be deprecated, so would we want that to be consistent with an input object type?

Also, should the spec disallow an argument to be deprecated if it is non-null (i.e. required)?  Otherwise, that effectively makes the field it is on deprecated.

cjoudrey:
@leebyron any thoughts on this?

leebyron:
Yeah I think this is probably an interesting idea, though we haven't put a lot of thought into it yet.

I'm less sure about making types themselves deprecated - rather the references to those types should be deprecated. Perhaps that means also being able to mark arguments as deprecated. If input obj fields can be transitioned from required to optional to deprecated, then arguments should be the same.

The next steps for an idea like this would be to draft an RFC, probably in the form of a PR against the spec along with a good description of how deprecation best practice should work and any potential caveats.

brandonblack:
> Perhaps that means also being able to mark arguments as deprecated. If input obj fields can be transitioned from required to optional to deprecated, then arguments should be the same.

I'd be very supportive and a huge fan of this change. 👍  We've already run into a few scenarios in our implementation where this would be valuable and I can foresee it becoming even more important in the future. 

I'd be happy to collaborate with anyone who wants to begin working on this.

sibelius:
few places that are useful

- [ ] args of connections/fields
- [ ] args of mutation

gjtorikian:
> The next steps for an idea like this would be to draft an RFC, probably in the form of a PR against the spec along with a good description of how deprecation best practice should work and any potential caveats.

[...]

PR #525, opened 2018-10-23
Allow deprecation of input values

smitt04:
Started the @deprecated proposal in reference to #197

Please let me know if there is anything i missed or you would like changed.  This was just a quick update to get the ball rolling again.

@IvanGoncharov Are you still interested in reviewing this?

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1901,7 +1901,7 @@ must *not* be queried if either the `@skip` condition is true *or* the
 ```graphql
 directive @deprecated(
   reason: String = "No longer supported"
-) on FIELD_DEFINITION | ENUM_VALUE
+) on FIELD_DEFINITION | ENUM_VALUE | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
 ```
 
 The `@deprecated` directive is used within the type system definition language
@@ -1912,11 +1912,16 @@ Deprecations include a reason for why it is deprecated, which is formatted using
 Markdown syntax (as specified by [CommonMark](https://commonmark.org/)).
 
 In this example type definition, `oldField` is deprecated in favor of
-using `newField`.
+using `newField` and `oldArg` is deprecated in favor of using `newArg`.
 
 ```graphql example
 type ExampleType {
   newField: String
   oldField: String @deprecated(reason: "Use `newField`.")
+
+  existingField(
+    newArg: String,
+    oldArg: String @deprecated(reason: "Use `newArg`.")
+  ): String
 }
 ```
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -143,7 +143,7 @@ type __Type {
   enumValues(includeDeprecated: Boolean = false): [__EnumValue!]
 
   # should be non-null for INPUT_OBJECT only, must be null for the others
-  inputFields: [__InputValue!]
+  inputFields(includeDeprecated: Boolean = false): [__InputValue!]
 
   # should be non-null for NON_NULL and LIST only, must be null for the others
   ofType: __Type
@@ -152,7 +152,7 @@ type __Type {
 type __Field {
   name: String!
   description: String
-  args: [__InputValue!]!
+  args(includeDeprecated: Boolean = false): [__InputValue!]!
   type: __Type!
   isDeprecated: Boolean!
   deprecationReason: String
@@ -163,6 +163,8 @@ type __InputValue {
   description: String
   type: __Type!
   defaultValue: String
+  isDeprecated: Boolean!
+  deprecationReason: String
 }
 
 type __EnumValue {
@@ -335,6 +337,8 @@ Fields
 * `name` must return a String.
 * `description` may return a String or {null}.
 * `inputFields`: a list of `InputValue`.
+  * Accepts the argument `includeDeprecated` which defaults to {false}. If
+    {true}, deprecated fields are also returned.
 * All other fields must return {null}.
 
 
@@ -374,6 +378,8 @@ Fields
 * `description` may return a String or {null}
 * `args` returns a List of `__InputValue` representing the arguments this
   field accepts.
+  * Accepts the argument `includeDeprecated` which defaults to {false}. If
+    {true}, deprecated arguments are also returned.
 * `type` must return a `__Type` that represents the type of value returned by
   this field.
 * `isDeprecated` returns {true} if this field should no longer be used,
@@ -395,6 +401,9 @@ Fields
 * `defaultValue` may return a String encoding (using the GraphQL language) of the
   default value used by this input value in the condition a value is not
   provided at runtime. If this input value has no default value, returns {null}.
+* `isDeprecated` returns {true} if this field or argument should no longer be used,
+  otherwise {false}.
+* `deprecationReason` optionally provides a reason why this input field or argument is deprecated.
 
 ### The __EnumValue Type
 
@@ -420,5 +429,7 @@ Fields
   locations this directive may be placed.
 * `args` returns a List of `__InputValue` representing the arguments this
   directive accepts.
+  * Accepts the argument `includeDeprecated` which defaults to {false}. If
+    {true}, deprecated arguments are also returned.
 * `isRepeatable` must return a Boolean that indicates if the directive may be
   used repeatedly at a single location.


spec/Section 3 -- Type System.md
```diff
@@ -1778,3 +1778,15 @@ type ExampleType {
   oldField: String @deprecated(reason: "Use `newField`.")
 }
 ```
+
```
IvanGoncharov: I think you also need to expand above text:
> The @deprecated directive is used within the type system definition language to indicate deprecated portions of a GraphQL service’s schema, such as deprecated fields on a type or deprecated enum values.

spec/Section 4 -- Introspection.md
```diff
@@ -332,6 +334,8 @@ Fields
 * `name` must return a String.
 * `description` may return a String or {null}.
 * `inputFields`: a list of `InputValue`.
+  * Accepts the argument `includeDeprecated` which defaults to {false}. If
+    {true}, deprecated enum values are also returned.
```
IvanGoncharov:
```suggestion
    {true}, deprecated fields are also returned.
```

spec/Section 4 -- Introspection.md
```diff
@@ -392,6 +396,9 @@ Fields
 * `defaultValue` may return a String encoding (using the GraphQL language) of the
   default value used by this input value in the condition a value is not
   provided at runtime. If this input value has no default value, returns {null}.
+* `isDeprecated` returns {true} if this field or argument should no longer be used,
+  otherwise {false}.
+* `deprecationReason` optionally provides a reason why this field or argument is deprecated.
```
IvanGoncharov:
```suggestion
* `deprecationReason` optionally provides a reason why this input field or argument is deprecated.
```

IvanGoncharov:
@smitt04 Looks good so far, I left a few review feedbacks.
Next step would be to work on PR for `graphql-js`: https://github.com/graphql/graphql-js
Just create a draft PR and I will review it.

Also, we need to discuss it on WG, here is the next one:
https://github.com/graphql/graphql-wg/blob/master/agendas/2018-11-08.md
You can add yourself to attendees and a new item to agenda like this:
https://github.com/graphql/graphql-wg/pull/103/files

smitt04:
@IvanGoncharov I have update the PR with suggested fixes. I have also submitted a PR to graphql-wg.

I can start looking into graphql-js, although I haven't worked in javascript graphql for a couple years.  I am mostly working with golang graphql as of late.

IvanGoncharov:
> I can start looking into graphql-js, although I haven't worked in javascript graphql for a couple years. I am mostly working with golang graphql as of late.

NP, you can just open an issue on `graphql-js` and I will try to implement it before WG.

tomasaschan:
I just want to cheer you on and add my hopes that this will be merged and added soon. 🙌 

It's quite a wonderful feeling to realize you're missing a feature, and when you start looking for discussions about it you find this (on graphql/graphql-js#1560):

image

Neitsch:
Should we only allow the deprecated directive on nullable arguments? This can not be expressed in the directive definition, but logically that might make sense.

Anufant:
@IvanGoncharov 
Hi Ivan! 
Could you tell me please, when this PR will be merged?
I want to deprecate some input arguments in my GQL schema very much :)

ydorea:
I'm also really looking forward to this feature!

IvanGoncharov:
@Anufant We discussed it on the last WG and everyone agreed that this issue should be fast-tracked so it definitely going into next spec release. Not sure about the exact date since there are a few things we need to discuss and implement in `graphql-js` before merging this.
You can read about the necessary RFC stages here: 
https://github.com/facebook/graphql/blob/master/CONTRIBUTING.md#rfc-contribution-champions

tjkirch:
Awesome, glad this is moving!  Does that mean the "Next Stage?" tag can be updated to a later stage?  :)

dice14u:
What is the next step here, it seems it has a champion with smitt04 having written both the propsal here and a (now stale) implementation in graphql-js "discussed it on the last WG and everyone agreed .. definitely going into next spec release" sounds like acceptance by WG so stage 3.  Is there something a community member can do to get this merged in on the fast tracked path ?

KenanSulayman:
Revisiting. Can this be merged soon?

xuorig:
👍 very much looking towards this. We've been "faking" argument deprecation (through argument descriptions) and its been working well.

What does it mean for a non-null argument/input value to be deprecated? I'm not seeing this discussed anywhere. At the very least implementations should probably not allow it without making the argument optional first, but I'm not sure what it means in terms of definition only. At GitHub our deprecation DSL warns you to make that field optional first.

  - Should a GraphQL server demote a required argument to nullable in practice when it is deprecated?
  - Is a required input value that is deprecated a valid state for a schema to be in?

axelson:
Is it possible to get a status update on this? (was going to ask why "CLA Signed" was removed, but it looks like that entire label was removed so that makes sense)

IvanGoncharov:
Hi @smitt04 
Sorry for the silence.
@andimarek brought this topic on the last WG and consensus was that there are no blockers We just need to update this PR and `graphql-js` PR and can move it to Stage2 on the next WG.
Do you still have time to work on this feature and present it on the next WG?
If you want I can help with `graphql-js` PR and rebase it on top current master?

smitt04:
@IvanGoncharov I updated this PR and it would be great if you can update `graphql-js`

modosc:
@IvanGoncharov @smitt04 if you need a hand let me know, i'm happy to help out to get this moving forward.

sonnenhaft:
pls merge

leebyron:
This needs a rebase and some cleanup before it can be considered Stage 2

lifeiscontent:
I can't wait for this 🤩

MidnightDesign:
> I can't wait for this 🤩

Please use reactions instead of spamming peolpe who are watching this issue.

leebyron:
Capturing discussion notes from WG meeting:

There is one remaining pitfall from this change, and that's deprecating a **required** argument or input field. There are two problems with this:

1. Deprecations are intended to be actionable: a client should be able to edit queries to contain no deprecated elements. It would not be possible to remove a deprecated required argument.

2. Introspection does not include deprecated portions of the schema by default, you must pass `(includeDeprecated: true)` to the relevant potion. This could result in a scenario where a client omits the deprecated portion of a schema from introspection and therefore does not know about an required argument. What appears valid during client development is rejected as invalid by the service, a surprising outcome. *Note that this is not an abstract concern since no current tool asks for deprecated arguments or input fields and so all current tools could suffer this scenario.*

The recommended addition to finalize this spec proposal (and linked PR) is to include a schema validation rule whereby any deprecated argument or input field is not also required. There is an open question whether this should be a "must" or a "should" rule, which @eapache will help investigate - either way the reference implementation GraphQL.js will include this schema validation rule.

eapache:
Apologies that it took me much longer than expected to get back to this. I've discussed this with my team internally and here's the current state of my thoughts:

> 1. Deprecations are intended to be actionable.

I don't think this is really a problem. Even setting aside our local schema versioning system (which was the motivating counter-example), not all deprecations are going to be actionable "inside the schema"; GraphQL should support deprecating required arguments or inputs where the action to be taken by the client is more complicated than just "use another field".

In our particular case, the action for any deprecation (even for output fields) always includes "upgrade your API version" (effectively switching to a different snapshot of the schema) alongside any necessary field/argument changes.

> 2. Introspection does not include deprecated portions of the schema by default.

This is a much bigger problem. It wouldn't surprise me if some developers use introspection dynamically in production, which would mean that simply deprecating a required input becomes a live breaking change for those clients.

---

My answers to these two problems are notably in conflict. For the purpose of supporting deprecations with actions "outside the schema", I do believe GraphQL should permit deprecating required inputs. However, this is a significant problem for tooling and introspection.

One option I think worth considering is simply changing the rules for introspection: if introspection always returns required parts of the schema by default (even when they're deprecated and `includeDeprecated: false`) then that problem vanishes. It does complicate deprecation a little bit, but in return you get increased expressivity (required arguments can be marked deprecated) and the underlying heuristic (introspection should always tell you the things you absolutely need to know) seems sound.

If that isn't acceptable, then I think making this a "MUST" validation is ok. There will be some cases on some schemas where `@deprecated` won't be usable, but it will still cover 99% of cases without any extra complexity.

eapache:
A brief summary of the discussion at the working group meeting: there was resistance to making changes or adding inconsistencies to introspection. In general, the GraphQL spec is pushing a very specific model of how schemas should evolve, and doesn't strongly feel a need to support alternative models. There was a proposal to use `SHOULD` language anyway, but given the introspection problem we probably wouldn't use it, so the conclusion was to just use `MUST` as originally proposed.

vlad-accedo:
wow

smyrick:
FYI for those tracking the issue, it has not officially merged into the spec yet, it has just moved to the next stage. See the actual changes here: https://github.com/graphql/graphql-spec/pull/805

Issue #410, opened 2018-02-26
Directives at the top of a SDL file


sorenbs:
In the latest [Working Group meeting](https://github.com/graphql/graphql-wg/blob/master/notes/2018-02-01.md) Graphcool presented [GraphQL Import](https://github.com/graphcool/graphql-import). It was brought up that we could change the syntax to be based on directives instead of comments:

```graphql
@import(defs: [Post, Comment] from: "./other-types.graphql")
```

instead of

```graphql
# import Post, Comment from "./other-types.graphql"
```

In the current spec, directives must always be placed after another node. As such, there is no way to place a directive at the top of a file. I volunteered to create an RFC to rectify this using the `@import` directive as motivation.

Discussion in the issue for [graphql-import](https://github.com/graphcool/graphql-import/issues/89#issuecomment-368077528) let to the conclusion that import should not switch to a directive based syntax.

I still think it might be valuable to allow directives in the beginning of a file and would love to hear other use cases for this before proceeding with an RFC.

benjamn:
If we didn't have top-level directives, you could still do a lot with _type_-level `@import` directives:
```gql
type Post @import(from: "./other-types.graphql") {}
type Comment @import(from: "./other-types.graphql") {}
```
Since you're importing into the `Post` and `Comment` types here, you wouldn't have to specify what to import in the arguments to the `@import` directive. The implementation of `@import` could infer what to import by examining the name of the type where the `@import` directive was found. In other words, there's no need for the `defs: [Post, Comment]` argument.

Presumably you could also rename types imported from the source schema, using other optional `@import` arguments, but I suspect that would not be necessary most of the time.

If we did have top-level directives, I imagine they would be equivalent to directives on the `schema` declaration, since a top-level schema directive should naturally operate on the `GraphQLSchema` object:
```gql
schema @import(...) {
  query: Query
  mutation: Mutation
}
```

Either way, there's a problem here, because anyone coming from other languages will probably assume they can have more than one top-level `@import` directive, but that's currently forbidden by the [directive uniqueness rules](http://facebook.github.io/graphql/draft/#sec-Directives-Are-Unique-Per-Location). Should we relax that restriction?

[...]

PR #556, opened 2019-01-31
[RFC] Document Directives

langpavel:
This addresses #410 *Directives at the top of a SDL file*

## Document Directives
```
DocumentDirective :
  - Directive[Const]
```

*DocumentDirective* is special kind of *Definition* and should be handled
on AST level by implementation in similar way like *TypeSystemDefinition*
or *ExecutableDefinition*.

Documents of any kind may contain zero or more *DocumentDirectives*.
Same directive may be repeated multiple times at document level:

```graphql example
@import(from: "./common")
@import(from: "./user")
```

Unknown directives may be ignored, but it is recommended for services to fail
if directive is not recognized.

:heavy_check_mark: Required grammar and introspection changes included


spec/Section 2 -- Language.md
```diff
@@ -1094,3 +1095,24 @@ and operations.
...
+{DocumentDirective} is special kind of {Definition} and should be handled
+on AST level by implementation in similar way like {TypeSystemDefinition}
+or {ExecutableDefinition}.
+
+Documents of any kind may contain zero or more {DocumentDirective}s.
+Same directive may be repeated multiple times at document level:
+
+```graphql example
+@import(from: "./common")
+@import(from: "./user")
```
IvanGoncharov: I don't think we should use `@import` directives as an example here.
Since some readers can think it a standard one and supported by GraphQL spec.
I think it would be better to use something like `@someTopLevelDirective`, `@foo`, `@test`, etc.

spec/Section 2 -- Language.md
```diff
@@ -1094,3 +1095,24 @@ and operations.
...
+```
+
+Unknown directives may be ignored, but it is recommended for services to fail
+if directive is not recognized.
```
IvanGoncharov: @langpavel It contradicts rules for all other directives:
https://facebook.github.io/graphql/June2018/#sec-Directives-Are-Defined

Personally, I think we shouldn't have any exceptions for top-level directives.

spec/Appendix B -- Grammar Summary.md
```diff
@@ -94,10 +94,14 @@ lines and uniform indentation with {BlockStringValue()}.
 Document : Definition+
 
 Definition :
+  - DocumentDirective
```
IvanGoncharov:
```suggestion
  - DocumentDirectives
```

langpavel: No, `DocumentDirective` (singular) is really what should be stated in grammar.

spec/Section 2 -- Language.md
```diff
@@ -1094,3 +1095,24 @@ and operations.
 
 As future versions of GraphQL adopt new configurable execution capabilities,
 they may be exposed via directives.
+
+
+## Document Directives
```
IvanGoncharov: Don't think they should be a special case of directives.
They should be covered by: https://facebook.github.io/graphql/June2018/#sec-Language.Directives

spec/Section 2 -- Language.md
langpavel: So, this sentence can be dropped entirely, ok?

spec/Appendix B -- Grammar Summary.md
IvanGoncharov: I see now
Then I think it makes more sense to do
Document: Directives[Const] Definition+

spec/Section 2 -- Language.md
langpavel: This is most relevant example I wish to show and indeed it is my intent to push on some common way to do import. I can add note or warning below that import is not built-in

spec/Section 2 -- Language.md
langpavel: Intent is place document level directives aside from others, there apply different rules:
- Document directives may be repeated, others not.
- Document level directives are top level **AST Document** `DocumentNode.definitions` nodes, other directives are listed in `XxxNode.directives` — this may need discussion

spec/Appendix B -- Grammar Summary.md
langpavel: This will prevent you from inserting document level directives into middle of document.
This may be problematic if you *concatenate* more documents into one.

langpavel: Moreover there is chicken egg problem,  when you wish define document directive you cannot use it in same document

spec/Section 2 -- Language.md
IvanGoncharov: > Document directives may be repeated, others not.

#472 add support of reputable directives.

> Document level directives are top level AST Document DocumentNode.definitions nodes, other directives are listed in XxxNode.directives — this may need discussion

I think `DocumentNode.directives` makes more sense.

IvanGoncharov:
I think we need to have a separator between top-level directives and rest of SDL file
```
@import(from: "./common")
@import(from: "./user")
---

type Query {
  foo: String
}
```
Similar to YAML: https://yaml.org/spec/1.1/#YAML%20directive/

Because without it we can cause some confusion:
```
@someDirective

scalar Url
```
`@someDirective` is top level directive but some user may think that `@someDirective` belongs to scalar definition especially since it looks very similar to decorators in JS and many other languages.

Plus I saw some people concatenate SDL files so if you have two files like this:
```
scalar SomeScalar
```
and 
```
@someDirective

scalar AnotherScalar
```
We would get valid SDL:
```
scalar SomeScalar
@someDirective

scalar AnotherScalar
```

But parser will parse `@someDirective` as part of `SomeScalar` definition.
With 3 dashes we will have:
```
scalar SomeScalar
@someDirective
---
scalar AnotherScalar
```
The parser will error saying you can't use `---` in the middle of the file.

If we decide to allow top-level directives we can always allow:
```
scalar SomeScalar

---
@someDirective
---

scalar AnotherScalar
```

spec/Appendix B -- Grammar Summary.md
langpavel: Oh, you are right…

langpavel:
Hmm, good point with grammar ambiguity, however I don't like `---` separator,
What about new keyword `document` followed by document level directives?

spec/Section 2 -- Language.md
mjmahone: As `@import` will have meaning in a bunch of libraries almost immediately, and those meaning will probably immediately diverge, let's keep it out of the spec entirely. There aren't any descriptions of any other user-level directives in the spec.

The only other usage is `@addedDirective` and `@example` because they illustrate the usage without introducing a new concept.

In this case, you could call it `@exampleDocumentDirective` or even just `@example`

Issue #566, opened 2019-02-28
Allow directives on directive definitions


benhead:
We'd like to be able to declare directives on directive definitions.

We have tooling that uses type system directives to perform special processing on a schema, and we'd like users of that tooling to be able to define their own directives, with metadata indicating to the tooling that this directive is for its use. For example, if we could define `directive @Handler(class:String!) on DIRECTIVE_DEFINITION` then our users could define `directive @CustomDirective(...) on FIELD_DEFINITION @Handler(class:"Foo")` telling our tooling that their directive is meant to be handled by it, passing arguments we declare to our tooling, and still providing whatever arguments they declare to their own hooks.

That directives can't be declared on directive definitions seems like an oversight.

IvanGoncharov:
@benhead Thanks for describing this proposal and especially for the PR.
I think having #567 is enough to track this proposal.
I'm closing this one since it's better to have entire disscussion in one place.

@benhead Can you please edit your initial comment in #567 to include detail description from this issue?

PR #567, opened 2019-02-28
allow directives on directive definitions

benhead:
This is the minimum viable edit to enable this feature. I imagine we'd also want to allow directive extensions if this feature is approved.

#566

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1607,7 +1607,7 @@ Expected Type | Internal Value   | Coerced Result
 
 ## Directives
 
-DirectiveDefinition : Description? directive @ Name ArgumentsDefinition? on DirectiveLocations
+DirectiveDefinition : Description? directive @ Name ArgumentsDefinition? on DirectiveLocations Directives[Const]?
 
 DirectiveLocations :
   - `|`? DirectiveLocation
@@ -1639,6 +1639,7 @@ TypeSystemDirectiveLocation : one of
   `ENUM_VALUE`
   `INPUT_OBJECT`
   `INPUT_FIELD_DEFINITION`
+  `DIRECTIVE_DEFINITION`
 
 A GraphQL schema describes directives which are used to annotate various parts
 of a GraphQL document as an indicator that they should be evaluated differently


benhead:
Original feature request:

> We'd like to be able to declare directives on directive definitions.
> 
> We have tooling that uses type system directives to perform special processing on a schema, and we'd like users of that tooling to be able to define their own directives, with metadata indicating to the tooling that this directive is for its use. For example, if we could define `directive @Handler(class:String!) on DIRECTIVE_DEFINITION` then our users could define `directive @CustomDirective(...) on FIELD_DEFINITION @Handler(class:"Foo")` telling our tooling that their directive is meant to be handled by it, passing arguments we declare to our tooling, and still providing whatever arguments they declare to their own hooks.
> 
> That directives can't be declared on directive definitions seems like an oversight.

Please note that I'm a Googler and we're reviewing that we can sign the CLA. I don't anticipate a problem, but bear it in mind until we confirm.

spawnia:
I would like to another use case that requires this change that might be a bit more common: Deprecation of directives.

```graphql
directive @foo on OBJECT @deprecated
```

spec/Section 3 -- Type System.md
```diff
@@ -1838,78 +1607,50 @@ Expected Type | Internal Value   | Coerced Result
 
 ## Directives
 
-DirectiveDefinition : Description? directive @ Name ArgumentsDefinition? `repeatable`? on DirectiveLocations
+DirectiveDefinition : Description? directive @ Name ArgumentsDefinition? on DirectiveLocations Directives[Const]?
```
benjie: One thing to consider is that this position would prevent future deprecation of one of the positions:

```
directive @foo on
| QUERY
| MUTATION @deprecated(reason: "Doesn't work well with side effects")
| SUBSCRIPTION
```

(I don't think this is particularly important, but it's worth noting.)

martinbonnin: You mean because the syntax would become ambiguous, right? I think I'm okay with this (famous last words 😅 )

martinbonnin:
Closing in favor of https://github.com/graphql/graphql-spec/pull/1206, which reached RFC1 during yesterday's working group. Thanks @BoD for presenting!

PR #574, opened 2019-03-28
Add missing '&' to Punctuator production (graphql#573)

CornedBee:
This resolves graphql#573.

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -39,7 +39,7 @@ Token ::
   - FloatValue
   - StringValue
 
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@CornedBee Can you please keep this issue specific to `&`?
Please open a separate PR if you see any other problems with GraphQL grammar.

CornedBee:
Sorry, that was accidental. I forgot that the PR would be updated if I just push to master. Will fix.

IvanGoncharov:
@CornedBee No problem.
Thank you for contributing and sorry for the delay in merging it.
The GraphQL spec currently in a process of migrating from FB into newly formed GraphQL Foundation so it causes some delays.

PR #538, opened 2018-11-24
Synchronise grammar rules across the spec

IvanGoncharov:

PR #576, opened 2019-04-10
Fix typo in schema definition grammar productions.

CornedBee:
The Schema section refers to and defines RootOperationTypeDefinition.
This production is missing in the appendix.

The SchemaExtension section refers to OperationTypeDefinition.
This production is not defined anywhere except in the appendix.

The two productions are identical. They are meant to be the same thing.


IvanGoncharov:
@CornedBee I think I already fixed it in as part of #538.
Can you please review it and confirm?

CornedBee:
You are right, that's the same issue.

Issue #399, opened 2018-01-20
Why does GraphQL merge fields across fragment type targets?


mike-marcacci:
I've recently been fighting with some of the ergonomic downsides to [field selection merging](http://facebook.github.io/graphql/October2016/#sec-Field-Selection-Merging) – notably field conflict errors and their stifling frequency when dealing with heterogeneous lists and mutations with relay. As I've been thinking through various solutions, I've become curious as to why merging happens at all.

And so that's my question for people with a better knowledge of GraphQL's history: why do we merge fields? I'm sure this was discussed at some point, but I couldn't find it, so I'm going to make a few arguments against merging, with the suspicion that somebody will have a better argument *for* merging that I missed. :-)

### Better Input/Output Symmetry
Input/output symmetry is one of GraphQL's features that helps make it so immediately intuitive. To me, the following query:

```graphql
query {
	something {
		id

		...on Dog {
			breed
		}

		...on Car {
			make
		}
	}
}
```

could just as intuitively return:

```json
{
	"something": {
		"id": "9121c30a-941a-4a05-b6ae-c202b6d62522",

		"__on<Dog>": {
			"breed": "Huskey"
		},

		"__on<Car>": {
			"make": "Toyota"
		}
	}
}
```

as its merged counterpart.

### Eliminates Most Field Conflicts
This also greatly reduces the liklihood of field conflicts, meaning queries like the one below can actually be run:

```graphql
query {
	someGeoQuery {
		id

		...on POI {
			__typename
			coordinates #GeoJSONPointCoordinates
		}

		...on City {
			__typename
			coordinates #GeoJSONPolygonCoordinates
		}
	}
}
```

### Better Interop With Clients & Tools Like Relay
My biggest frustration with the current implementation is not being able to actually reuse fragments when I expect a heterogeneous response. For example, I have an offline-first app where changes are staged in a changeset; this changeset can then be committed when the user is online. Because of precautionary field merge conflicts (not *real* conflicts, but ones that may exist in the future if I changed a concrete type to be an interface) it's exceedingly difficult to reuse fragments for the mutation response. There are plenty of ways around this, but I'm not sure this is the kind of thing that should need to be worked around.

### Simpler Server Implementation
The simplest way to do something is *almost* always to not do it :-)

---

Hopefully I've made some interesting points here, and I'm very curious to hear some thoughts in the opposite direction!

leebyron:
[...]

Issue #215, opened 2016-09-30
Proposal: Support union scalar types


stubailo:
It makes sense often to have a field that could return, for example, an `Int` or a `String`. It would be convenient to be able to use a `resolveType`-like mechanism to determine how the value is serialized. This would be even more useful for custom scalars.

The specification currently says: "GraphQL Unions represent an object that could be one of a list of GraphQL Object types", which basically means that unions are the same as interfaces, but without any shared fields. Supporting unions of scalars would further differentiate unions from interfaces.

rmosolgo:
Could you share a couple of concrete use cases? I ask because, personally, I'm usually reading values right out of MySQL (or running those values through transformations with predictably-typed output), so I haven't needed this yet!

migueloller:
@rmosolgo,

Here is a concrete use case that we've come across at my team. I have a couple more if you need them.

We've got an algorithm that produces results based on location. We have an enum `AreaOfTown` that contains all possible areas of town to filter the results. Here's the catch, the algorithm can also filter locations by "near me". Because we can't simply make a union that is `AreaOfTown | "NEAR_ME"` for the input type, then we have to create a completely new enum that contains all of the values in `AreaOfTown` and `"NEAR_ME"`. This is undesirable because if we ever change the `AreaOfTown` enum by adding more (this will 100% happen in the future), then we have to make sure we also update our other enum.

If we could write scalar unions then it would solve this problem for us, among others (we would like to have constructs like `Int | [Int!]` or `Int | String` like @stubailo mentioned).

stubailo:
In our case, it's because we were returning a structure that represented a JSON tree, where keys can be integers (in the case of arrays) or strings (in the case of objects). We ended up compromising for now to just use strings, but I can imagine how this would be extra useful in the case where you want to return one of a set of custom scalars, which might have different serialization logic.

migueloller:
Here are other relevant discussions:
- graphql/graphql-js#207
- graphql/graphql-js#291
- #202

@leebyron made some good points in his comments [here](https://github.com/graphql/graphql-js/issues/207#issuecomment-243638111).

marcintustin:
Allowing uniontypes to contain primitives allows us to model any kind of source that can be polymorphic in its source data. For example, avro data or elasticsearch. 

This is a severe limitation on exposing existing data, rather than creating new systems that are defined by graphql from the ground up.

leebyron:
[...]

PR #586, opened 2019-05-18
[RFC] Input Objects accepting exactly @oneField

benjie:
This is an RFC for adding explicit GraphQL support (and with it improved type safety) for a pattern that supports inputs of multiple types (composite and scalar).

## Explanation of the problem

Often a GraphQL field needs to accept one of many input types.

Example 1 (below) outlines the need for the `addContent` mutation to accept a range of different content types.

Example 2 (below) outlines the need for the `findPosts` record finder to accept a range of different filtering criteria.

Currently in GraphQL the type safe solution to this is to have different fields, one per type:

Example 1 (more details in "Example 1" section below):
- `addPostContent(post: [PostInput!]!)`
- `addImageContent(image: [ImageInput!]!)`
- `addHrefContent(href: [String!]!)`

Note to add lots of content blocks of different types you must add multiple mutations to your GraphQL request, which is a lot less efficient and loses the transactional nature of the request.

Example 2 (more details in "Example 2" section below):

- `findPostsByIds(ids: [Int!])`
- `findPostsMatchingConditions(conditions: [PostCondition!])`

Another solution is the "tagged input union" pattern, which results in a less overwhelming schema thanks to fewer fields, solves the issue of inputting lots of data in a single mutation, but loses type safety:

```graphql
input MediaBlock {
  post: PostInput
  image: ImageInput
  href: String
}
extend type Mutation {
  addContent(content: [MediaBlock!]!): AddContentPayload
}
```

(The loss of type safety is that each block should require that exactly one of the fields is set, however GraphQL does not enforce this so it must be done at run-time, and generic tooling will not be able to confirm the validity of inputs.)

## Prior Art

This RFC is similar to the "tagged input union" pattern discussed as an alternative in the `inputUnion` RFC #395 (please see the reasons people may not be happy with it there also). Since we are failing to reach consensus of on the `inputUnion` type in that proposal, I'd like to revisit [Lee's original counter-proposal](https://github.com/graphql/graphql-spec/pull/395#issuecomment-361373097) with the small addition of a directive to require that exactly one field be present on the input object (as [originally proposed by Ivan](https://github.com/graphql/graphql-spec/pull/395#issuecomment-361704974)).

### The input object `requiresExactlyOneField` property (exposed via `@oneField`)

This new `requiresExactlyOneField` introspection property for input objects can be exposed via IDL as the `@oneField` directive (in a similar way to how the `deprecationReason` introspection property is exposed as `@deprecated(reason: ...)`).

This property applies to an input object type:

```graphql
input MyInput @oneField {
  field1: Type1
  field2: Type2
  field3: Type3
}
```

(In the IDL, we could alternatively expose it as a different type: `inputOne MyInput`, rather than `input MyInput @oneField`, if using a directive was found to be confusing.)

The fields of an input object type defined with `@oneField` must all be nullable, and must not have default values. On input, exactly one field must be specified, and its value must not be null.

(For the avoidance of doubt: the field types may be scalars or input objects, and the same type may be specified for multiple fields. This is the same as for regular input objects.)

I've called the flag `@oneField` (and the related introspection field `requireExactlyOneField`), but it could equally be called `@requireExactlyOneField`, `@exactlyOne`, `@one`, or many other variants depending on your position in the terseness vs explicitness spectrum.

### Example 1

```graphql
input PostInput {
  title: String!
  body: String!
}

input ImageInput {
  photo: String!
  caption: String
}

"""
Each media block can be one (and only one) of these types.
"""
input MediaBlock @oneField {
  post: PostInput
  image: ImageInput
  href: String
}

type Mutation {
  addContent(content: [MediaBlock!]!): Post
}
```

```graphql
mutation AddContent($content: [MediaBlock!]!) {
  addContent(content: $content) {
    id
  }
}

```

Example input:

```
{
  content: [
    { post: { title: "@oneField directive", body: "..." } },
    { image: { photo: "https://..." } },
    { href: "https://graphql.org" }
  ]
}
```

### Example 2

```graphql
"""
Options available for matching a post title
"""
input PostTitleFilter @oneField {
  equals: String
  contains: String
  doesNotContain: String
}

"""
You may search within the title of the post, or within it's full text.
"""
input PostCondition @oneField {
  title: PostTitleFilter
  fullTextSearch: String
}

"""
Identify the relevant posts either by a list of IDs, or a list of conditions
for them to match.
"""
input PostMatch @oneField {
  ids: [Int!]
  conditions: [PostCondition!]
}

type Query {
  findPosts(matching: PostMatch!): [Post!]
}
```

```graphql
query FindPosts($matching: PostMatch!) {
  findPosts(matching: $matching) {
    id
    title
  }
}
```

Example inputs:

```
{
  matching: {
    ids: [27, 30, 93]
  }
}
```

```
{
  matching: {
    conditions: [
      { fullTextSearch: "GraphQL" },
      { title: { contains: "Facebook" } }
    ]
  }
}
```

### Example 3

```graphql
input OrganizationAndEmailInput {
  organization: String!
  email: String!
}

"""
Conditions that uniquely identify a user, supply exactly one.
"""
input UserWhere @oneField {
  id: ID
  databaseId: Int
  organizationAndEmail: OrganizationAndEmailInput
  username: String
}

type Query {
  getUser(where: UserWhere!): User
}
```

```graphql
query GetUser($where: UserWhere!) {
  getUser(where: $where) { id databaseId username email organization { id name slug } }
}
```

Example inputs:

```
{ where: { id: 27 } }
```

```
{ where: { username: "Benjie" } }
```

```
{
  where: {
    organizationAndEmail: {
      organization: "graphql-wg",
      email: "example@example.com"
    }
  }
}
```

### Backwards Compatibility

This change involves a small addition to the introspection system. Old clients will continue to work as they do currently, and can issue requests involving `@oneField` input objects without needing to be updated, they simply will not benefit from the type validation of knowing exactly one field is required.

All pre-existing queries (and schemas) will still be valid, and old clients can query servers that support this feature without loss of functionality.

### Performance is a feature

There is minimal overhead in this change; input objects continue to work as they did before, with one additional validation rule (that exactly one field must be specified). This can be asserted during the regular [Input Object Required Fields Validation](https://graphql.github.io/graphql-spec/draft/#sec-Input-Object-Required-Fields) at very little cost. (In the spec I've written it up as a separate rule, but implementations can implement this more efficiently.)

### Favour no change

The change itself is small (and hopefully easy to implement), whilst bringing significant value to type safety. It's not possible to represent this use case in a type safe way in GraphQL currently, without having a proliferation of object type fields such as `addPostMediaBlock`, `addImageMediaBlock`, ... (which gets even more complex when there's nested concerns).

### Enable new capabilities motivated by real use cases

It's clear from #395 and #488 (plus the multiple discussions during GraphQL working groups, and the community engagement) that a feature such as this is required. It's currently possible to express this in GraphQL (simply omit the `@oneField` directive in the examples above), however it is not expressed in a type-safe way and requires the schema's resolvers to perform validation, resulting in run-time rather than compile-time errors (and no hinting in editors/tooling).

### Simplicity and consistency over expressiveness and terseness

I believe this has been achieved (bike-shedding over the directive name notwithstanding).

### Preserve option value

This RFC is only on using this directive with input objects, I have been deliberately strict to keep the conversation on topic. There is room for expansion (and even for the addition of an explicit `inputUnion`) in future.

### Understandability is just as important as correctness

I've done my best; please suggest edits that may make the spec changes more clear 👍

## Comparison with inputUnion

The field name takes the place of the "discriminator" discussed in #395.

The implementation is significantly simpler, for example the only change required to support autocompletion in GraphiQL is to stop autocompleting once one field has been supplied.

The implementation is backwards compatible: older clients may query a schema that uses `@oneField` without loss of functionality.

The implementation is forwards compatible: newer clients may query a schema that implements an older GraphQL version without causing any issues (whereas the unexpected addition of `__inputname` may break existing GraphQL servers, hence complexity of only specifying it where a union is used).

There's no `__inputname` / `__typename` required, and heterogeneous input types are still supported.

The field types may be input objects or scalars.

This pattern is already in use in the wild, but with the execution layer handling validation, rather than the schema itself.

`@oneField` does not mirror output unions/interfaces.

`@oneField` increases the depth of input objects vs the `inputUnion` proposal (but, interestingly, does not increase the size of the request:

```
# `inputUnion` example
{ pets: [
  { __inputname: "Cat", lives: 3 },
  { __inputname: "Dog", wagging: true }
]}
```

```
# `@exactlyOne` example
{ pets: [
  { cat: { lives: 3 } }
  { dog: { wagging: true } }
]}
```

## Potential concerns, challenges and drawbacks

- It's not called an `input union`, and does not mirror the `union` syntax for output types
- It is not an ideal syntax (is there one?), and adds an additional layer of depth to input objects
- It requires a small addition to the introspection system: `requiresExactlyOneField`

This pattern is already used in the wild, and the introduction of this feature will not break any existing APIs. APIs that already use this pattern could benefit from this feature retroactively without needing to update any any existing clients.


spec/Section 5 -- Validation.md
```diff
@@ -1389,6 +1389,24 @@ arguments, an input object may have required fields. An input field is required
...
+    * For the sole {fieldDefinition} in {fieldDefinitions}:
+      * Let {fieldName} be the name of {fieldDefinition}.
+      * Let {field} be the input field in {fields} named {fieldName}
+      * Let {value} be the value of {field}.
+      * {value} must not be the {null} literal.
+
+**Explanatory Text**
+
+When an input object requires exactly one field, exactly one field must be
+supplied and that field may not be null.
```
jadedevin13: > supplied and that field **must** not be null.

binaryseed:
Re: names..  This is called `oneof` in protobufs https://developers.google.com/protocol-buffers/docs/proto#oneof and I think that name feels pretty intuitive

binaryseed:
I do like that this seems like a minimal change to achieve a meaningful improvement..

The tricky part is the layer of nesting that it requires. To get a data structure that matches on the query side, it means you'd have to ignore `interface` and `union` and re-create that via field nesting in the same way.

I think this is used in the wild (we use it at New Relic) not because it's ideal, but because it's the only work around available...

spec/Section 5 -- Validation.md
benjie: Fixed 👍

benjie:
Due to confusion, I've removed the word `directive` from a number of places. This PR really outlines an introspection property on input objects named `requiresExactlyOneField` which, like `deprecationReason`, happens to be exposed via IDL as a directive. It is not truly a directive.

akomm:
@benjie I like your proposal. 

I want to add something regarding this statement:
```
@oneField increases the depth of input objects vs the `inputUnion` proposal (but, interestingly, does not increase the size of the request:
```

In some cases it would not even increase the depth compared to `inputUnion`. For example when the variant is an array

Example with `@oneField`:
input PostMatch @oneField {
  ids: [Int!]
  conditions: [PostCondition!]
}

Example with `inputUnion`:
```gql
inputUnion PostMatch = PostMatchByIds | PostMatchByConditions
# input PostMatchByIds = [Int!] <- not possible
# scalar PostMatchByIds = [Int!] <- not possible
# inputUnion PostMatch = [Int!] | [PostCondition!] <- not possible
# same depth:
input PostMatchByIds = {
  ids: [Int!]
}
```
How would you have to define the above inputUnion to have less depth than using `@oneField`? I could not find a solution in the graphql spec. Please correct me if I am wrong.

If it is a scalar, you would have to declare a custom scalar type to preserve the intent, e. G.
```gql
# what is Int?
inputType = int | PostConditions
# vs
scalar PostId
inputType = PostId | PostCondition
```
Also additionally loosing the Int validation.

While the `@OneField` approach increases the depth in some cases, it removes the number of types required, as you might have already noticed in the above example.

Given your 2. example:
```gql
"""
Options available for matching a post title
"""
input PostTitleFilter @oneField {
  equals: String
  contains: String
  doesNotContain: String
}

"""
You may search within the title of the post, or within it's full text.
"""
input PostCondition @oneField {
  title: PostTitleFilter
  fullTextSearch: String
}

"""
Identify the relevant posts either by a list of IDs, or a list of conditions
for them to match.
"""
input PostMatch @oneField {
  ids: [Int!]
  conditions: [PostCondition!]
}

type Query {
  findPosts(matching: PostMatch!): [Post!]
}
```
You would need this number of types, to preserve the intent of the variants:
- PostMatch (union)
- PostMatchByIds (input)
- PostMatchByConditions (input)
- PostCondition (union)
- PostConditionTitleFilter (union)
- PostConditionTitleFilterEquals (*custom scalar)
- PostConditionTitleFilterContains (* custom scalar)
- PostConditionDoesNotContain (* custom scalar)
- PostConditionFullTextSearch (* custom scalar)

*custom scalar: because `String | String | String` doesn't make any sense. Or you would create an input type for each option like this:
```gql 
input PostConditionTitleFilterEquals {
    value: String!
}
```
Hereby again having not less depth than required with `@oneField`.

benjie:
Interesting analysis, @akomm; thanks for sharing. I will definitely be referencing this when I next get a few minutes to advance this proposal.

benjie:
@binaryseed I was thinking more about your `oneof` suggestion; I like that it already exists as a concept, and that it's even shorter, but am hesitant over it's all-lower-case value (makes it hard to pronounce if you've not seen it before), and feel that perhaps `oneField` is more explicit/transparent. I do like it though. Would you consider updating your suggestion to `@oneOf`, or is the all-lower-case important to you?

xuorig:
`@oneOf` is also used by the [OpenAPI spec](http://spec.openapis.org/oas/v3.0.2#fixed-fields-20). Although `oneOf` usually implies it will receive a list of members: `@oneOf(fields: [FieldName])` vs saying only one of this type's field will be present 🤷‍♂

binaryseed:
> is the all-lower-case important to you?

it's not. `@oneOf` looks great

spec/Section 3 -- Type System.md
```diff
@@ -1650,6 +1664,10 @@ GraphQL implementations that support the type system definition language must
 provide the `@deprecated` directive if representing deprecated portions of
 the schema.
 
+GraphQL implementations that support the type system definition language must
+provide the `@oneField` directive if representing input types that should
```
akomm: Rather `must` instead of `should`?

I think the possibility of optional input is handled via making the input object itself null-able, but not its variants if it is a tagged union.

spec/Section 3 -- Type System.md
```diff
@@ -1440,6 +1440,20 @@ Input object type extensions have the potential to be invalid if incorrectly def
    the original Input Object.
 5. Any directives provided must not already apply to the original Input Object type.
 
+### Input Objects Requiring Exactly One Field
+
+A GraphQL schema may wish to represent a situation where an input could take
+one of many different options. When using the type system definition language,
+`@oneField` is used to indicate that the input object accepts exactly one
```
akomm: `accepts` to `requires` ?

spec/Section 3 -- Type System.md
benjie: How about referencing the title?

> GraphQL implementations that support the type system definition language must
provide the `@oneField` directive if representing input types that implement
the "Input Objects Requiring Exactly One Field" feature.

(Can I link to the title directly somehow?)

spec/Section 3 -- Type System.md
benjie: Done

spec/Section 3 -- Type System.md
akomm: Title in which context? Like [this](https://github.com/benjie/graphql-spec/blob/one-field/spec/Section%205%20--%20Validation.md#input-object-with-requires-exactly-one-field)?

benjie: Done 👍

akomm:
I noticed one thing, which I am not sure about whether this is a problem or not:

You have added this section in validation: [Input Object With Requires Exactly One Field](https://github.com/benjie/graphql-spec/blob/one-field/spec/Section%205%20--%20Validation.md#input-object-with-requires-exactly-one-field)
Just above it, there is another section which handles the [Input Object Required Fields](https://github.com/benjie/graphql-spec/blob/one-field/spec/Section%205%20--%20Validation.md#input-object-required-fields)

Now I am not expert at spec writing, but when I would start implement the [later](https://github.com/benjie/graphql-spec/blob/one-field/spec/Section%205%20--%20Validation.md#input-object-required-fields), I am actually already doing it wrong, because I have at this point to take into account the new `oneField` logic and exclude those input objects with the `oneField` enabled.

benjie:
@akomm You wouldn't be doing it wrong if you did that; both sections need to be honoured for `@oneField` since it is also an input object. However, by combining them you could optimise your code. In the spec, I think the focus is on clarity and correctness rather than on performance, so I kept them separate.

jensneuse:
How about inlining input unions?

```graphql
input UserByID {
  id: ID!
}

input UserByName {
  name: String!
}

type Query {
  getUser(where: UserByID | UserByName): User
}```

**Comment by benjie** [comment]
@jensneuse That's not something I would like to enable with this proposal. There's a lot of questions that raises, please see the discussion in https://github.com/graphql/graphql-spec/pull/395 for more details.

**Comment by mazikwyry** [comment]
@benjie First: Thank you very much for all your contributions here and for trying to solve the problem with inputs. I like this proposal and it will match some use cases, but I feel that it is not universal and eventually, GraphQL may end up with 2 standards, cos other use cases will still not be covered.

`oneField` only applies to inputs where the client just needs to select one of the fields. In many of the examples that I saw in https://github.com/graphql/graphql-spec/issues/488 we have some fields that are required and then some fields where you need to provide one of:

```js
input MediaBlock {
  projectId: PostInputKind!
  title: String!
  photoContent: String
  videoContent: String
}
```

or 

```js
input PageField {
  pageId: ID!
  code: String!,
  singleValue: String
  listOfValues: [String]
}
```

In the examples above we use the tagged union approach which lacks the validation and `oneOf` won't solve the problem here. I really don't know what is under the hood, but from the API perspective maybe we can have something like this:

```js
input PageField {
  pageId: ID!
  code: String!,
  singleValue: String?, // required if listOfValues is not present
  listOfValues: [String]?, // required if singleValue is not present
  anotherFieldThatIsNotRequired: String
}
```
I mean the client marks fields from which one is required somehow `?` is just and example.

akomm:
@mazikwyry I disagree. The client is not in charge of (knowing/constraining) the schema. But this is the case in your `?` example. 

It is not clear whether you think  'input union' is better than this, or that 'input union' is equaly wrong, because you did not mention it explicitly. If the later is the case, ignore it, i the former is the case, than can you provide `gql` that solves your problem with `input union`?

mazikwyry:
@akomm Hello and thanks for your comment.

> I disagree. The client is not in charge of (knowing/constraining) the schema. But this is the case in your ? example.

I also disagree with myself. This is not what I meant. (Sorry, it was early morning, before coffee). I meant  "I mean that in schema we mark fields from which one is required somehow. ? is just an example"

> It is not clear whether you think 'input union' is better than this, or that 'input union' is equally wrong because you did not mention it explicitly.

I'm not comparing those solutions. Not saying one is better. Just expressing my concerns about `oneOf`.

akomm:
@mazikwyry 

Oh, you are right, should have realized it from the schema above. :)

However, what do you do if you need two groups of `?`-fields in a single type?

For example `@oneField`:
```gql
input PageField {
  pageId: ID!
  code: String!,
  content: PageFieldContent!
  anotherFieldThatIsNotRequired: String
}
input PageFieldContent @OneField {
  singleValue: String
  listOfValues: [String!]
}
```

This allows you do make as many **pick one of** fields as you wish, solves your problem and makes a sum of things I will not enumerate here much better.

mazikwyry:
@akomm Thank you very much. I this is possible with `@OneField` it then solves all my concerns about validation.

benjie:
Regarding the `?` style fields, @akomm has hit the nail on the head; there would be no concise way to then have two alternative "pick ones" in the same set of arguments - I evaluated this before moving on to the `@oneField` solution.\*

@akomm has also outlined what the `@oneField` solution to this would be (i.e. the `content: PageFieldContent` field). Note that input unions also do not solve this particular issue, I think adding an additional level is required to express "A and B and (one of C or D)"; instead you'd say "A and B and E" where E is "one of C or D", and this can then be used for multiple fields. In this example both source and linkTarget are tagged unions, where source is required but linkTarget is optional.

```graphql
input Media {
  title: String
  caption: String
  source: MediaSource!
  linkTarget: LinkTarget
}

input MediaSource @oneOf {
  url: String # e.g. "http://example.com/example.png"
  clipart: String # e.g. "cats"
  savedMedia: String # e.g. "MyPersonalUpload-150.png"
}

input LinkTarget @oneOf {
  web: String # e.g. "http://example.com"
  email: String # e.g. "foo@bar.com"
  internal: String # e.g. "/other/page"
  scriptAction: String # e.g. "flashPage()"
}
```

<details>
<summary>* Click for some rambling about the `?` approach</summary>

\*  (If you were to extend the `?`-like approach so you can specify multiple sets of exclusive fields you could do something like the HTML `radio` input type, where fields that share the same `name` (tag) would be alternatives, e.g.

```graphql
input Foo {
  int: Int @oneOf(tag: "int-or-float")
  float: Float @oneOf(tag: "int-or-float")
  bool: Boolean @oneOf(tag: "bool-or-string")
  string: String @oneOf(tag: "bool-or-string")
}
```

but if you do that then it is (a) cumbersome, and more importantly (b) you can't clearly and consistently specify if those one-ofs are required or not.)
</details>

IvanGoncharov:
We have begun building a single RFC document for this feature, please see:
https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md

The goal of this document is to consolidate all related ideas to help move the proposal forward, so we decided to close all overlapping issues including this one.
Please see discussion in #627

If you want to add something to the discussion please feel free to submit PR against the RFC document.

benjie:
This comment from @derekdon is a good reason why `@oneField` would also be useful on queries: https://github.com/graphql/graphql-spec/issues/215#issuecomment-551309936

Namely if you have the schema:

```graphql
type BoolVal {
  value: Boolean
}

type StrVal {
  value: String
}

union Union = BoolVal | StrVal

type Query {
  test: Union
}
```

Then this GraphQL query does not validate:

```graphql
{
  test {
    __typename
    ... on BoolVal {
      value
    }
    ... on StrVal {
      value
    }
  }
}
```

Because:

```
{ errors:
   [ { GraphQLError: Fields "value" conflict because they return conflicting types Boolean and String. Use different aliases on the fields to fetch both if this was intentional.
```

This is despite the `__typename` being present to disambiguate. With the `@oneField` approach this conflict with the default aliases would not occur and the user can request the data in a type safe (and easy to discover) way without needing to use aliases (and with a smaller query and response size, to boot).

```graphql
type BoolVal {
  value: Boolean
}

type StrVal {
  value: String
}

type Value @oneField {
  bool: BoolVal
  str: StrVal
}

type Query {
  test: Value
}
```

```graphql
{
  test {
    bool {
      value
    }
    str {
      value
    }
  }
}
```

<details>
<summary>Code showing the existing error</summary>

```js
const {
  GraphQLObjectType,
  GraphQLSchema,
  GraphQLBoolean,
  GraphQLString,
  GraphQLUnionType,
  printSchema,
  graphql,
} = require("graphql");

const BoolVal = new GraphQLObjectType({
  name: "BoolVal",
  fields: {
    value: {
      type: GraphQLBoolean,
      resolve: val => val,
    },
  },
  isTypeOf(value) {
    return typeof value === "boolean";
  },
});
const StrVal = new GraphQLObjectType({
  name: "StrVal",
  fields: {
    value: {
      type: GraphQLString,
      resolve: val => val,
    },
  },
  isTypeOf(value) {
    return typeof value === "string";
  },
});

const Union = new GraphQLUnionType({
  name: "Union",
  types: [BoolVal, StrVal],
});

  name: "Query",
  fields: {
    test: {
      type: Union,
      resolve() {
        return Math.random() >= 0.5 ? true : "string";
      },
    },
  },
});

  query: Query,
});

graphql(
  schema,
  `
    {
      test {
        __typename
        ... on BoolVal {
          __typename
          value
        }
        ... on StrVal {
          __typename
          value
        }
      }
    }
  `
).then(result => {
  console.dir(result);
});
```

</details>

akomm:
@benjie I have more and more the feeling, that the current graphql enumeration should have been more powerful in terms of its allowed values and work similar to what the proposed oneField currently works. Basically like the enums you have in rust (which in turn has the idea from other langs, but it does not matter here ^^). It should have been basically a structural oneField "pick one" type, which does not need artificial discriminator for both input and output. But its probably to late to make such changes, at least in graphql v1, should there ever be a graphql v2 it might be considered an option ;P

jturkel:
@benjie - Could your example be handled without `@oneField` if the field selection algorithm was smarter about detecting mutually exclusive fragment branches for unions as discussed in [#399 (comment)](https://github.com/graphql/graphql-spec/issues/399#issuecomment-416436719)? In the example an object could never be both a `BoolVal` and a `StrVal` so there should never actually be a conflict.

benjie:
I think that would work for unions, I'm not sure if it would work for interfaces though - I'd have to think on it harder. Was mostly just noting for my own future reference ;)

spec/Section 4 -- Introspection.md
```diff
@@ -141,6 +141,9 @@ type __Type {
   # should be non-null for ENUM only, must be null for the others
   enumValues(includeDeprecated: Boolean = false): [__EnumValue!]
 
+  # should be non-null for INPUT_OBJECT only, must be null for the others
+  requiresExactlyOneField: Boolean!
```
benjie: Should not have a `!` here.

sabit990928: Sorry for asking here. 
Thanks for the features. But, do you know why @oneOf calls as unknown directive in my project, the same behavior with inputunion. Maybe I'm doing somethiing wrong
@benjie

benjie: GraphQL does not have @oneOf or inputUnion yet; we’re actively working on picking a solution to add to the GraphQL specification. If you’re interested in helping out, please contribute to the input union RFC document.

sabit990928: Thanks for the answer. I'm actually interested in to help you guys, but actually that's my first experience of kinda contribution. That is why I need mentor or someone to take tasks or telling me what to do)

For clarification, for the current version we can only use union to read data, make some queries and not to write. And there is no working solution to realize it now. Right?

benjie: Yes; there's polymorphic output types but no polymorphic input types. You can read more in the problem statement: https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md#-problem-statement Have a look through that document and see if you think there's anything you can add that would help move the conversation forward.

spec/Section 4 -- Introspection.md
```diff
@@ -331,7 +334,12 @@ Fields
 * `kind` must return `__TypeKind.INPUT_OBJECT`.
 * `name` must return a String.
 * `description` may return a String or {null}.
+* `requiresExactlyOneField` must return a Boolean
 * `inputFields`: a list of `InputValue`.
+  * if `requiresExactlyOneField` is {true}:
+    * for each {inputValue} in {inputFields}
+      * the {type} of {inputValue} must not be Non-Null
+      * the {defaultValue} of {inputValue} must be {null}
```
leebyron, line 342: I think this would be no defaultValue instead of `null` since you don't want to allow `field: Type = null`

spec/Section 4 -- Introspection.md
```diff
@@ -331,7 +334,12 @@ Fields
 * `kind` must return `__TypeKind.INPUT_OBJECT`.
 * `name` must return a String.
 * `description` may return a String or {null}.
+* `requiresExactlyOneField` must return a Boolean
 * `inputFields`: a list of `InputValue`.
+  * if `requiresExactlyOneField` is {true}:
+    * for each {inputValue} in {inputFields}
+      * the {type} of {inputValue} must not be Non-Null
```
leebyron, line 341: This answered my question on the call

spec/Section 4 -- Introspection.md
```diff
@@ -331,7 +334,12 @@ Fields
 * `kind` must return `__TypeKind.INPUT_OBJECT`.
 * `name` must return a String.
 * `description` may return a String or {null}.
+* `requiresExactlyOneField` must return a Boolean
 * `inputFields`: a list of `InputValue`.
+  * if `requiresExactlyOneField` is {true}:
```
leebyron, line 339: In other places we have "Type Validation" sub-sections. That would be a better place for this list set

leebyron (commented):
Adding some thoughts here after our meeting since you're revisiting this path

spec/Section 4 -- Introspection.md
benjie: Spot on :+1:

benjie:
Thanks Lee; I'll factor that feedback into the rewrite :+1:

benjie:
Rewritten (and resurrected) here: https://github.com/graphql/graphql-spec/pull/825

benjie:
For anyone following; `@oneOf` input objects are now part of the GraphQL Spec!

https://spec.graphql.org/September2025/#sec-OneOf-Input-Objects

:tada:

PR #598, opened 2019-07-23
Avoid parse ambiguity on types & extensions

leebyron:
Partial fix to #564, adds lookahead constraints to type system extensions to remove ambiguity or inefficiency from the grammar.

The GraphQL grammar should be parsed in linear type with at most one lookahead. Since extensions have an optional `{ }` body, finding the token `{` should always attempt to parse the relevant portion of the type extension rather than completing the type extension and attempting to parse the query shorthand SelectionSet.

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -222,7 +222,7 @@ SchemaDefinition : Description? schema Directives[Const]? { RootOperationTypeDef
 
 SchemaExtension :
   - extend schema Directives[Const]? { RootOperationTypeDefinition+ }
-  - extend schema Directives[Const]
+  - extend schema Directives[Const] [lookahead != `{`]
 
 RootOperationTypeDefinition : OperationType : NamedType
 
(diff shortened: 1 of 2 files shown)


leebyron:
> P.S. In retrospective, legacy SDL syntax that always requires curly braces (even if empty) offered a better solution to this problem

Yeah, I agree. I believe I made a mistake with that change for type definitions.

leebyron:
Updated this change to use a single token lookahead since that's more accurate to how we expect parser implementations to implement this restriction, and it reinforces the idea that parsing should be linear time.

Also includes both type extensions and type definitions, and updates the grammar appendix

leebyron:
Here's a parser test case:

```graphql
interface Foo

{ alpha: beta }
```

Before this change, this could be interpreted either as:

A. An incompletely defined interface Foo `interface Foo` followed by a query shorthand querying the field "beta" with the alias "alpha" `{ alpha: beta }`
B. An interface "Foo" with a single field "alpha" of type "beta" `interface Foo { alpha: beta }`

After this change, only B is possible. A is not possible since the incompletely defined interface must not be followed by `{`.

PR #600, opened 2019-07-29
Upgrade spec-md@1.1.0

leebyron:
https://github.com/leebyron/spec-md/releases/tag/v1.1.0


leebyron:
Nice new feature here is that arbitrary selections can be given stable links

PR #602, opened 2019-08-03
remove redundant step from SameResponseShape

ccbrown:
Unless I'm misinterpreting something, this step from `SameResponseShape` doesn't serve any purpose except to make me question myself.

If `typeA` or `typeB` is not a composite type, one of them is a scalar or enum, and the function would have returned in the previous step.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -432,7 +432,6 @@ SameResponseShape(fieldA, fieldB) :
   * If {typeA} or {typeB} is Scalar or Enum.
     * If {typeA} and {typeB} are the same type return true, otherwise return
       false.
-  * If {typeA} or {typeB} is not a composite type, return false.
   * Let {mergedSet} be the result of adding the selection set of {fieldA} and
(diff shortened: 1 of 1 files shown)


leebyron:
You're right, but I think it's still worth clarifying that point in the algorithm.

How about instead replacing this step with:

```
* Assert: {typeA} and {typeB} are both composite types.
```

spec/Section 5 -- Validation.md
```diff
@@ -432,7 +432,6 @@ SameResponseShape(fieldA, fieldB) :
   * If {typeA} or {typeB} is Scalar or Enum.
     * If {typeA} and {typeB} are the same type return true, otherwise return
       false.
-  * If {typeA} or {typeB} is not a composite type, return false.
```
leebyron: You're right, but I think it's still worth clarifying that point in the algorithm.

How about instead replacing this step with:

* Assert: {typeA} and {typeB} are both composite types.

ccbrown: Works for me. Updated.

Issue #300, opened 2017-04-28
Expose IDL directive information via introspection API


OlegIlyenko:
With growing popularity of IDL as a way to define a GraphQL schema, I think it would be quite beneficial to expose directive information via introspection API.

From what I can tell, the directive information is the only missing piece of information that is not obtainable via introspection API. For example in this schema definition:

```graphql
type User {
  id: ID!
  name: String
  admin: Boolean! @important
}

type Query {
  user: User
}
```

`@important` directive is only available at schema materialization time, but discarded as soon as schema is materialized from AST definitions.

One can see directives as a way to instruct server to construct the schema is specific way. But I would argue that directives have also a lot of potential as a way to expose additional meta-information about the schema. This may include things like: field cost/complexity (the use case I'm quite interested in), auth information, caching characteristics, and in general any kind of structured meta-information that is specific to a particular application or subset of applications. I think a lot of interesting use-cases and scenarios can emerge as soon as this meta-information is exposed via introspection. I can also imagine community working together on a set of common directives that may be adopted by client tooling and frameworks like apollo and relay. These common directives (exposed via introspection API) may provide deeper insights on the schema, relations between fields and objects, etc.

I described this feature in context of IDL, but in fact it's quite independent from it (though integrates with it very naturally).  I was thinking about different ways how this kind of user-defined meta-information can be exposed via introspection API and I feel that directive-based approach is the most natural and well integrated way of doing it.

I would love to hear you opinions on this topic!

calebmer:
> field cost/complexity (the use case I'm quite interested in)

I had some experiments on this that I didn’t release. I’d love to hear your thoughts 😊

* * *

I agree that being able to put arbitrary information into introspection is incredibly powerful, but I don’t think that we should be translating directives one-to-one into the introspection. Directives are meta instructions for the tools which consume the IDL. Making them a first class part of introspection reveals too many implementation details. It would also be very tough to type the directive arguments well.

I’d rather see tools that can interpret directives and then translate that to fields in the introspection 😊. For example, a server could:

```graphql
extend type __Field {
  important: Boolean
}
```

…and then no matter where you define your schema whether it be in the GraphQL IDL, GraphQL.js, or some other GraphQL server framework this flag can be set.

I don’t like the idea of making the IDL the one source of truth when creating a GraphQL server, but I do *really* like the idea of allowing users to extend the introspection types with arbitrary extra information.

smolinari:
[...]

Issue #236, opened 2016-11-01
Proposal: remove union types and allow interfaces to have no fields


calebmer:
I’m only semi-serious with this proposal, and I’m more interested to hear a defense of union types (because I currently can’t think of one) then to see this proposal proceed.

* * *

In the specification, it is currently written that [“An Interface type must define one or more fields.”](http://facebook.github.io/graphql/#sec-Interface-type-validation). From a precursory glance at the specification it seems like that limitation is arbitrary. In addition, if we remove that validation requirement (and perhaps change some other spec wordings) an interface could behave the exact same as a union, and perhaps be strictly better. An interface with no fields would be better than a union because the interface gives the schema designer the flexibility to add common fields in the future. Because interfaces are nominal and not structural this could be allowed.

A major benefit of unions in functional programming languages is that pattern matching statements with unions can be exhaustive. Whenever a programmer adds a new variant to the union, the type system breaks and requires all exhaustive pattern matches to be update appropriately. In GraphQL such a benefit is not realistic given that the code (mobile apps) which executes against a GraphQL schema can not always easily be updated when adding a new variant to a union. It would seem that the spec makes the argument that unions are good because:

> [Unions] also differ from interfaces in that Object types declare what interfaces they implement, but are not aware of what unions contain them. [(source)](They also differ from interfaces in that Object types declare what interfaces they implement, but are not aware of what unions contain them.)

However, it seems to me that the only case where the property of objects not declaring union membership is useful is in a pattern matching scenario which I think we can agree is a scenario that doesn’t always work well with the design goals of GraphQL.

Furthermore, if we make this change removing unions from the specification and allowing interface types with no fields, it wouldn’t be a breaking change. The access patterns from a GraphQL client that assumes unions exist would be the same for an interface with no fields.

So I think it’s plausible that we can remove unions from the spec and maintain backwards compatibility (I’d love to be proven wrong! That’s what this issue is for :wink:), and I also think it would have the following benefits:

- Simplifies the specification without breaking it, making it easier to learn for beginners and easier to build off of for implementors both on the client and server side. I’ve run into my fair share of Relay 1 union type bugs and gotchas. (https://github.com/facebook/relay/issues/1203, https://github.com/facebook/relay/issues/1230)
[...]

PR #606, opened 2019-08-05
[RFC] Support empty composite types

### LINKED ISSUE #490: Proposal: Do not require query as a root operation type
[The official spec](http://facebook.github.io/graphql/June2018/#sec-Root-Operation-Types) states:

> A schema defines the initial root operation type for each kind of operation it supports: query, mutation, and subscription; this determines the place in the type system where those operations begin.
The query root operation type must be provided and must be an Object type.
The mutation root operation type is optional; if it is not provided, the service does not support mutations. If it is provided, it must be an Object type.
Similarly, the subscription root operation type is also optional; if it is not provided, the service does not support subscriptions. If it is provided, it must be an Object type.

Why must the `query` root operation be provided? It seems to me that it should be perfectly legal to have an API that only contains a mutation or subscription. Especially with the advent of subscriptions, which are similar to queries in that they're only meant for retrieving data rather than modifying it, I'm not sure I see a purpose in requiring a query specifically. 

Some more details on my specific use case, to provide some more context: I'm employing [schema stitching](https://www.apollographql.com/docs/graphql-tools/schema-stitching.html) to merge numerous disjoint schemas together and present them as one. I have a subscription that I want to host in its own API, since doesn't logically belong in any of my other APIs; but the GraphQL implementation I'm using expects that a query will exist in the schema, forcing me to add an unnecessary query into my schema. Removing this stipulation from the spec (as well as from the implementation) would allow for this use case.

This stipulation is mentioned in #218, but in a slightly different context, and I think it warrants further discussion.

### LINKED ISSUE #568: The ability to represent empty objects
Based on this: https://github.com/graphql/graphql-js/issues/937#issuecomment-432857507

Currently we can't represent an empty object type:

```graphql
type Mutation {
  createUser(email: String!, password: String!): CreateUserPayload!
}

type UserWithEmailAlreadyExists
type WeakPassword
type AuthPayload {
  user: User!
  token: AccessToken!
}
union CreateUserPayload = UserWithEmailAlreadyExists | WeakPassword | AuthPayload
```

This fails with `graphql-js` with the following error:
```
Error: Type UserWithEmailAlreadyExists must define one or more fields.

Type WeakPassword must define one or more fields.
    at assertValidSchema (/Users/skainswo/dev/kumo/newer-world/api/node_modules/graphql/type/validate.js:71:11)
    at Object.validate (/Users/skainswo/dev/kumo/newer-world/api/node_modules/graphql/validation/validate.js:55:35)
    at doRunQuery (/Users/skainswo/dev/kumo/newer-world/api/node_modules/apollo-server-core/src/runQuery.ts:181:30)
    at /Users/skainswo/dev/kumo/newer-world/api/node_modules/apollo-server-core/src/runQuery.ts:80:39
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

The usecase is legit though.

There is an explicit check to disallow types with zero fields: https://github.com/graphql/graphql-js/blob/4116e2fc4fe36688f683258388f4a2d52076d199/src/type/validate.js#L273-L278

How about explicitly allowing in the spec? It's super useful for implementing algebraic types.

victorandree:
Supporting empty types, e.g. objects, input objects and interfaces without any fields, has concrete use cases.

- Support for GraphQL APIs without any `Query` operation type fields, for example, if the API only support mutations or subscriptions (see #490). This allows defining an empty `Query` object type, while still supporting introspection.
- Support for algebraic data types (see #568), where `__typename` is the only relevant field
- Potentially to support "well-known" (but empty) extensions of the introspection schema (see #300)

This is a minimalist spec change, which simply removes the relevant items under the type validation sub sections. It would probably be helpful to motivate the change and mention that one or more fields _should_ be present for a (typically) useful schema.

The requirement for composite types (object, input objects and interfaces) to define "one or more fields" was introduced in 059941486fcea9b93c5a156fe80df03d2021c0b4. This change references graphql/graphql-js#368, motivating the change with:

> Since GraphQL always requires you to select fields down to scalar values, an Object type without any defined fields cannot be accessed in any way in a query. This could be even more problematic for Input Objects where a required input object argument with no fields could result in a field that is impossible to query without producing an error.

With regards to these objections:

- It's always possible to select `__typename` and therefore even empty object types can be useful (as e.g. algebraic data types)
- Passing an empty input object appears syntactically valid:

    ```gql
    mutation {
      update(input: {}) {
        name
      }
    }
    ```

I think this proposal fulfills the guiding principles by enabling new capabilities motivated by real use cases. This change does not make any previously valid schema invalid, so largely preseves backwards compatibility.

Fixes #568 and fixes #490 at the specification level. I've forked graphql-js to remove the corresponding validations (see https://github.com/victorandree/graphql-js/tree/rfc/empty-objects).

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -96,6 +96,9 @@ type Query {
 }
 ```
 
+Note: A `query` root operation type which does not define any fields still
+allows for [Schema Introspection](#sec-Schema-Introspection).
+
 Similarly, the following mutation is valid if a `mutation` root operation type
 has a field named "setName". Note that the `query` and `mutation` root types
 must be different types.
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -792,8 +792,7 @@ Objects are never valid inputs.
 Object types have the potential to be invalid if incorrectly defined. This set
 of rules must be adhered to by every Object type in a GraphQL schema.
 
-1. An Object type must define one or more fields.
-2. For each field of an Object type:
+1. For each field of an Object type:
    1. The field must have a unique name within that Object type;
       no two fields may share the same name.
    2. The field must not have a name which begins with the
@@ -804,8 +803,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          characters {"__"} (two underscores).
       2. The argument must accept a type where {IsInputType(argumentType)}
          returns {true}.
-4. An object type may declare that it implements one or more unique interfaces.
-5. An object type must be a super-set of all interfaces it implements:
+2. An object type may declare that it implements one or more unique interfaces.
+3. An object type must be a super-set of all interfaces it implements:
    1. The object type must include a field of the same name for every field
       defined in an interface.
       1. The object field must be of a type which is equal to or a sub-type of
@@ -1052,8 +1051,7 @@ Interfaces are never valid inputs.
 
 Interface types have the potential to be invalid if incorrectly defined.
 
-1. An Interface type must define one or more fields.
-2. For each field of an Interface type:
+1. For each field of an Interface type:
    1. The field must have a unique name within that Interface type;
       no two fields may share the same name.
    2. The field must not have a name which begins with the
@@ -1410,8 +1408,7 @@ Literal Value            | Variables               | Coerced Value
 
 **Type Validation**
 
-1. An Input Object type must define one or more input fields.
-2. For each input field of an Input Object type:
+1. For each input field of an Input Object type:
    1. The input field must have a unique name within that Input Object type;
       no two input fields may share the same name.
    2. The input field must not have a name which begins with the
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1400,8 +1400,10 @@ Literal Value            | Variables               | Coerced Value
 `$var`                   | `{ var: "abc123" } }`   | Error: Incorrect value
 `{ a: "abc", b: "123" }` | `{}`                    | Error: Incorrect value for field {b}
 `{ a: "abc" }`           | `{}`                    | Error: Missing required field {b}
-`{ b: $var }`            | `{}`                    | Error: Missing required field {b}.
+`{ b: $var }`            | `{}`                    | Error: Missing required field {b}
+`{ }`                    | `{}`                    | Error: Missing required field {b}
 `$var`                   | `{ var: { a: "abc" } }` | Error: Missing required field {b}
+`$var`                   | `{ var: { } }`          | Error: Missing required field {b}
 `{ a: "abc", b: null }`  | `{}`                    | Error: {b} must be non-null.
 `{ b: $var }`            | `{ var: null }`         | Error: {b} must be non-null.
 `{ b: 123, c: "xyz" }`   | `{}`                    | Error: Unexpected field {c}
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -98,6 +98,8 @@ identify which actual type of the possible types has been returned.
 
 This field is implicit and does not appear in the fields list in any defined type.
 
+Note: The `__typename` field can always be queried, even if an Object or Interface does not define any fields.
+
 
 ## Schema Introspection
 


spec/Section 3 -- Type System.md
```diff
@@ -1052,8 +1051,7 @@ Interfaces are never valid inputs.
 
 Interface types have the potential to be invalid if incorrectly defined.
 
-1. An Interface type must define one or more fields.
```
benjie: Interesting discussion around this point specifically:

https://github.com/graphql/graphql-spec/issues/236

> This also drives the requirement that Interfaces must define at least one field - otherwise the Interface would not actually describe a contract to be fulfilled. This concept of contract-less interfaces is often called "marker interfaces" and in my opinion (hopefully not an uncommon opinion) are a bad practice borrowed from limited object-oriented languages. 
> -- @leebyron 

---

I'm personally in favour of field-less interfaces as a placeholder for adding shared fields later. E.g. you might say that 3 different types are `Searchable` and in future you may require that they have `name` and `url` in common so that you can display them in a search results page even if you don't have specific handling for some of the concrete types, but when you start designing your schema you might not know what those "Searchable" shared fields are/should be named yet.

I also think this can be useful for [interfaces that implement other interfaces](https://github.com/graphql/graphql-spec/pull/373):

```graphql
interface Searchable implements Node {}
```

e.g. to say that everything that's searchable is a Node, but not everything that's a Node is searchable.

victorandree:
I've pushed 3 commits to this PR. The first adds a note saying you can always query for `__typename` on an Object. The second adds empty input objects to the examples for input objects. The third adds a note about empty query root operation types (#490).

In my opinion, the spec doesn't have to motivate why these validations were removed, or make a judgment about the usability of empty types.

Feedback and discussion is kindly requested. I'm not sure what I can do to move this ahead, or if it could be considered ready for the next stage (I'd be happy to be the champion if so).

## Other solution: Interfaces must represent a contract (only empty Objects/Input Objects)

I can see the point in requiring interfaces to define one or more fields in a production ready, well-defined schema ([as linked](https://github.com/graphql/graphql-spec/pull/606#discussion_r312899229) by @benjie). An alternative approach would therefore be to only remove validation for Objects and Input Objects in this RFC.

Personally, I think enabling empty Object, Input Objects _and_ Interfaces in one change better aligns with the [Guiding Principles](https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md#guiding-principles) of "simplicity and consistency" while being worse from a "backwards compatibility" and "favor no change" perspective.

I think it would allow for more flexibility in evolving a schema from "placeholder" types. However, I would agree with @leebyron that "marker interfaces" are commonly used as "[metadata annotations for languages that don't have metadata annotations](https://github.com/graphql/graphql-spec/issues/236#issuecomment-257747315)" and that this is bad practice if you have a good system for metadata annotations. I think this demonstrates the need for an extensible introspection system, so that GraphQL _has_ a good system for metadata annotations.

victorandree:
@leebyron @benjie Is there something I can do to get this moving (or close it)?

IvanGoncharov:
@victorandree You can add it to the WG agenda:
https://github.com/graphql/graphql-wg/blob/master/agendas/2019-10-10.md#agenda
We will have it tomorrow would you be able to join?
If so you can add yourself and this proposal to the agenda (e.g. https://github.com/graphql/graphql-wg/pull/262)

spec/Section 3 -- Type System.md
leebyron: Based on the latest from the transitive interface implementation RFC

`interface Searchable implements Node {}`

would actually be an error, unless Node itself did not have a contract with any fields. `implements` instead of `extends` is intentional here, so `Searchable` would have to implement (declare) all fields `Node`'s contract requires.

leebyron:
I'm personally concerned about this approach. We were very intentional about this limitation to avoid problematic schema and type system anti-patterns (you've seen the past discussions about this). This may be the direction with the best solution, but we need to tread carefully to ensure such a change would be net-positive.

I'd love to see some exploration around alternative approaches to solving the problems outlined to support a strong case that this approach is the best one.


> Support for GraphQL APIs without any Query operation type fields, for example, if the API only support mutations or subscriptions (see #490). This allows defining an empty Query object type, while still supporting introspection.

I can envision some alternatives for this specific problem, like allowing not providing a Query type, similar to the existing behavior of allowing not providing a Mutation type. This might also be a bad idea, but it's worth exploring the tradeoffs with an empty Query type.

> Support for algebraic data types (see #568), where __typename is the only relevant field

This is certainly interesting, and probably the most compelling reason for empty Object types. In the past we've viewed this as an API anti-pattern, since it over-relies on type differentiation which can be a sub-par API response to interpret. However it's clear that Unions are becoming more broadly used, and the algebraic type pattern is particularly useful for mutations.

I still think this is probably an anti-pattern for other reasons, like error conditions for a mutation should probably come with data about how a client should behave (error message, persistent vs transient error, etc) that algebraic data types require the client to just have knowledge of ahead of time, but that's more of just my opinion that actually something we should design limitations for.

> Potentially to support "well-known" (but empty) extensions of the introspection schema (see #300)

Empty interfaces feels like the wrong solution to this problem. Most of the extensions / metadata want to belong to fields or arguments instead of types and they often contain more than just a named flag. 

Empty interfaces are also an explicit anti-pattern we intentionally excluded to avoid marker-interfaces which are less powerful than union types

leebyron:
Also something to consider is whether we should support not just empty definitions but empty queries?

Should you be able to write `{}` as a legal query? or `{ field {} }`? These were originally excluded because we thought it was more likely to be a mistake than an intentional decision.

benjie:
> I can envision some alternatives for this specific problem, like allowing not providing a Query type

How do you envision introspection working in this scenario ─ perhaps an introspection-only Query type is used if you don't provide your own Query? (If so: how do you rename this automatic Query type?) I think I'd prefer an explicit empty Query type be defined than implicitly adding one if I try to "opt-out" of Query.

> Also something to consider is whether we should support not just empty definitions but empty queries?

If we have:

```graphql
type MyType {}
type Query {
  field: MyType
}
```

then the query `{ field {} }` could (assuming no errors) return `{ "field": null }` or `{ "field": {} }`. There is a tiny bit of information here (e.g. you can tell that the MyType object _exists_ without forcing the backend to retrieve it). I think the slightly more verbose query `{ field { __typename } }` is preferable, it should be \~equivalent in terms of backend effort, and it helps to rule out accident.

In conclusion, I don't see any need to support empty queries, `{ __typename }` is valid against any GraphQL API/composite type.

victorandree:
I really appreciate the comments @leebyron!

Your comments highlight what problems an RFC should solve. I admit that this RFC is almost a solution looking for a problem given how the mentioned use cases are expressed. Unfortunately, out of the three, I'm personally most concerned about an approach to solving #300 – which this RFC isn't close to solving on its own.

Maybe each problem deserves its own treatment, instead of muddying them together like this? I did a write-up on each problem below with alternatives I've seen.

It's lengthy.

## Support schemas without a root query type (#490)

### Motivation

Not all GraphQL APIs need a `query` interface, but do fine with just `subscription` or `mutation`. This is especially apparent when writing GraphQL micro services. These APIs are forced to expose a dummy Query type because a "`query` root operation type must be provided" ([3.2.1 Root Operation Types](https://graphql.github.io/graphql-spec/June2018/#sec-Root-Operation-Types))

The current spec requires a `query` root operation type, because schema introspection happens "on the root of a query operation" ([4.5 Schema introspection](https://graphql.github.io/graphql-spec/June2018/#sec-Schema-Introspection)). Thus, there's a dependency on _a_ root query type for introspection to work.

### Solution: Allow the query root type to be empty

By not defining any own fields, the schema effectively doesn't expose a root query type, without breaking the introspection dependency. This approach maintains a large degree of backwards compatibility and only requires a small change.

It does introduce an inconsistency, however, because the other root operation types can be _not_ provided _or_ they can be empty. Which one is better?

### Alternative: Allow not providing a Query type

This approach puts the Query type on equal footing with other root operation types. This approach is more consistent in some ways since all root query types would be the same, but introspection would stick out more.

Without moving the schema introspection to its own root operation type, you'd still be able to query these implicit fields on the non-existent root query operation. These already _are_ implicit, so maybe it's not so bad. `__schema` and `__type` don't show up if you introspect the query operation type, so conceptually (and implementation wise) they just happen to have the same entry point.

This would require changing the `__Schema` introspection type to make `queryType` nullable.

## Support for algebraic data types (#568)

### Motivation

The original motivation is to support "empty variants" in union types, to differentiate between different kinds of errors (see #568). The original issue only discusses errors, but there may be other use cases.

### Solution: Allow empty objects differentiated by `__typename`

By allowing empty objects, the `__typename` field can still be used to differentiate between different empty variants. This enables a pattern of switching directly on the `__typename` for different error conditions.

This allows encoding variants inside GraphQL's type system directly. This adds a lot of flexibility and extensibility since you can add more properties to different variants _later_.

### Alternative: Use regular fields for switching, `null` for `Nothing`

With regards to errors specifically, you can use a `name` (or `type` or `code`) to differentiate between types of errors based on your application logic:

```graphql
type CreateUserError {
  "Machine-readable error name, such as `UserWithEmailAlreadyExists`"
  name: String!
}

union CreateUserPayload = CreateUserSuccess | CreateUserError
```

Since these errors all have the same structure from a GraphQL perspective, does it really make sense to differentiate them in GraphQL's type system? You can't do anything interesting with them anyway (you'd probably do a switch in your application code).

A potential downside is that the set of possible errors is not encoded in the union type. You could, however, still use an enum for `name` to signal that.

Finally, there shouldn't be a need for a `Maybe` type in GraphQL, since that's captured by simply allowing a field to be nullable.

## Extending the introspection schema (#300)

### Motivation

In a [comment on #300](https://github.com/graphql/graphql-spec/issues/300#issuecomment-504734306), I sketch some ideas for how extending the introspection schema could work (building on a lot of prior art).

In line with how `extensions` is already used in errors, I think it's desirable to namespace extensions of the `__Type` or `__Field` types rather than extending them directly.

### Solution: Empty placeholder types

If e.g. `__Type` is guaranteed to contain a field `extensions: __TypeExtensions` of type `type __TypeExtensions`, we would have a standard type to extend.

However, such a type should probably be empty unless there are extensions.

A well-known type is preferable to allowing users to provide their own `TypeExtensions` type if you have extensions from different sources (say for Apollo Federation extensions and for some authorization or validation library).

### Alternative: Use machine-readable descriptions

You can use field and type descriptions to put schema metadata in, using some standard to signal what the primary keys are on an object type. Here's an Apollo Federation directive simply moved to the description field.

```graphql
"""
@key(fields: "id")
"""
type User {
  "@external"
  id: ID!
  username: String
}
```

I wouldn't favor this, because I don't think descriptions are meant for computers.

### Alternative: Parsing the SDL

This is how Apollo Federation implements additional meta data for its federation types currently, since it's the only way to expose directives.

I don't think this is a good approach, because it relies on the SDL (so you can't programmatically construct these schemas) and I think it doesn't use directives "correctly" (see the lengthy discussion in #300).

victorandree:
In response to feedback at the latest WG meeting, I've decided to close this PR in favor of trying to solve the relevant problems instead of starting with a solution.

- I've opened a new PR to make the root query operation optional (see #631)
- I think using "marker types/interfaces" for algebraic data types is the wrong way to go because the types wouldn't be structurally different from the perspective of the GraphQL type system. Just put a `kind: ErrorKind` or `type: ErrorType` field to differentiate.
- The extensible introspection use-case is being worked on elsewhere and I don't want to derail that work. I hadn't read the [notes from the 2019-05-02 WG meeting](https://github.com/graphql/graphql-wg/blob/master/notes/2019-05-02.md) where @IvanGoncharov presented 2-stage introspection when I started this.

I think there could be a use case for making empty composite type validation _optional_ or a "warning", when you're still developing schemas or are preparing them for future extension.

alamothe:
It's pretty insane to see such a simple, widely useful removal of a completely arbitrary and artificial check in the GraphQL spec be rejected. 

>I think using "marker types/interfaces" for algebraic data types is the wrong way to go because the types wouldn't be structurally different from the perspective of the GraphQL type system. Just put a kind: ErrorKind or type: ErrorType field to differentiate.

What does the first part mean? Why does GraphQL needs structural difference?

The proposed solution is not OK, as some of the types could have fields and some would not. Consider:

```
type Mutation {
  login(username: String!, password: String!): LoginResponse!
}

union LoginResponse = WrongUsernamePassword | TwoFactorRequired

type WrongUsernamePassword = {}

type TwoFactorRequired = {
  sessionKey: String!
}
```

In this case, the only way to write type-safe code is to discriminate using `__typename`.

So, GraphQL spec forces us to put a `dummy` inside every empty type.

I am also left wondering which programming language does not allow for empty classes/types.

I will continue to use a patched GraphQL package (two lines of meaningless check removed) to allow empty types in my schemas.

victorandree:
@alamothe Note that the author (me, who is not officially affiliated with GraphQL) is the one abandoning this PR because I'm not interested in it anymore (for the reasons stated). I don't think this is "insane". If you're willing to champion it, feel free to pick it up – you may not be alone in thinking this is a good idea.

> What does the first part mean? Why does GraphQL needs structural difference?

That's a personal interpretation on my part about some of the design goals and the context in which GraphQL is used. I perceive the goals of the type system to be about covering the shape of the data. In practice, it is about describing the shape of a simple JSON object. Therefore, if two objects don't differ in structure, they should be the same type.

I'd make the case that a "good" error type in a schema should include a `message: String!` (for people) field. If you want to classify errors for computers, use a `code: ErrorCode!` field. If you want some errors to have additional properties, _then_ add a new type.\*

<sup>\* Actually, I don't think errors should be part of a GraphQL schema definition, but that you should use the standard [error mechanism](https://graphql.github.io/graphql-spec/draft/#sec-Errors)  for all errors. This is yet another personal opinion, but it does mean that I don't think errors are an interesting argument for ADTs.</sup>

alamothe:
Error messages should not come from the server, error codes should. In any case, ADTs go much beyond errors. Is wrong username/password in my example an error? I doubt anyone who does functional programming would agree.

Thank you for the time you put into this. It's a huge disappointment for me to see this being rejected. I pretty much lost faith in GraphQL at this point. I have no intention to champion it and waste my time with that as I don't see it going anywhere. I am happy that we can patch these libraries and fix obvious lack of foresight such as this.

akdor1154:
More support for this here - I am trying to represent ADTs in my schema, for domain objects, not errors. Working around this by including dummy variables which is just ridiculous.

rocketraman:
Came here from https://github.com/graphql/graphql-spec/issues/568 and just wanted to add a note of support.

A simple use case: I am modeling a financial concept "Cash on Hand", calculated based on a current cash position and a certain rate of revenue and expenditure. The result of this can basically be either a finite duration, or not applicable / infinity (if revenue less expenditures is positive).

One approach could be:

```
type CashOnHand {
  duration: Duration!
}

union Duration = FiniteDuration | InfiniteDuration

type FiniteDuration = {
  days: Int!
}

type InfiniteDuration = {}
```

There is a lot of pushback above to representing ADTs as errors, but this use case has nothing to do with errors -- it is not an error to have cash on hand be of infinite duration. And there is no point to adding any kind of dummy field to `InfiniteDuration` -- its confusing and unnecessary.

As far as I can tell, the best way to do this in GraphQL is to ignore the type system and use C-style "magic values" e.g. the client just has to know that `-1` means "infinite duration" / "not applicable". Unless I'm missing some idiomatic approach to solving this simple problem, this is a bit silly.

mathieuprog:
I keep wasting my time thinking like, now what insignificant dummy value should I add inside my `EmailAlreadyUsed` error? 🤦

PR #607, opened 2019-08-05
fix ExecuteField usage arguments

ccbrown:
*ExecuteField* is defined as:

> ExecuteField(objectType, objectValue, fieldType, fields, variableValues)


IvanGoncharov:
@ccbrown Thanks for PR 👍 Merged 🎉

PR #618, opened 2019-09-15
Add InputUnion problem statement

binaryseed:
This PR adds a section to the `InputUnion` RFC document that provides an overview of the problem being addressed.

Open to feedback!

@IvanGoncharov @leebyron @benjie

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -106,7 +106,7 @@ In addition, relying on this layer of abstraction means that this domain must be
 }
 ```
 
-Another common approach is to provide a unique mutation for every type. A schema employing this technique might have `logCatDropOff`, `logDogDropOff` and `logSnakeDropOff` mutations. This removes the potential for modeling non-sensical situations, but it explodes the number of mutations in a schema, making the schema less accessible.
+Another common approach is to provide a unique mutation for every type. A schema employing this technique might have `logCatDropOff`, `logDogDropOff` and `logSnakeDropOff` mutations. This removes the potential for modeling non-sensical situations, but it explodes the number of mutations in a schema, making the schema less accessible. If the type is nested inside other inputs, this approach simply isn't feasable.
 
 These workarounds only get worse at scale. Real world GraphQL schemas can have dozens if not hundreds of possible types for a single `Interface` or `Union`.
 
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -61,10 +61,10 @@ mutation {
 Unfortunately, this opens up a set of problems, since the Tagged union input type actually contains many fields, any of which could be submitted.
 
 ```graphql
-input AnimalDropOff {
-  cat: Cat
-  dog: Dog
-  snake: Snake
+input AnimalDropOffInput {
+  cat: CatInput
+  dog: DogInput
+  snake: SnakeInput
 }
 ```
 
@@ -118,7 +118,7 @@ mutation {
     location: "Portland, OR"
     animals: [
       {name: "Buster", age: 3, livesLeft: 7},
-      {name: "Ripple", age: 2, __typename: "Dog"},
+      {name: "Ripple", age: 2, __typename: "DogInput"},
     ]
   )
 }
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -18,9 +18,13 @@ To help bring this idea to reality, you can contribute through two channels:
 
 ## Problem Statement
 
-GraphQL currently provides abstract types that operate on **Object** types, but they aren't allowed on **Input** types.
+GraphQL currently provides polymorphic types that enable schema authors to model complex **Object** types that have multiple shapes while remaining type-safe, but lacks an equivilant capability for **Input** types.
 
-Over the years there have been numerous proposals from the community to add an abstract input type. Without an abstract type, users have had to resort to a handful of work-arounds to model their domains. These work-arounds have led to schemas that aren't as expressive as they could be, and schemas where some mutations that would be expected to reflect queries have instead been modeled differently by necessity.
+Over the years there have been numerous proposals from the community to add a polymorphic input type. Without such a type, schema authors have resorted to a handful of work-arounds to model their domains. These work-arounds have led to schemas that aren't as expressive as they could be, and schemas where mutations that ideally mirror queries are forced to be modeled differently.
+
+## Problem Sketch
+
+To understand the problem space a little more, we'll sketch out an example that explores a domain from the perspective of a Query and a Mutation. However, it's important to note that the problem is not limited to mutations, since `Input` types are used in field arguments for any GraphQL operation type.
 
 Let's imagine an animal shelter for our example. When querying for a list of the animals, it's easy to see how abstract types are useful - we can get data specific to the type of the animal easily.
 
@@ -106,7 +110,7 @@ Another common approach is to provide a unique mutation for every type. A schema
 
 These workarounds only get worse at scale. Real world GraphQL schemas can have dozens if not hundreds of possible types for a single `Interface` or `Union`.
 
-The goal of the **Input Union** is to bring an abstract type to inputs. This would enable us to model situations where an input may be of different types in a type-safe and elegant manner, like we can with outputs.
+The goal of the **Input Union** is to bring a polymorphic type to Inputs. This would enable us to model situations where an input may be of different types in a type-safe and elegant manner, like we can with outputs.
 
 ```graphql
 mutation {
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -20,7 +20,7 @@ To help bring this idea to reality, you can contribute through two channels:
 
 GraphQL currently provides abstract types that operate on **Object** types, but they aren't allowed on **Input** types.
 
-Over the years there have been numerous proposals from the community to add an abstract input type. Without an abstract type, users have had to resort to a handful of work-arounds to model their domains. These work-arounds lead to schemas that aren't as expressive as they could be, and even schemas that model data differently across queries & mutations.
+Over the years there have been numerous proposals from the community to add an abstract input type. Without an abstract type, users have had to resort to a handful of work-arounds to model their domains. These work-arounds have led to schemas that aren't as expressive as they could be, and schemas where some mutations that would be expected to reflect queries have instead been modeled differently by necessity.
 
 Let's imagine an animal shelter for our example. When querying for a list of the animals, it's easy to see how abstract types are useful - we can get data specific to the type of the animal easily.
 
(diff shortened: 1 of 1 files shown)


rfcs/InputUnion.md
```diff
@@ -18,8 +18,119 @@ To help bring this idea to reality, you can contribute through two channels:
...
+      ... on Dog { breed }
+      ... on Snake { venom }
+    }
+  }
+}
+```
+
+However, when we want to submit data, we can't use an `interface` or `union`, so we must model around that.
+
+One technique commonly used to is a **Tagged union** pattern. This essentially boils down to a "wrapper" input that isolates each type into it's own field. The field name takes on the convention of representing the type.
```
benjie:
```suggestion
One technique commonly used to is a **tagged union** pattern. This essentially boils down to a "wrapper" input that isolates each type into it's own field. The field name takes on the convention of representing the type.
```

rfcs/InputUnion.md
```diff
@@ -18,8 +18,119 @@ To help bring this idea to reality, you can contribute through two channels:
...
+
+```graphql
+input AnimalDropOff {
+  cat: Cat
+  dog: Dog
+  snake: Snake
+}
+```
+
+This leads to non-sensical mutations that are totally valid, like an animal that is both a `Cat` and a `Dog`
```
benjie:
```suggestion
This allows non-sensical mutations to pass GraphQL validation, for example representing an animal that is both a `Cat` and a `Dog`.
```

rfcs/InputUnion.md
```diff
@@ -18,8 +18,119 @@ To help bring this idea to reality, you can contribute through two channels:
...
+      {
+        cat: {name: "Buster", age: 3, livesLeft: 7},
+        dog: {name: "Ripple", age: 2, breed: WHIPPET}
+      }
+    ]
+  )
+}
+```
+
+In addition, relying on this layer of abstraction means that this domain is modeled differently across input & output. This puts a large burden on the developer interacting with the API, as they have to code against this.
```
benjie:
```suggestion
In addition, relying on this layer of abstraction means that this domain must be modelled differently across input & output. This can put a larger burden on the developer interacting with the schema, both in terms of lines of code and complexity.
```

rfcs/InputUnion.md
```diff
@@ -18,8 +18,119 @@ To help bring this idea to reality, you can contribute through two channels:
...
+// JSON structure submitted to a mutation
+{
+  "animals": [
+    {"cat": {"name": "Ruby", "age": 2, "livesLeft": 9}},
+    {"snake": {"name": "Monty", "age": 13, "venom": "POISON"}}
+  ]
+}
+```
+
+Another common approach is to provide a unique mutation for every type. So you'd have a `logCatDropOff` and `logDogDropOff` and `logSnakeDropOff`. That removes the potential for modeling non-sensical situations, but it explodes the number of mutations in an API.
```
benjie:
```suggestion
Another common approach is to provide a unique mutation for every type. A schema employing this technique might have `logCatDropOff`, `logDogDropOff` and `logSnakeDropOff` mutations. This removes the potential for modeling non-sensical situations, but it explodes the number of mutations in a schema, making the schema less accessible.
```

rfcs/InputUnion.md
```diff
@@ -18,8 +18,119 @@ To help bring this idea to reality, you can contribute through two channels:
...
+
+These workarounds only get worse at scale. Real world APIs can have dozens if not hundreds of possible types.
```
benjie: If you're referring to the schema, I think the numbers are much higher, so have edited accordingly. If you're referring to a single query/mutation field (e.g. the types returned from a union) then please adjust the text to make this clear :+1: 

```suggestion
These workarounds only get worse at scale. Real world GraphQL schemas can have hundreds if not thousands of possible types.
```

rfcs/InputUnion.md
```diff
@@ -18,8 +18,119 @@ To help bring this idea to reality, you can contribute through two channels:
...
+
+The idea of the **Input Union** is to bring an abstract type to inputs. This would enable us to model our domain as elegantly in input as we can in output.
```
benjie: I think the "would [...] as elegantly" claim might be a little strong; IMO none of the proposed solutions live up to this. I've softened it marginally:

```suggestion
The idea of the **Input Union** is to bring an abstract type to inputs. This would enable us to model situations where an input may be of different types in a type-safe and elegant manner, like we can with outputs.
```

rfcs/InputUnion.md
```diff
@@ -18,8 +18,119 @@ To help bring this idea to reality, you can contribute through two channels:
...
+inputunion AnimalDropOff = Dog | Cat | Snake
+
+type Mutation {
+  logAnimalDropOff(location: String, animals: [AnimalDropOff]): Boolean
+}
+```
+
+In this mutation, we encounter the main challenge of the **Input Union** - we need to determine the correct type of the data submitted. In a query, we have the `... on Cat` fragment which explicitly identifies the concrete type of the interface or union.
+
+A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions).
```
benjie:
```suggestion
A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions). In the example above we employed a `__typename` field to indicate the type of animal, but this is only one possible solution.
```

IvanGoncharov:
@binaryseed Thanks for keeping the ball rolling 👍 
But I don't agree that the core problem here is that:
> GraphQL currently provides abstract types that operate on **Object** types, but they aren't allowed on **Input** types.

GraphQL has two distinct type systems by design: nominal (for output type) and structural (for input types). But the current problem statement looks like core issue here is to switch input types to nominal and support abstract types. That definitely can be one of the solutions but it's not the single one, e.g. @benjie `@oneField` proposal doesn't add any abstract type and keeps input types nominal. 

I think we need to describe what we want to achieve not how we want to do that.
So I think core problem here is the ability to describe complex input types that can have multiple shapes (e.g. sets of fields) without losing type-safety so instead of `abstract type` we can use `polymorphic type`.

I also think the problem statement should be short and mutation example (with tag union workaround) should be part of `use cases` section because right now it looks like it's the only problem we trying to solve.

rfcs/InputUnion.md
```diff
@@ -18,8 +18,123 @@ To help bring this idea to reality, you can contribute through two channels:
...
+// JSON structure submitted to a mutation
+{
+  "animals": [
+    {"cat": {"name": "Ruby", "age": 2, "livesLeft": 9}},
+    {"snake": {"name": "Monty", "age": 13, "venom": "POISON"}}
+  ]
+}
+```
+
+Another common approach is to provide a unique mutation for every type. A schema employing this technique might have `logCatDropOff`, `logDogDropOff` and `logSnakeDropOff` mutations. This removes the potential for modeling non-sensical situations, but it explodes the number of mutations in a schema, making the schema less accessible.
```
jturkel: Is it worth calling out that the unique mutation per type approach often doesn't work if the input polymorphism is nested in a list or another input object e.g. the [dashboard example](https://gist.github.com/binaryseed/f2dd63d1a1406124be70c17e2e796891#dashboards), the [content widgets example](https://github.com/graphql/graphql-js/issues/207#issuecomment-308344371), etc.?

binaryseed: yes!

binaryseed: yes

rfcs/InputUnion.md
binaryseed: clarified this

rfcs/InputUnion.md
```diff
@@ -18,8 +18,123 @@ To help bring this idea to reality, you can contribute through two channels:
...
+mutation {
+  logAnimalDropOff(
+    location: "Portland, OR"
+    animals: [
+      {name: "Buster", age: 3, livesLeft: 7},
+      {name: "Ripple", age: 2, __typename: "DogInput"},
+    ]
+  )
+}
+```
```
IvanGoncharov: Using `__typename`, `__inputname` was one of the most controversial topics in the previous discussions so I don't think it worth having it in `Problem Sketch` section.
It can look like this instead:
--------------------
``` graphql
mutation {
  logAnimalDropOff(
    location: "Portland, OR"
    animals: [
      {name: "Buster", age: 3, livesLeft: 7},
      {name: "Ripple", age: 2},
    ]
  )
}
```
In this mutation, we encounter the main challenge of the **Input Union** - we need to determine the correct type of the data submitted.

A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions).

-----------------------

benjie: In my original editing I made this change, but it didn't feel sufficiently clear - I figured someone not familiar with the issue may not have detected the ambiguity. I then added comments to the end of each line `# This is meant to be CatInput` / `# This is meant to be DogInput` but it seemed messy and didn't really fix the issue. In the end I gave up and left it as the `__typename` solution.

However, I agree that if possible this should be addressed.

How about we add a comment `# Problem: the type of this Animal is ambiguous` to the second input?

```graphql
mutation {
  logAnimalDropOff(
    location: "Portland, OR"
    animals: [
      { name: "Buster", age: 3, livesLeft: 7 }
      { name: "Ripple", age: 2 } # Problem: the type of this Animal is ambiguous
    ]
  )
}
```

IvanGoncharov: @benjie Agree, comments works.
One note: type of both animals is ambiguous since `livesLeft` can become ambiguous if client extends `Dog` with `livesLeft` field or add new animal type with `livesLeft` in it. We can solve this by introducing some restrictions of having unique fields but it would part of solution and not problem definition.

So I would suggest:
```graphql
mutation {
  logAnimalDropOff(
    location: "Portland, OR",

    # Problem: we need to determinate type of the Animal for every item
    animals: [
      # This is meant to be CatInput
      { name: "Buster", age: 3, livesLeft: 7 }

      # This is meant to be DogInput
      { name: "Ripple", age: 2 }
    ]
  )
}
```

benjie: ^ Well written, @IvanGoncharov; I agree.

IvanGoncharov: @benjie @binaryseed I reviewed the entire document and everything looks great.
This is the last suggestion I have, so to clarify on final version:

--------------------
``` graphql
mutation {
  logAnimalDropOff(
    location: "Portland, OR",

    # Problem: we need to determinate type of the Animal for every item
    animals: [
      # This is meant to be CatInput
      { name: "Buster", age: 3, livesLeft: 7 },

      # This is meant to be DogInput
      { name: "Ripple", age: 2 },
    ]
  )
}
```
In this mutation, we encounter the main challenge of the **Input Union** - we need to determine the correct type of the data submitted.

A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions).

-----------------------
I think mutation query now is self-explaining so we can skeep SDL part and don't start a conversation on `inputunion` vs `inputUnion` vs `@oneField` vs all other possible SDL syntaxes on this stage of RFC process.

binaryseed: updated, great feedback!

IvanGoncharov:
Merged 🎉 
@binaryseed Thanks a lot for keeping the ball rolling 👍 
@benjie @jturkel Thanks for taking the time to review 👍

PR #620, opened 2019-09-28
Propose a consensus solution for Input Union

binaryseed:
This is a first go at proposing a solution to the Input Union.

@leebyron @IvanGoncharov @benjie

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -443,3 +443,39 @@ type Mutation {
   }
 }
 ```
+
+-----
+
+### Consensus Solution
+
+Basic steps:
+
+1) Attempt to read `__inputname`
+2) Attempt basic structural discrimination
+3) Fail
+
+Features:
+
+* Members of an Input Union must be Input Object Types
+  * Matches the behavior of the Union
+
+
+DiscriminateInputType(_providedArguments_, _inputUnion_):
+
+1. Let _possibleInputTypes_ be a list of _inputUnion_'s member Input Object Types
+1. Let _possibleInputNames_ be a list of the Names of _possibleInputTypes_
+1. If an Argument with the Name `__inputname` is a member of _providedArguments_:
+   1. Let _explicitInputName_ be the Value of `__inputname`
+   1. If _explicitInputName_ is a member of _possibleInputNames_:
+      1. Return _explicitInputName_
+   1. If _explicitInputName_ is not a member of _possibleInputNames_:
+      1. Return a query error
+1. Let _providedArgumentNames_ be the set of argument names in _providedArguments_
+1. Let _matchingTypes_ be an empty list
+1. For each _possibleInputType_ in _possibleInputTypes_:
+   1. Let _possibleInputTypeFieldNames_ be the set of field names in _possibleInputType_
+   1. If _providedArgumentNames_ is a subset of _possibleInputTypeFieldNames_:
+      1. Add _possibleInputType_ to _matchingTypes_
+1. If _matchingTypes_ is a list with a single item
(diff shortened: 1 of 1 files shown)


rfcs/InputUnion.md
```diff
@@ -443,3 +443,39 @@ type Mutation {
...
+
+Features:
+
+* Members of an Input Union must be Input Object Types
+  * Matches the behavior of the Union
+
+
+DiscriminateInputType(_providedArguments_, _inputUnion_):
+
+1. Let _possibleInputTypes_ be a list of _inputUnion_'s member Input Object Types
```
sungam3r: Always 1
изображение

binaryseed: How long did you spend circling all those ones? :)

Screen Shot 2019-09-27 at 11 01 20 PM

sungam3r: :)

IvanGoncharov:
@binaryseed Thanks for PR but we already discussed this and similar solutions during WG.
Problem with this one in particular: It's very easy to break old queries but adding new types to input union or by adding new fields to old types in a way that makes type detection ambiguous.

As I was telling during last two WGs, we are going in circles (for the last two years!) and discussing the same stuff all over again. So we need to work on RFC document that aggregates all the knowledge and all proposed solutions. 
For example, we can have section explaining that we can't break old queries by adding new fields or types.
So I still think we need to follow a plan and the next step would be to aggregate all use cases from different issues.

binaryseed:
This is that document, it already lists out many proposed solutions. I'm trying to get us to move forward (and stop going in circles)!

I submitted a PR that references use cases that the community have presented here: https://github.com/graphql/graphql-spec/pull/619

What do you expect will happen after we gather use cases? Do you think there is a solution that could exist that we just haven't thought of yet?

benjie:
I think it's worth adding this to the document with a status of rejected or similar and explanation why. That way it might discourage others from proposing it (and reduce the circles :wink:). This gives us at least one requirement that the final solution must meet:

1. Must not break previously valid operations when adding a new type to the input union.

I think it would be good to include an example demonstrating the breakage.

IvanGoncharov:
> I submitted a PR that references use cases that the community have presented here: #619

@binaryseed Sorry completely missed that. If I don't review PR for more than a couple of days please feel free to ping me in PR.

> What do you expect will happen after we gather use cases? Do you think there is a solution that could exist that we just haven't thought of yet?

@binaryseed Most importantly we didn't document why certain proposals are blocked.
Everyone (including me) was so focused on finding "ideal" solution we never bother to document why we opposed to certain proposals. Also since it so hot topic for the community it's impossible to have constructive discussion inside issues due to the amount of comments and the fact that majority of people don't bother to read the entire thread before posting so they are constantly repeating the same points.

I think we need to schedule separate "Input Union WG" meeting sometime after Oct WG and have a requirement that everyone reads RFC document (and included links) so we all on the same page.

To make this effective, my proposal would be to document all opposition to different proposals and don't accept any new objections during WG discussion unless a person is willing to submit them as PR for RFC document and same goes for all alternative solutions or ideas.

It could happen that even after a couple of rounds of discussion we will fail to find the "ideal" solution but in the process, we will evaluate all the proposal and document their pros/cons.
And personally I would accept non-"ideal" solution if I would see that we exhausted design space and every idea got a fair trial.

IvanGoncharov:
> This gives us at least one requirement that the final solution must meet:
>> Must not break previously valid operations when adding a new type to the input union.

@benjie Agree 👍

binaryseed:
Opened a PR to start a Requirements section: https://github.com/graphql/graphql-spec/pull/621

PR #621, opened 2019-10-01
Input Union: Begin requirements section

binaryseed:
Starting a section to gather requirements as we identify them

@leebyron @benjie @IvanGoncharov

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -123,7 +123,7 @@ mutation {
       {name: "Buster", age: 3, livesLeft: 7},
 
       # This is meant to be a DogInput
-      {name: "Ripple", age: 2},
+      {name: "Ripple", age: 2}
     ]
   )
 }
@@ -132,6 +132,10 @@ mutation {
 In this mutation, we encounter the main challenge of the **Input Union** - we need to determine the correct type of the data submitted.
 
 A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions).
+
+## Requirements
+
+1. **Backwards compatible** Adding a new member type to an Input Union must not break previously valid operations
 
 ## Use Cases
 


IvanGoncharov:
@binaryseed Thanks 👍

jturkel:
A few related things have come up in previous discussions that would be useful to align around being must haves, nice to haves or non-requirements:

1. Adding a new optional argument to an input union member type should not break previously valid operations
1. Converting an argument to an input union should not break previously valid operations if the previous argument type is a member type of the input union
1. Making an optional argument required on an input union member type should not break previously valid operations that either used other input member types or included the optional argument for the given input member type

All of them feel pretty important to me to facilitate continuous schema evolution so hopefully we can find a solution that satisfies all of them.

eapache:
2 seems like a really difficult constraint to satisfy given the strongly-typed way in which variables currently work.

jturkel:
🤔 In theory it *could* work if the variable's type was a one of the input union's member types right?

binaryseed:
> 1. Adding a new optional argument to an input union member type should not break previously valid operations

To clarify, Inputs don't have arguments, they have fields - do you mean "Adding a new optional field..."? This matches the behavior of Input types already, so I agree this should continue to be the case.

> 2. Converting an argument to an input union should not break previously valid operations if the previous argument type is a member type of the input union

A hard requirement for this would make it difficult to build a solution - it'd require structural discrimination, and we're finding it difficult to devise such a solution.

Additionally, on the object side, I don't think you can take an Object and turn it into a Union w/o a breaking change, so I'm not sure we need to require such a behavior for Input -> Input Union.

> 3. Making an optional argument required on an input union member type should not break previously valid operations that either used other input member types or included the optional argument for the given input member type

This is a complex requirement! Though when you look at the simple situation of making an optional argument required, I don't think there's any way around that being a breaking change regardless of wether or not the Input Union is involved...

benjie:
I've added a couple more requirements for non-breaking in https://github.com/graphql/graphql-spec/pull/622 ─ let me know there if you think these ones are contentious.

PR #622, opened 2019-10-02
Input Union: expand backwards-compatiblity requirements

benjie:
Following on from #621 and the discussion therein, here's a couple more requirements I think we're generally agreed on ─ standard non-breaking GraphQL behaviours.

@leebyron @binaryseed @IvanGoncharov

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -135,7 +135,7 @@ A wide variety of solutions have been explored by the community, and they are ou
 
 ## Requirements
 
-1. **Backwards compatible** ─ adding a new member type to an Input Union or doing any non-breaking change to existing member types must not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable must not break previously valid client operations.
+1. **Doesn't inhibit [schema evolution](https://graphql.github.io/graphql-spec/draft/#sec-Validation.Type-system-evolution)** ─ adding a new member type to an Input Union or doing any non-breaking change to existing member types must not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable must not break previously valid client operations.
 
 **Note**: Input Unions may enforce some restrictions on member types, but if they do so compliance with these restrictions must be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
 
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -135,10 +135,9 @@ A wide variety of solutions have been explored by the community, and they are ou
 
 ## Requirements
 
-1. **Backwards compatible** ─ the following actions must not break previously valid operations:
-    1. Adding a new member type to an Input Union
-    1. Adding a new optional field to a member Input Object type of an Input Union
-    1. Changing a field from non-nullable to nullable in a member Input Object type of an Input Union
+1. **Backwards compatible** ─ adding a new member type to an Input Union or doing any non-breaking change to existing member types must not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable must not break previously valid client operations.
+
+**Note**: Input Unions may enforce some restrictions on member types, but if they do so compliance with these restrictions must be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
 
 ## Use Cases
 
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -135,7 +135,10 @@ A wide variety of solutions have been explored by the community, and they are ou
 
 ## Requirements
 
-1. **Backwards compatible** Adding a new member type to an Input Union must not break previously valid operations
+1. **Backwards compatible** ─ the following actions must not break previously valid operations:
+    1. Adding a new member type to an Input Union
+    1. Adding a new optional field to a member Input Object type of an Input Union
+    1. Changing a field from non-nullable to nullable in a member Input Object type of an Input Union
 
 ## Use Cases
 


rfcs/InputUnion.md
```diff
@@ -135,7 +135,10 @@ A wide variety of solutions have been explored by the community, and they are ou
 
 ## Requirements
 
-1. **Backwards compatible** Adding a new member type to an Input Union must not break previously valid operations
+1. **Backwards compatible** ─ the following actions must not break previously valid operations:
+    1. Adding a new member type to an Input Union
+    1. Adding a new optional field to a member Input Object type of an Input Union
+    1. Changing a field from non-nullable to nullable in a member Input Object type of an Input Union
```
IvanGoncharov: There a few other non-breaking changes like:
1. Changing field to non-nullable but providing default value at the same time
2. Removing default value
And probably a few others I just wrote stuff from the top of my head.

So I would suggest to generalize it to:
> Adding a new member type to an Input Union or doing any change to existing member types (except for adding new required fields) must not result must not break previously valid client queries.

I would also add a note saying something like:
> Note: Input unions can enforce some restrictions on member types that should be fully validated during schema building.

Without this note, we will automatically reject all solutions with custom discriminator fields (since you can't add them in types themself) or structural proposals (based on having some restrictions on fields, e.g. having at least one unique required field).

In sense is very similar to limitations that Interface enforce of objects that implement it. You can add an interface to an existing object and it never breaks client queries, however, interface enforce some restrictions that validated during schema construction.

IvanGoncharov: Update: I forget that changing field type is also breaking except for switching from non-null type to nullable.
So a new proposal:
> Adding a new member type to an Input Union or doing any non-breaking change to existing member types must not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable must not break previously valid client queries.

eapache: Yeah, I think that's a good way to frame it: any change that is non-breaking on a regular input object must still be non-breaking if that input object is part of an input union.

benjie: Good points. I've edited these suggestions slightly and applied them to my PR here: https://github.com/graphql/graphql-spec/pull/622/commits/07970a2ce5e7d2a0458d1f065a8a51fe93441258

I think an important thing to note about interfaces is that the types themselves explicitly opt into the interface (`type Foo implements BarInterface { ... }`), whereas with unions member types can be added without any awareness within the type itself. Adding an output type to an output union does not require any preparation - it does not affect the future compatibility of the type that went into the union. This is not the case for input unions if they don't have similar rules, and perhaps if that's the case we should encourage them to be expressed as input interfaces instead (since the input type itself should opt into the constraint).

binaryseed:
Is it safe to say that the "Breaking changes" requirements are no different for an Input Union than other cases?

ie: Are there any Input Union specific breaking change requirements?

rfcs/InputUnion.md
jturkel: Is there a canonical list of breaking vs. non-breaking changes we should be referencing here for input member types? The spec mentions [type system evolution](https://graphql.github.io/graphql-spec/June2018/#sec-Validation) but doesn't go into much detail. The best list I could find was from [Apollo](https://www.apollographql.com/docs/graph-manager/schema-validation/#breaking-change-detection) but that doesn't include making a non-nullable input field nullable in the list of non-breaking changes so it doesn't seem complete.

benjie: I think Apollo's list of _breaking_ changes is pretty exhaustive, it covered everything that I could think of off the top of my head:

- Changing an input field or argument from nullable to non-nullable (previous operations may omit it, but now it's required -> breaking)
- Changing an output field from non-nullable to nullable (previous code needn't check for null, but now null is possible -> breaking)
- Changing the concrete type of an input field or argument (previous code may have stated the type explicitly in the variables (`$foo: Int`) and that will no longer validate)
- Removing fields from an object type/input object type
- Removing arguments from a field
- Removing fields from an interface (even if the underlying types still have them, fragments over the interface may still depend on it being present: `... on MyInterface { fieldWeRemoved }`)
- Removing an interface from a type (you can no longer spread on that interface within the type, or on that type within the interface)
- Removing a type from a `union` (you may spread a fragment of the type that was removed, the reason this would fail is because of a specified validation rule: https://graphql.github.io/graphql-spec/June2018/#sec-Fragment-spread-is-possible)
- Removing a value from an enum (it may be specified in a query)

I guess it's harder to do an exhaustive list of non-breaking changes?

IvanGoncharov: I think here is the most exhaustive list of breaking changes I saw so far:
https://github.com/graphql/graphql-js/blob/ed3ef8b4a75dfc642255715cd031d400825fc2ca/src/utilities/findBreakingChanges.js#L39-L55
This code is used by FB and many other companies in their CI so I'm pretty sure it's handling absolute majority of cases. (Note: `findDangerousChanges` is not used by FB and less used in general so I'm sure only in `findBreakingChanges`).

IvanGoncharov:
> I think an important thing to note about interfaces is that the types themselves explicitly opt into the interface (type Foo implements BarInterface { ... }), whereas with unions member types can be added without any awareness within the type itself. Adding an output type to an output union does not require any preparation - it does not affect the future compatibility of the type that went into the union. This is not the case for input unions if they don't have similar rules, and perhaps if that's the case we should encourage them to be expressed as input interfaces instead (since the input type itself should opt into the constraint).

@benjie Excellent point. I totally agree that it's preferable to either not restrict input types at all or to allow them to explicitly accept the required restriction.
However, I don't think it's a deal-breaker on its own but if any proposal doesn't fulfil this requirement it 100% should be noted in `Cons` section.

How about creating another section for such criteria?
For example, I think the ability to update from `input` type to `inputUnion` type without breaking changes is another example of such requirements.
Not sure how to call this section: `Optional requirements`, `optional design goals`?
I think it extremely important that this name reflects our goal achieve all of them but none of them should be considered a deal-breaker by itself.

rfcs/InputUnion.md
```diff
@@ -135,7 +135,9 @@ A wide variety of solutions have been explored by the community, and they are ou
 
 ## Requirements
 
-1. **Backwards compatible** Adding a new member type to an Input Union must not break previously valid operations
+1. **Backwards compatible** ─ adding a new member type to an Input Union or doing any non-breaking change to existing member types must not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable must not break previously valid client operations.
```
IvanGoncharov: Thinking more about this I think `Backwards compatible` can mean different stuff in this RFC, for example, proposed changes to introspection can be `Backwards compatible` or switching type from `input` to `inputUnion` can also be `Backwards compatible`.

How about renaming it to `No implicit breaking changes` or something similar?

benjie: Or “doesn’t inhibit schema evolution”?

IvanGoncharov: @benjie Sounds good 👍

benjie: Done

IvanGoncharov:
@benjie Merged 🎉

PR #625, opened 2019-10-10
InputUnion: Add performance considerations

spawnia:
As discussed in the working group meeting, i added this criteria.

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -133,10 +133,12 @@ In this mutation, we encounter the main challenge of the **Input Union** - we ne
 
 A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions).
 
-## Requirements
+## Criteria
+
+The following list of criteria allow us to compare the proposed solutions:
 
 1. **Doesn't inhibit [schema evolution](https://graphql.github.io/graphql-spec/draft/#sec-Validation.Type-system-evolution)** ─ adding a new member type to an Input Union or doing any non-breaking change to existing member types must not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable must not break previously valid client operations.
-1. **Does not increase payload size** ─ when issuing queries that use an Input Union, is the client required to send a larger amount of data to pass the arguments or variables?
+1. **Increase payload size** ─ when issuing queries that use an Input Union, is the client required to send a larger amount of data to pass the arguments or variables in comparison with a plain input type?
 
 **Note**: Input Unions may enforce some restrictions on member types, but if they do so compliance with these restrictions must be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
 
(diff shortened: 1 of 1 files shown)


eapache:
👍 we should rename the section to "Aims" while we're here. This is definitely an aim, even if it's not necessarily a hard requirement.

spawnia:
@benjie yes, the `__typename` argument is what i had in mind.

I think a useful metric is too look at the relative size increase, which may vary.
Hypothetical example: An `inputUnion` with hundreds of possible `input` types, that all have very long names and you are passing large inputs. That could easily end up doubling the request size.

Realistically, this should not be a concern for most if not at all real use cases. But as you noted, it is something we might consider.

benjie:
I think that's going to be true of any input union spec that we adopt (I think it's very unlikely that we'll end up with one that does purely structural discrimination as I think the trade-offs are too high). But sure, it's still a concern :+1: 

Could you remove the `Criteria` title from this change and simplify it down to just the payload size criterion? We can discuss the Criteria vs Aims in #626

binaryseed:
Do we have any other "size" considerations in the GraphQL spec? I'm unclear as to why this is really related to the topic of polymorphic types...

spawnia:
> Do we have any other "size" considerations in the GraphQL spec? I'm unclear as to why this is really related to the topic of polymorphic types...

Performance considerations might have influenced the current spec and it should be considered when deciding upon new designs.

binaryseed:
I appreciate the discussion, but I'm personally 👎 on this as a criteria. I agree that performance considerations matter, but I don't think the possibility of one additional field in each input union rises to the level of a performance concern. If optimizing payload size were a goal in GraphQL we wouldn't be sending an entire query string up with every HTTP request.

GraphQL's guiding principles already contains `Performance is a feature` so that already applies to any spec change we consider...

https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md#guiding-principles

What if we just put a reference to the Guiding Principles inside this section and note that they all apply?

spawnia:
We agreed in the working group to add all possible considerations into the document. Increasing payload size *does* impact performance, i don't follow how you could argue against that?

I generalized the section to a more brief criteria about performance and mentioned the tradeoffs between the different approaches. I think that explicitly naming the performance impacts of the possible solutions is useful.

We can make an informed decision about the performance impacts once we gathered all the criteria. If we don't even put it in the document, that is basically saying that this point is totally invalid and we don't even need to think about it.

binaryseed:
As an alternative to identifying one specific kind of performance consideration, I added a reference to _all_ the GraphQL Spec Guiding Principles in my PR:

https://github.com/graphql/graphql-spec/pull/626/files/36e1c34fb1846d080937c24b5679c29584ec74c6..a2c00574ea56b9f2956f10d96ba904a8726edb1d

eapache:
Is this stale, or should it be rebased/merged?

spawnia:
Superseded by #640

PR #626, opened 2019-10-14
Input Union RFC: Move Requirements to "Criteria"

binaryseed:
As discussed in the last Working Group meeting, this PR converts the "Requirements" section to an "Aims" section.

This is inline with the purpose of this document which is to gather all perspectives on the issue.

I've also added a few basic aims, one of which has an "objection"

@leebyron @IvanGoncharov @benjie

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -133,13 +133,13 @@ In this mutation, we encounter the main challenge of the **Input Union** - we ne
 
 A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions).
 
-## Solution Shape - Aims & Objections
+## Solution Criteria
 
 Hypothetical goals that a solution might attempt to fulfill.
 
 ### GraphQL should contain a polymorphic Input type
 
-The premise of this RFC - GraphQL should contain a polymorphic Input Union type.
+The premise of this RFC - GraphQL should contain a polymorphic Input type.
 
 ### Doesn't inhibit [schema evolution](https://graphql.github.io/graphql-spec/draft/#sec-Validation.Type-system-evolution)
 
@@ -149,15 +149,12 @@ Adding a new member type to an Input Union or doing any non-breaking change to e
 
 If a solution places any restrictions on member types, compliance with these restrictions should be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
 
-### A member type may be a Scalar
+### A member type may be a Leaf type
 
-In addition to containing Input types, an Input Union may also contain Scalars.
+In addition to containing Input types, member type may also contain Leaf types like `Scalar`s or `Enum`s.
 
-### A member type may be an Enum
-
-In addition to containing Input types, an Input Union may also contain `Scalar`s or `Enum`s.
-
-* Objection: An input union containing both a `String` and an `Enum` would be impossible to distinguish.
+* Objection: Multiple Leaf types serialize the same way, making it impossible to distinguish the type. For example, a `String`, `ID` and `Enum`.
+* Objection: Output polymorphism is restricted to Object types only. Supporting Leaf types in Input polymorphism would create a new inconsistency.
 
 ## Use Cases
 
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -135,7 +135,15 @@ A wide variety of solutions have been explored by the community, and they are ou
 
 ## Solution Criteria
 
-Hypothetical goals that a solution might attempt to fulfill.
+Hypothetical goals that a solution might attempt to fulfill. These goals will be evaluated with the [GraphQL Spec Guiding Principles](https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md#guiding-principles) in mind:
+
+* Backwards compatibility
+* Performance is a feature
+* Favor no change
+* Enable new capabilities motivated by real use cases
+* Simplicity and consistency over expressiveness and terseness
+* Preserve option value
+* Understandability is just as important as correctness
 
 ### GraphQL should contain a polymorphic Input type
 
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -143,7 +143,7 @@ The premise of this RFC - GraphQL should contain a polymorphic Input type.
 
 ### Input polymorphism matches output polymorphism
 
-Any data structure that can be modeled with output type polymorphism should be able to be modeled with Input polymorphism.
+Any data structure that can be modeled with output type polymorphism should be able to be mirrored with Input polymorphism. Minimal transformation of outputs should be required to send a data structure back as inputs.
 
 ### Doesn't inhibit [schema evolution](https://graphql.github.io/graphql-spec/draft/#sec-Validation.Type-system-evolution)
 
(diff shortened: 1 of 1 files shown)


rfcs/InputUnion.md
```diff
@@ -133,11 +133,31 @@ In this mutation, we encounter the main challenge of the **Input Union** - we ne
...
+
+### A member type may be a Scalar
+
+In addition to containing Input types, an Input Union may also contain Scalars.
+
+### A member type may be an Enum
+
+In addition to containing Input types, an Input Union may also contain `Scalar`s or `Enum`s.
+
+* Objection: An input union containing both a `String` and an `Enum` would be impossible to distinguish.
```
IvanGoncharov: I would merge `Scalar` and `Enum` into `Leaf type` especially since we use this term through GraphQL spec so it's not something new.
As for objection it also applied to scalar types since `ID` can intersect with `String`/`Int` and `Float` can intersect with `Int`. 
Also, I don't think it's objection since it just requires boxing leaf values e.g. `@oneOf` solves this issues `{ string: 'foo' }`, `{ enum: foo }`.

rfcs/InputUnion.md
```diff
@@ -3,9 +3,9 @@ RFC: GraphQL Input Union
 
 The addition of an Input Union type has been discussed in the GraphQL community for many years now. The value of this feature has largely been agreed upon, but the implementation has not.
 
-This document attempts to bring together all the various solutions that have been discussed with the goal of reaching a shared understanding of the problem space.
+This document attempts to bring together all the various solutions and perspectives that have been discussed with the goal of reaching a shared understanding of the problem space.
 
-From that shared understanding, this document can then evolve into a proposed solution.
+From that shared understanding, the GraphQL Working Group aims to reach a concensus on how to address the proposal.
```
benjie:
```suggestion
From that shared understanding, the GraphQL Working Group aims to reach a consensus on how to address the proposal.
```

rfcs/InputUnion.md
```diff
@@ -133,11 +133,31 @@ In this mutation, we encounter the main challenge of the **Input Union** - we ne
...
 A wide variety of solutions have been explored by the community, and they are outlined in detail in this document under [Possible Solutions](#Possible-Solutions).
 
-## Requirements
+## Solution Shape - Aims & Objections
 
-1. **Doesn't inhibit [schema evolution](https://graphql.github.io/graphql-spec/draft/#sec-Validation.Type-system-evolution)** ─ adding a new member type to an Input Union or doing any non-breaking change to existing member types must not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable must not break previously valid client operations.
+Hypothetical goals that a solution might attempt to fulfill.
 
-**Note**: Input Unions may enforce some restrictions on member types, but if they do so compliance with these restrictions must be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
+### GraphQL should contain an Input Union type
```
benjie: I think that `Input Union` is a charged term and actually shapes how people think the solution should be shaped. Personally I prefer the more neutral "polymorphic input" term, since this is actually what we're after, and it means that we do not have to mirror the output types opening more potential solutions (with potentially better trade-offs).

```suggestion
### GraphQL should contain a polymorphic Input type
```

rfcs/InputUnion.md
benjie: The spec isn't explicit on this, but the term "leaf" may also be interpretted include lists of scalars and enums since these also do not have selection sets. I think we want to explicitly preclude lists from this particular title. If lists in the input union (as opposed to lists OF the input union) are desired I'd suggest adding another header for that.

spawnia:
I propose to name the section **Criteria** as that is a single, more general term that covers positive and negative aspects (and everything in between).

https://github.com/graphql/graphql-spec/pull/625

rfcs/InputUnion.md
spawnia: I agree with @IvanGoncharov, in the space of all solutions, there are definitely some that allow the inclusion of leaf types.

eapache: A stronger objection IMO (or at least a different objection) is that output polymorphism (Union and Interface types) don't support leaf types/scalars/etc, and if we want to support that in polymorphism generally we should do it for both.

benjie: (@eapache I agree, and `@oneField` can be extended to do this very easily. I've left it out for now to keep the proposal small and focussed.)

binaryseed: merged those two & added that additional objection

rfcs/InputUnion.md
```diff
@@ -133,11 +133,28 @@ In this mutation, we encounter the main challenge of the **Input Union** - we ne
...
+### Any member type restrictions are validated in schema
+
+If a solution places any restrictions on member types, compliance with these restrictions should be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
+
+### A member type may be a Leaf type
+
+In addition to containing Input types, member type may also contain Leaf types like `Scalar`s or `Enum`s.
+
+* Objection: Multiple Leaf types serialize the same way, making it impossible to distinguish the type. For example, a `String`, `ID` and `Enum`.
+* Objection: Output polymorphism is restricted to Object types only. Supporting Leaf types in Input polymorphism would create a new inconsistency.
```
jturkel: Should the core of this objection also be a solution criteria i.e. consistency between input polymorphism and output polymorphism is desirable everything else being equal?

binaryseed: Sounds good, added

rfcs/InputUnion.md
```diff
@@ -133,11 +133,32 @@ In this mutation, we encounter the main challenge of the **Input Union** - we ne
...
+Hypothetical goals that a solution might attempt to fulfill.
 
-**Note**: Input Unions may enforce some restrictions on member types, but if they do so compliance with these restrictions must be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
+### GraphQL should contain a polymorphic Input type
+
+The premise of this RFC - GraphQL should contain a polymorphic Input type.
+
+### Input polymorphism matches output polymorphism
+
+Any data structure that can be modeled with output type polymorphism should be able to be modeled with Input polymorphism.
```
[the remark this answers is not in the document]
binaryseed: Good feedback, I updated this statement to be stronger & added the second note

rfcs/InputUnion.md
```diff
@@ -133,11 +133,40 @@ In this mutation, we encounter the main challenge of the **Input Union** - we ne
...
+* Preserve option value
+* Understandability is just as important as correctness
+
+### GraphQL should contain a polymorphic Input type
+
+The premise of this RFC - GraphQL should contain a polymorphic Input type.
+
+### Input polymorphism matches output polymorphism
+
+Any data structure that can be modeled with output type polymorphism should be able to be mirrored with Input polymorphism. Minimal transformation of outputs should be required to send a data structure back as inputs.
```
IvanGoncharov: @binaryseed I personally object to this one since GraphQL was never about CRUD.
Moreover, it's perfectly normal to have two distinct type systems, for example, most the programming languages have classes (support polymorphism) and structs (doesn't support polymorphism at the same time). To not block I will try to summarise my objection into `Objection` note and submit as a separate PR.

IvanGoncharov:
@binaryseed Sorry for delay and thanks for reaching out on Slack 👍

PR #630, opened 2019-10-22
Input Union RFC: add more solution criteria

benjie:
I've pulled in more of the criteria/objections from the various Input Unions issues/PRs.


rfcs/InputUnion.md
```diff
@@ -160,9 +162,48 @@ If a solution places any restrictions on member types, compliance with these res
 
 In addition to containing Input types, member type may also contain Leaf types like `Scalar`s or `Enum`s.
 
-* Objection: Multiple Leaf types serialize the same way, making it impossible to distinguish the type. For example, a `String`, `ID` and `Enum`.
+* Objection: Multiple Leaf types serialize the same way, making it impossible to distinguish the type without additional information. For example, a `String`, `ID` and `Enum`.
```
benjie: (Minor clarification - boxing the value or putting it in a tagged union would solve this issue so it's not always impossible.)

rfcs/InputUnion.md
```diff
@@ -160,9 +162,48 @@ If a solution places any restrictions on member types, compliance with these res
...
+For consistency with output unions/interfaces each union should resolve to a single JSON object; i.e. `{__typename:"MyInput",value:3}` rather than `{MyInput:{value:3}}`.
+
+* Objection: wrapper objects can be more concise than including `__typename` or similar in each value.
+* Objection: the "tagged union" pattern is already in common usage.
+* Objection: this pattern enables support for leaf types (e.g. scalars) in unions.
+* Objection: this pattern enables support for individual input object list types in unions.
+
+### Input unions should be easy to upgrade from existing solutions
+
+Many people in the wild are solving the need for input unions with validation at run-time (e.g. using the "tagged union" pattern). Formalising support for these existing patterns in a non-breaking way would enable existing schemas to become retroactively more type-safe.
```
jturkel: Do you have a list of patterns currently being used in the wild and some notion of how popular they are? We (Salsify) opted to go with a discriminator field in an input type that unions fields for all member types (with tooling support to auto-generate the "union" input type + implement server side validation) rather than the "tagged union" pattern.

benjie: No; that would be a really valuable resource! I'm just commenting based on what I've read from the issues in this repo. @IvanGoncharov has a great list of public GraphQL APIs here: https://github.com/APIs-guru/graphql-apis perhaps we can extract some of these common patterns from there?

rfcs/InputUnion.md
```diff
@@ -148,6 +148,8 @@ The premise of this RFC - GraphQL should contain a polymorphic Input type.
 
 Any data structure that can be modeled with output type polymorphism should be able to be mirrored with Input polymorphism. Minimal transformation of outputs should be required to send a data structure back as inputs.
 
+* Objection: input types and output types are distinct. Output types support aliases and arguments whereas input types do not. Marking an output field as non-nullable is a non-breaking change, but marking an input field as non-nullable is a breaking change.
```
binaryseed: "input types and output types are distinct".. the criteria here is more about the fact that there are data structures that you can model in output types that are impossible to model with input types.

It seems to me that the fact that input and output types are distinct and have different behaviors doesn't really address the intention of this criteria

rfcs/InputUnion.md
```diff
@@ -160,9 +162,48 @@ If a solution places any restrictions on member types, compliance with these res
...
 * Objection: Output polymorphism is restricted to Object types only. Supporting Leaf types in Input polymorphism would create a new inconsistency.
 
+### Changing field from an input type to an input union including that type is non-breaking
+
+Since the input object type is now a member of the input union, existing input objects being sent through should remain valid.
+
+* Objection: achieving this by indicating the default in the union (either explicitly or implicitly via the order) is undesirable as it may require multiple equivalent unions being created where only the default differs.
+* Objection: achieving this by indicating a default type in the input field is verbose/potentially ugly.
+
+### Input unions may include other input unions
```
binaryseed: Not sure I understand this one - what does it mean to "include" another input union?

rfcs/InputUnion.md
benjie: This item currently uses the terms "matches" and "mirrors" indicating that input and output polymorphism should look and feel nearly identical. This objection raises the point that input and output objects don't need to be similar in order to achieve the goal of enabling modelling polymorphic inputs.

rfcs/InputUnion.md
benjie:
```
inputUnion U1 = A | B
imputUnion U2 = U1 | C
```

Here U2 includes U1 (it does not need to explicitly list `A` and `B` as members, making maintenance easier).

benjie:
Sorry for the delayed responses, I was at a conference last week.

PR #636, opened 2019-11-06
InputUnion: note a possible inconsistency between solution criteria.

IdeaHunter:
There is some inconsistency may be seen between criteria that can result in producing solution which is not 100% compatible with existing aspects of output unions:
Sometimes there is a case when you cant have a priory knowledge what input data some of specific union should have, and output union have `__resolveType` to solve this problem. IMHO the whole point of union is to describe a situation when you do not have exact knowledge of type until you have it and when you can only describe what it can be. Please correct me if im wrong on this and `__resolveType` have different purpose.

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -156,6 +156,8 @@ Adding a new member type to an Input Union or doing any non-breaking change to e
 
 If a solution places any restrictions on member types, compliance with these restrictions should be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
 
+However this restriction may be in conflict with criteria `Input polymorphism matches output polymorphism` because for describing output union one can use `__resolveType` in order to resolve which exactly type this union member resolves to.
+
 ### A member type may be a Leaf type
 
 In addition to containing Input types, member type may also contain Leaf types like `Scalar`s or `Enum`s.


benjie:
As a point of clarification: many of the criteria are in conflict with each other. I don't believe there's a solution that matches all of them. The idea is that these are checks we can apply against the various solutions so we can easily analyse which goals they achieve and which they fail; we can then use this information to help us decide which is the best candidate going forward.

benjie:
I think your input here might be in the wrong section. This section says:

```md
### Any member type restrictions are validated in schema

If a solution places any restrictions on member types, compliance with these restrictions should be fully validated during schema building (analagous to how interfaces enforce restrictions on member types).
```

This is referring to restrictions analogous to the restrictions that an interface places on a type; e.g.

```graphql
interface MyInterface {
  myField: Int
}

# THIS IS NOT VALID
type MyType implements MyInterface {
  myOtherField: String
}
```

^ This is not valid beucase there's a validation step during schema building that ensures that all types that implement interfaces actually do implement the interfaces. In this case `MyType` does not have a `myField: Int` field, so it does not pass this build-time validation. It is unrelated to `__resolveType`.

IvanGoncharov:
@IdeaHunter Problem with `__resolveType` is we are losing the ability to fully validate queries and query variables on client-side, for example in GraphiQL, `eslint-graphql` and similar tools.
Also, it will prevent us from generating client-side typings that prevent invalid values to be passed.

IdeaHunter:
@IvanGoncharov oh, i see, i was thinking different way. 
The `__resolveType` can be called before actual validation happened, because it can tell which type in gql schema framework should validate agains to. There is a reason why `__resolveType` is in output union and i believe same should be applied to requirement for input unions

IdeaHunter:
@benjie 
> As a point of clarification: many of the criteria are in conflict with each other.
At first my PR was meant to provide a possible solution, but then i decided to validate solution against criteria given for solution and have found that criteria itself isnt in agreement with it own.

>I think your input here might be in the wrong section
Good point! It was completely misread that one.

Closing this PR, because my comment to document wasn't relevant.

PR #637, opened 2019-11-07
Update copyright and URLs from Facebook to GraphQL Foundation

leebyron:
Long overdue update! This changes all links in the spec to the new github org links to avoid the redirect, and updates the spec history and copyright to reflect the 2019 ownership transfer to the GraphQL Foundation


leebyron:
This doesn't touch the section about the OWFa signers, which we'll need to update before cutting the next version.

PR #640, opened 2019-11-08
Add performance criteria to input unions

eapache:
@spawnia @IvanGoncharov


spawnia:
That captures my original intent quite well, thank you.

binaryseed:
I still don't think any of the solutions present any computational complexity that will have a measurable impact, but we might as well note it

PR #642, opened 2019-11-10
Input Union: Remove incorrect objection

binaryseed:
In this PR, I'd like to make the case that this objection is not correct and should be removed (or replaced with clarified objections)

> Objection: input types and output types are distinct. Output types support aliases and arguments whereas input types do not. Marking an output field as non-nullable is a non-breaking change, but marking an input field as non-nullable is a breaking change.

1) "Input" and "output" have differences but are not distinct. An "input" can be an Input Object, Enum, Scalar, List. An "output" can be a Object, Enum, Scalar, List. Both are structures that contain fields. What is different is that arguments and input fields must be Input Objects.

1) Strictly speaking, output types do _not_ support aliases or arguments. Fields support aliases and arguments. A field has a name which can be aliased. Fields may have arguments & an "output type" - the type itself does not have arguments.

1) Types themselves do not have a notion of nullability, only fields & arguments do. Being able to represent a specific data structure with types is orthogonal to the important ability for a schema to evolve in a non-breaking manner. Even then, there are a whole host of changes that are already possible in GraphQL that will be a breaking change, so that isn't a reason not to support input polymorphism.

-------

I'm open to ideas about how to capture this kind of debate in the document vs just deleting the objection...

@IvanGoncharov @leebyron @benjie @jturkel @eapache

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -148,8 +148,6 @@ The premise of this RFC - GraphQL should contain a polymorphic Input type.
 
 Any data structure that can be modeled with output type polymorphism should be able to be mirrored with Input polymorphism. Minimal transformation of outputs should be required to send a data structure back as inputs.
 
-* Objection: input types and output types are distinct. Output types support aliases and arguments whereas input types do not. Marking an output field as non-nullable is a non-breaking change, but marking an input field as non-nullable is a breaking change.
-
 ### Doesn't inhibit [schema evolution](https://graphql.github.io/graphql-spec/draft/#sec-Validation.Type-system-evolution)
 
 Adding a new member type to an Input Union or doing any non-breaking change to existing member types does not result in breaking change. For example, adding a new optional field to member type or changing a field from non-nullable to nullable does not break previously valid client operations.


eapache:
> "Input" and "output" have differences but are not distinct.

I dunno, to me "having differences" is what "distinct" means.

jturkel:
Small process suggestion: This objection has come up periodically during the input union discussion so should we capture any rationale for making it a non-objective in the RFC document to avoid having the discussion again in the future?

It does seem fair to callout out the input and output types are distinct which could in theory lead to different models of supporting polymorphism between input and output. That said, IMO we should strive to minimize the differences between input and output type systems to make it easier for developers to ramp up on GraphQL. Everything else being equal, it's much easier to learn one polymorphism scheme than two.

I think the bigger issue that the objection doesn't call out is the difference in how references to associated entities are handled in output (objects) vs. input (foreign keys). Most GraphQL schemas I've seen do something like this which requires a transformation to unnest foreign keys between output and input e.g.

```graphql
type Cat {
  name: String!
  owner: Person
}

input CatInput {
  name: String!
  ownerId: ID
}
```
I guess you could make the argument that if you're already doing a transformation for ids, then doing additional transformations to support input unions isn't a big deal but I still think fewer transformations is better.

benjie:
I've fixed the semantics issues you raised here: https://github.com/graphql/graphql-spec/pull/645

Google defines `distinct` as `recognizably different in nature from something else of a similar type` and I stand by that assertion - _composite_ input types and _composite_ output types are distinct. If they weren't then we wouldn't need a distinction between them.

PR #650, opened 2019-11-21
Evaluate solutions against criteria

spawnia:
This lays out a structure to evaluate each of the solutions against each of the criteria.

I intentionally did not fill out the evaluations yet, as that would blow up the size and need for discussion of this PR. Instead, i propose to do them one-by-one in separate PRs.

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -260,7 +260,7 @@ the solution does avoid the negative aspact, it gets a o
 |---|---|---|---|---|---|
 | A |   |   |   |   |   |
 | B |   |   |   |   | x |
-| C |   |   |   | x |   |
+| C |   |   | x | x |   |
 | D |   |   |   |   |   |
 | E |   |   |   |   |   |
 | F |   |   |   |   |   |
@@ -510,6 +510,53 @@ type Mutation {
 ##### [Input polymorphism matches output polymorphism](#b-input-polymorphism-matches-output-polymorphism)
 
 ##### [Doesn't inhibit schema evolution](#c-doesnt-inhibit-schema-evolution)
+
+Adding a nullable field to an input object could change the detected type of
+fields or arguments in pre-existing operations.
+
+Assuming we have this schema:
+
+```graphql
+input InputA {
+  foo: Int
+}
+input InputB {
+  foo: Int
+  bar: Int
+}
+input InputC {
+  foo: Int
+  bar: Int
+  baz: Int
+}
+inputUnion InputU = InputA | InputB | InputC
+
+type A {...}
+type B {...}
+type C {...}
+union U = A | B | C
+
+type Query {
+  field(u: InputU): U
+}
+```
+
+Given this query:
+
+```graphql
+query ($u: InputU = {foo: 3, baz: 3}) {
+  field(u: $u) {
+    __typename
+    ... on C {
+      # ...
+    }
+  }
+}
+```
+
+the default value of `$u` would be detected as type InputC. However adding the
+field `baz: Int` to type InputA would result in the same value now being detected
+as type InputA, which could be a breaking change for application behaviour.
 
 ##### [Any member type restrictions are validated in schema](#d-any-member-type-restrictions-are-validated-in-schema)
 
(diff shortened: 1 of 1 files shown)


rfcs/InputUnion.md
```diff
@@ -327,11 +339,7 @@ type Mutation {
 }
 ```
 
-##### Problems:
-
-* The discriminator field is non-sensical if the input is used _outside_ of an input union, or in multiple input unions.
```
binaryseed: Looks like this PR is removing all the `Problems` sections. I'd prefer we didn't delete them now, so that we can incorporate them into the evaluations...

rfcs/InputUnion.md
```diff
@@ -288,11 +284,27 @@ type Mutation {
...
 
-##### Variations:
+#### Variations:
 
 * A `default` type may be defined, for which specifying the `__typename` is not required. This enables a field to migration from an `Input` to an `Input Union`
 
-#### Single user-chosen field; value is the `type`
+#### Evaluation
+
+- A:
```
binaryseed: Perhaps the letter could be a link to the solution...

spawnia: I decided to include the title and link it. Switching back and forth makes it really hard to navigate.

rfcs/InputUnion.md
spawnia: I included them into the evaluation.

spawnia:
> Would it make sense to have a summary table with solutions as rows and evaluation criteria as columns so we can easily see how all of the solutions score against all of the evaluation criteria? The current structure requires lots of jumping around the document to figure out that summary information.

I considered that, but figured that would be problematic if lengthy explanations are required.

We might still add that in later, i agree it could be very useful to get an overview. What scale or what kinds of score could we use? I am thinking something simple, like: :smiley:, :cry: and :neutral_face: ?

binaryseed:
I was thinking something that stands out a bit more, with colors: 

🚫 ⚠️ ✅

benjie:
Adding in the evaluations in future PRs makes sense, but let's add in one of the
negatives of solution 3 before merging so that the community don't mistake it for a clear winner with no drawbacks.

For example, we could add:

```md
##### [Doesn't inhibit schema evolution](#c-doesnt-inhibit-schema-evolution)

Adding a nullable field to an input object could change the detected type of
fields or arguments in pre-existing operations.
```

If it needs more explanation, here's an example:

````md
Assuming we have this schema:

```graphql
input InputA {
  foo: Int
}
input InputB {
  foo: Int
  bar: Int
}
input InputC {
  foo: Int
  bar: Int
  baz: Int
}
inputUnion InputU = InputA | InputB | InputC

type A {...}
type B {...}
type C {...}
union U = A | B | C

type Query {
  field(u: InputU): U
}
```

Given this query:

```graphql
query ($u: InputU = {foo: 3, baz: 3}) {
  field(u: $u) {
    __typename
    ... on C {
      # ...
    }
  }
}
```

the default value of `$u` would be detected as type InputC. However adding the
field `baz: Int` to type InputA would result in the same value now being detected
as type InputA, which could be a breaking change for application behaviour.
````

**Comment by spawnia** [comment]
@binaryseed i tried the emoji's first, but soon realized that copy-and-pasting them around is stressful, so i went for a more boring:

- `o` Positive
- `-` Neutral/Not applicable
- `x` Negative

**Comment by binaryseed** [comment]
@spawnia ready for me to merge?

**Comment by binaryseed** [comment]
So, I'm thinking the lists of all the criteria repeated over and over for each solution is a bit overwhelming... There's just so much repeated text that I worry the document is getting hard to comprehend...

Also, adding any new criteria or solutions will be pretty tedious.

Perhaps we should "default" everything to neutral, and only list "Evaluation" items that stand out positive / negative?

File: rfcs/InputUnion.md [hunk 3]
```diff
@@ -288,11 +308,21 @@ type Mutation {
...
-#### Single user-chosen field; value is the `type`
+#### Evaluation
+
+##### [Input unions should accept plain data from clients](#h-input-unions-should-accept-plain-data-from-clients)
+
+No, as there is an additional field that has to be added.
+
+##### [Input unions should be expressed efficiently in the query and on the wire](#k-input-unions-should-be-expressed-efficiently-in-the-query-and-on-the-wire)
+
+The additional field bloats the query.
```
binaryseed: Are we really considering a single additional field to be "bloat"? The APIs I deal with already have hundreds of fields, I don't think 1 more used in some types makes much difference

PR #652, opened 2019-11-26
Input Union: Evaluation cleanup

binaryseed:
After further reviewing the state of this document, I'm not sure we're ready to lay out a matrix of evaluations yet. I believe we need to first decide what criteria matter to us first.

To that end, I've temporarily removed the evaluation matrix (we can add it back in the future), done a small amount of cleanup, and added a few more "Objections" to the criteria.

I think we should try to reach agreement about which of the criteria are `MUST`. If a criteria is  `MUST`, then failing it would strike a solution off the table for consideration...

Perhaps each criterial should have `MUST`, `SHOULD` or `MAY` in it's title ?

cc/ @spawnia @jturkel @benjie

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -244,7 +244,7 @@ Preferably without loss of functionality.
 
 The less typing and fewer bytes transmitted, the better.
 
-* Objection: The quantity of "typing" isn't a worthwhile metric, most interactions with an API are programatic.
+* Objection: The quantity of "typing" isn't a worthwhile metric, most interactions with an API are programmatic.
 * Objection: Simply compressing an HTTP request will reduce the bytes transmitted more than anything having to do with the structure of a Schema.
 
 ### L) Input unions should be performant for servers
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -291,15 +291,16 @@ type Mutation {
 }
 ```
 
-#### Variations:
+#### Variations
 
 * A `default` type may be defined, for which specifying the `__typename` is not required. This enables a field to migration from an `Input` to an `Input Union`
 
 #### 🔬 Evaluation
 
-##### [Input unions should accept plain data](#h-input-unions-should-accept-plain-data)
-
-* ⚠️ One additional field is required.
+* [A - GraphQL should contain a polymorphic Input type](#a-graphql-should-contain-a-polymorphic-input-type)
+  * ✅
+* [H - Input unions should accept plain data](#h-input-unions-should-accept-plain-data)
+  * ⚠️ One additional field is required.
 
 -----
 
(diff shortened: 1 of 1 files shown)


rfcs/InputUnion.md
```diff
@@ -228,52 +244,19 @@ Preferably without loss of functionality.
 
 The less typing and fewer bytes transmitted, the better.
 
+* Objection: The quantity of "typing" isn't a worthwhile metric, most interactions with an API are programatic.
```
sungam3r:
```suggestion
* Objection: The quantity of "typing" isn't a worthwhile metric, most interactions with an API are programmatic.
```

rfcs/InputUnion.md
```diff
@@ -228,52 +244,19 @@ Preferably without loss of functionality.
...
-* [Content Widgets](https://github.com/graphql/graphql-js/issues/207#issuecomment-308344371)
-* [Filtering](https://github.com/graphql/graphql-spec/issues/202#issue-170560819)
-* [Observability Cloud Integrations](https://gist.github.com/binaryseed/f2dd63d1a1406124be70c17e2e796891#cloud-integrations)
-* [Observability Dashboards](https://gist.github.com/binaryseed/f2dd63d1a1406124be70c17e2e796891#dashboards)
+* Objection: None of the solutions discussed so far do anything that is computationally expensive.
 
 
 ## Possible Solutions
 
-Each of the solutions should be evaluated against each of the criteria.
```
spawnia: I think this overview is the most useful tool for us to eventually come to a conclusion. Let's not move backwards.

rfcs/InputUnion.md
```diff
@@ -467,52 +444,8 @@ type Mutation {
 
 ##### [Doesn't inhibit schema evolution](#c-doesnt-inhibit-schema-evolution)
 
-Adding a nullable field to an input object could change the detected type of
```
spawnia: That's removing useful information.

spawnia:
> I believe we need to first decide what criteria matter to us first.

We tried that before and have since then decided that it is more useful to collect and evaluate all possible criteria. This document is supposed to collect as much information about the RFC as possible.

Trimming, refining and evaluating that information would be the next step, after we collected everything. I think we are not here yet, so let's keep on adding to that document instead of removing things.

rfcs/InputUnion.md
binaryseed: The example was very verbose, all other Evaluations are a single line. I'm concerned about this document blowing out in size / understandability

I'm going to use the Github markdown collapse section to collapse it by default. Also I'll re-phrase it using the Animal example we use in the rest of the doc

binaryseed:
Ok I got in the zone and made quite a few changes / additions, I've done a rough evaluation on many of the criteria, and sprinkled quite a few emoji around to give the document some structure & fun.

I've also added 2 criteria regarding the impact the change would have on Introspection & SDL backwards compatibility...

🦃

spawnia:
@binaryseed great effort, the content is preserved and structured nicely.

I still think that a comparison matrix is more useful to get an overview. As of now, you have to aggregate that information in your head if you want to compare a solution against all criteria. I think that is ultimately the goal of this document.

rfcs/InputUnion.md
```diff
@@ -171,109 +183,144 @@ This section sketches out the potential goals that a solution might attempt to f
...
 
-### A) GraphQL should contain a polymorphic Input type
+### 🎯 A) GraphQL should contain a polymorphic Input type
 
 The premise of this RFC - GraphQL should contain a polymorphic Input type.
 
-### B) Input polymorphism matches output polymorphism
+| [1](#-1-explicit-__typename-discriminator-field) | [2](#-2-explicit-configurable-discriminator-field) | [3](#-3-order-based-discrimination) | [4](#-4-structural-uniqueness) | [5](#-5-one-of-tagged-union) |
+|----|----|----|----|----|
+| ✅ | ✅ | ✅ | ✅ | ⚠️ |
```
spawnia: What do those symbols mean exactly? A legend would be helpful.

rfcs/InputUnion.md
```diff
@@ -171,109 +183,144 @@ This section sketches out the potential goals that a solution might attempt to f
...
+### 🎯 B) Input polymorphism matches output polymorphism
 
 Any data structure that can be modeled with output type polymorphism should be able to be mirrored with Input polymorphism. Minimal transformation of outputs should be required to send a data structure back as inputs.
 
-* Objection: composite input types and composite output types are distinct. Fields on composite output types support aliases and arguments whereas fields on composite input types do not. Marking an output field as non-nullable is a non-breaking change, but marking an input field as non-nullable is a breaking change.
+* ✂️ Objection: composite input types and composite output types are distinct. Fields on composite output types support aliases and arguments whereas fields on composite input types do not. Marking an output field as non-nullable is a non-breaking change, but marking an input field as non-nullable is a breaking change.
+
+| [1](#-1-explicit-__typename-discriminator-field) | [2](#-2-explicit-configurable-discriminator-field) | [3](#-3-order-based-discrimination) | [4](#-4-structural-uniqueness) | [5](#-5-one-of-tagged-union) |
+|----|----|----|----|----|
+| ✅⚠️ | ✅ | ✅ | ✅⚠️ | 🚫 |
```
eapache: I actually consider an explicit `__typename` to have a stronger case here than any of the others, since it so precisely mirrors the `__typename` field you get back from output polymorphic types.

binaryseed: The field name does, but critically, the value of the field will never match since input & output types are distinct. That's why the evaluation for `1` has two evaluation "scores"

Solution `2` with some kind of literal values are the only explicit versions that would enable both the field name and the field value to match exactly...

binaryseed: The evaluation explanations are spelled out further under each of the solutions

rfcs/InputUnion.md
binaryseed: Added

binaryseed:
Added back the overview grid

binaryseed:
Merged! Feel free to PR with refinements / evaluations of other criteria

PR #655, opened 2019-12-04
InputUnion: tweak to plain data

benjie:
Regarding "Input unions should accept plain data", I think either 1,2 and 5 should all be 🚫 or all be ⚠️. Here's my reasoning:

Imagine we want to represent the cat `{name: "Buster", age: 3, livesLeft: 7}`

Options 1 and 2 require augmenting this object with an additional field, e.g. `{__typename: "CatInput", name: "Buster", age: 3, livesLeft: 7}`. This is certainly not the "plain data" the user supplied.

Option 5 instead wraps the "plain data" the user is using with a tag: `{cat: {name: "Buster", age: 3, livesLeft: 7}}`

Thus option 5 uses the plain data that the user provided, and wraps it with a simple tag. 1 and 2 require modification to the data itself - so it's certainly not "plain" any more.

Also: aesthetically, `{cat: {name: "Buster", age: 3, livesLeft: 7}}` seems more "plain" than `{__typename: "CatInput", name: "Buster", age: 3, livesLeft: 7}` to me.


binaryseed:
I disagree that the wrapped data is more "plain" since it's a different structure.

Perhaps this signals to us that we should better define this criteria...

That said, i'll merge if you update the summary table at the bottom of the doc as well...

benjie:
Ah; I missed that - fixed :heavy_check_mark: 

Excellent work on the document @binaryseed and everyone else! :+1:

PR #659, opened 2019-12-09
Fixed that `CoerceArgumentValues` incorrectly refers to `variableType`

fluidsonic:
… instead of `argumentType`


leebyron:
Thank you for the careful read and fix!

Issue #627, opened 2019-10-14
Consolidate Input Union discussions


binaryseed:
Over the years there have been many discussions around the topic of Input Unions in GraphQL. To help further the effort, we have begun building a single RFC document. The goal of this document is to consolidate all related ideas to help move the proposal forward.

* https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md

To help this effort, I'd like @leebyron to consider commenting on & closing all related open Issues / PRs in favor of the main RFC document, which will contain & reference all of these previous discussions.

* Polymorphic input types https://github.com/graphql/graphql-spec/issues/114 @luisobo
* RFC: inputUnion type https://github.com/graphql/graphql-spec/pull/395 @tgriesser
* Input polymorphism https://github.com/graphql/graphql-spec/issues/415 @jayhasyee
* [RFC] GraphQL Input Union type https://github.com/graphql/graphql-spec/issues/488 @frikille
* [RFC] Input Objects accepting exactly @oneField https://github.com/graphql/graphql-spec/pull/586 @benjie

@leebyron @IvanGoncharov

dendrochronology:
Thanks for all your hard work in consolidating this @binaryseed, @benjie, et al!

To whom can I send virtual beers so this gets on the working group's agenda? 🍻

benjie:
Thanks for the support and offer @dendrochronology; we discuss this frequently at working groups so it’s definitely already on our agenda :) We’re aiming to have the RFC doc updated by the next WG to enable participants to have a more informed discussion. It seems that there is no perfect solution (at least we haven’t found one yet!) so now it’s time to weigh up the different proposals and find out which trade-offs we can/can’t live with, which ones are the easiest to implement cross-platform, and which features we need/want/don’t care about.

binaryseed:
@leebyron Would you mind taking a look at this?!

IvanGoncharov:
@binaryseed Terribly sorry for the delay.
I closed all the issues and will review all pending PRs to RFC tomorrow.
Anything else I can do before closing this issue?

akomm:
I refer the [this](https://github.com/graphql/graphql-spec/pull/586#issuecomment-552362053) answer, I do not want to continue in the closed topic.

@benjie I have more and more the feeling, that the current graphql enumeration should have been more powerful in terms of its allowed values and work similar to how the proposed `oneField` works. Basically like the enums you have in rust (or similar languages ...). 

[...]

PR #661, opened 2019-12-17
Input unions: switch to reference-style links

benjie:
As discussed with @binaryseed here: https://github.com/graphql/graphql-spec/issues/627#issuecomment-564058007

Makes it a lot easier to add cross-references (no longer have to copy/paste the whole link) and makes it a lot easier to read the raw markdown (the tables in particular). Also means if we change a heading's text we only need to fix the URL in one place.


binaryseed:
Awesome, I've never seen this before but it's great!

Issue #577, opened 2019-04-13
Is the order of argument definitions significant?


spawnia:
I am aware the spec defines that the order in which the arguments are passed does not matter: https://graphql.github.io/graphql-spec/draft/#sec-Language.Arguments

We are currently dealing with an issue where the order in which those arguments are applied is significant, see https://github.com/nuwave/lighthouse/issues/728

Let me give you a simple example. Suppose we defined a query that concats two strings:

```graphql
type Query {
	concat(a: String!, b: String!): String
}
```

Given the rule that arguments are unordered, the following calls must return the same result. Both fields should return `foobar`.

```graphql
{
	concat(a: "foo", b: "bar")
    concat(b: "bar", a: "foo")
}
```

This indicates that on the server side, there must be an implicit ordering that determines how those arguments should be handled. In this particular case, the implementation will probably just do that by just doing something like this in my resolver.

```js
return args.a + args.b
```

Things are working fine at this point.

Now, i might want to grow the number of allowed arguments to the concat function. As the spec declares, the order in which the user passes them must be normalized. Instead of spelling out each argument, i might want to do something like this in my resolver:

```js
return reduce(args, (acc, item) => acc + item);
```

In that case, it would be practical for me if the arguments were always passed to the resolver in the order that i defined in the schema, instead of me having to take care of ensuring the correct order myself. This can be a subtle cause of bugs.

I am not quite sure if this issue falls within the scope of the spec or is just an implementation detail that might be handled by the various GraphQL libraries.
It would surely make it easier on implementors if they would not have to worry about
the order in which a user passes the arguments and are rather always passed them in
a previously defined order.

It would be worthwile to clarify if the schema definition itself can be used
to define the canonical order that arguments will be normalized to, or if the following two definitions are considered equivalent:

```graphql
type Query {
	concat(a: String!, b: String!): String
    # Switch the argument order in the schema definition
	concat(b: String!, a: String!): String
}
```

nodkz:
For named parameters, the order is not significant.
Sorting arguments for every request at runtime are a costly task, and the majority of users does not require it. So for what purpose we should do a redundant job?

So if you need `concat` operation with ordered args, then you need 
- to use one arg with Array shape. Order in arrays is guaranteed.
- or you may read fourth argument `info`  in your resolve method, with will contain AST of the query from which you may read the order of provided args from the client.

I don't feel that reordering of arguments from client according to the schema is common task.

[...]

PR #673, opened 2020-01-13
Servers must preserve lexical ordering between SDL and introspection

spawnia:
Resolves https://github.com/graphql/graphql-spec/issues/577

This change adds the proposed addition by @leebyron from https://github.com/graphql/graphql-spec/issues/577#issuecomment-519226398

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -56,6 +56,7 @@ would return
 }
 ```
 
+
 ## Reserved Names
 
 Types and fields required by the GraphQL introspection system that are used in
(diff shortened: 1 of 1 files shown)


benjie:
I approve of the intent behind this. It doesn't feel like putting the current proposed edit in chapter 4 is quite right though. We have effectively this global rule:

> The order of fields, arguments, etc in a GraphQL schema is significant

Then that comes with two corollaries:

1. The order of fields, arguments, etc in SDL is significant - when building a schema from SDL these orders should be maintained.
2. The order of fields, arguments, etc in introspection is significant - when introspecting the schema these orders should be maintained.

(1) Should be in section 3 ("Type System"), whereas (2) should be in section 4 ("Introspection"). I think (2) should come under each of the `Fields:` prompts, e.g.

> - fields must return the set of fields that can be selected for this type.
>   - Accepts the argument includeDeprecated which defaults to false. If true, deprecated fields are also returned.
>   - Fields must be returned in the order in which they are defined in the schema

Similar localised edits should be made in section 3 I think. I'm not sure where the global rule belongs exactly, it could be that section 3 covers it.

PR #676, opened 2020-01-24
[Input Union] Fix inconsistency vs evaluation

benjie:
Evaluation notes that solution 5 passes this criteria since a tagged union is a valid form of polymorphism (as shown by the prior art); but the emoji don't match up with this.


benjie:
cc @binaryseed

binaryseed:
Thanks!

PR #677, opened 2020-01-24
[Input Union] Change criteria scoring

benjie:
This is a PR to open a discussion on a few of the Input Union criteria scores (I don't think we have discussion on this elsewhere?)

Relevant: https://github.com/graphql/graphql-spec/pull/668

Please treat the three discussions separately :)

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -218,7 +218,7 @@ Any data structure that can be modeled with output type polymorphism should be a
 |----|----|----|----|----|
 | ✅⚠️ | ✅ | ✅ | ✅⚠️ | 🚫 |
 
-Criteria score: 🥇
+Criteria score: 🥈
 
 ## 🎯 C. Doesn't inhibit schema evolution
 
@@ -255,7 +255,7 @@ In addition to containing Input types, member type may also contain Leaf types l
 |----|----|----|----|----|
 | 🚫 | 🚫 | ✅⚠️ | 🚫 | ✅ |
 
-Criteria score: 🥉
+Criteria score: 🥈
 
 ## 🎯 F. Migrating a field to a polymorphic input type is non-breaking
 
@@ -292,7 +292,7 @@ Clients should be able to pass "natural" input data to unions without specially 
 |----|----|----|----|----|
 | ⚠️ | ⚠️ | ✅ | ✅ | ⚠️ |
 
-Criteria score: 🥈
+Criteria score: 🥉
 
 ## 🎯 I. Input unions should be easy to upgrade from existing solutions
 
@@ -838,13 +838,13 @@ A quick glance at the evaluation results. Remember that passing or failing a spe
 |    | [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
 | -- | -- | -- | -- | -- | -- |
 | [A][criteria-a] 🥇 | ✅ | ✅ | ✅ | ✅ | ⚠️ |
-| [B][criteria-b] 🥇 | ✅⚠️ | ✅ | ✅ | ✅⚠️ | 🚫 |
+| [B][criteria-b] 🥈 | ✅⚠️ | ✅ | ✅ | ✅⚠️ | 🚫 |
 | [C][criteria-c] 🥇 | ✅ | ✅ | 🚫 | ⚠️ | ✅ |
 | [D][criteria-d] 🥇 | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [E][criteria-e] 🥉 | 🚫 | 🚫 | ✅⚠️ | 🚫 | ✅ |
-| [F][criteria-f] 🥉 | ✅⚠️ | ✅⚠️ | ✅ | ⚠️ | ✅ |
+| [F][criteria-f] 🥈 | ✅⚠️ | ✅⚠️ | ✅ | ⚠️ | ✅ |
 | [G][criteria-g] 🥉 | ❔ | ❔ | ❔ | ❔ | ❔ |
-| [H][criteria-h] 🥈 | ⚠️ | ⚠️ | ✅ | ✅ | ⚠️ |
+| [H][criteria-h] 🥉 | ⚠️ | ⚠️ | ✅ | ✅ | ⚠️ |
 | [I][criteria-i] 🥉 | ✅⚠️ | ✅⚠️ | ✅ | ⚠️ | ✅ |
 | [J][criteria-j] 🥇 | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [K][criteria-k] 🥉 | ❔ | ❔ | ❔ | ❔ | ✅ |


rfcs/InputUnion.md
```diff
@@ -218,7 +218,7 @@ Any data structure that can be modeled with output type polymorphism should be a
 |----|----|----|----|----|
 | ✅⚠️ | ✅ | ✅ | ✅⚠️ | 🚫 |
 
-Criteria score: 🥇
+Criteria score: 🥈
```
benjie: I'd like to challenge the importance of input polymorphism matching output polymorphism. To me, this seems like it'd be a nice to have, but certainly not a requirement. It's clear whatever we do input unions and output unions will never be 100% equivalent. The fields on output types in GraphQL are very flexible: supporting aliases and arguments, and allowing you to specify the same field multiple times, or omit the field even if it's non-nullable. Input objects are much more rigid by comparison. Output type fields also support fragment spreads, directives, and other features that likely don't make sense for input types. It's common for input and output objects to evolve separately, and often you'll have input types that have no associated output type and vice-versa.

rfcs/InputUnion.md
```diff
@@ -255,7 +255,7 @@ In addition to containing Input types, member type may also contain Leaf types l
 |----|----|----|----|----|
 | 🚫 | 🚫 | ✅⚠️ | 🚫 | ✅ |
 
-Criteria score: 🥉
+Criteria score: 🥈
```
benjie: I think that this might be higher importance than currently rated; it's certainly a popular request here: https://github.com/graphql/graphql-spec/issues/215

rfcs/InputUnion.md
```diff
@@ -292,7 +292,7 @@ Clients should be able to pass "natural" input data to unions without specially
 |----|----|----|----|----|
 | ⚠️ | ⚠️ | ✅ | ✅ | ⚠️ |
 
-Criteria score: 🥈
+Criteria score: 🥉
```
benjie: I'm not convinced this is important - in fact I actively think this is a bad idea. All the transport-level prior art except OpenAPI 3 and JSON Schema have some extra metadata somehow to indicate the data type. The programming languages also have this (they can know what a type is in memory), but they don't seem so relevant here, so I've only included ones that talk about serialization/memory format:

- Protocol Buffers / [Oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof)
  > You also get a special method for checking which value (if any) in the oneof is set. 
- FlatBuffers / [Union](https://google.github.io/flatbuffers/flatbuffers_guide_writing_schema.html)
  > additionally a field with the suffix _type is generated that holds the corresponding enum value, allowing you to know which type to cast to at runtime
- Cap'n Proto / [Union](https://capnproto.org/language.html#unions) (interestingly this is basically solution 5, and they call it Union anyway)
  > A union is two or more fields of a struct which are stored in the same location. Only one of these fields can be set at a time, and a separate tag is maintained to track which one is currently set.
- Thrift / [Union](https://thrift.apache.org/docs/idl#union) (solution 5)
  > Unions are similar to structs, except that they provide a means to transport exactly one field of a possible set of fields, just like union {} in C++. Consequently, union members are implicitly considered optional (see requiredness).
- Arvo / [Union](https://avro.apache.org/docs/current/spec.html#Unions)
  >  two types with different names are permitted. (Names permit efficient resolution when reading and writing unions.)

benjie:
CC @binaryseed

rfcs/InputUnion.md
binaryseed: The reason I consider this so critical is because it supports the use cases that drove me to need this in the first place - ones where the goal is to manage a complex data structure. Without the ability to model a concept with the same structure, we push complexity into the client code that has to deal with that.

Take a look at those use cases I linked to like "Dashboards":
* https://gist.github.com/binaryseed/f2dd63d1a1406124be70c17e2e796891#dashboards

To solve this use case, we are forced to abandon object side polymorphism all-together...

rfcs/InputUnion.md
binaryseed: I ranked it bronze is because it's not currently supported in object polymorphism so it'd create a brand new discontinuity between input and object polymorphism

rfcs/InputUnion.md
binaryseed: This is an interesting situation - binary transports have a much easier time here since they can just encode the type with a few bytes and move on with things. Since we're dealing with "human readable" JSON it becomes a "style" thing and people have really strong opinions.

I'm inclined to agree with you here.

rfcs/InputUnion.md
benjie: I don't think that the objects have to necessarily be the exact same format to enable your widgets use-case - if they did with no transformation then it would suggest that you'd have to only expose fields on the output that were available on the input, you'd have to be very careful to request exactly the right fields, and if you had relations you'd have to ensure there were no cycles otherwise it could not be represented as an input.

I don't think anyone is arguing for these restrictions (the section even states "Minimal transformation of outputs should be required to send a data structure back as inputs"), so it's clear that the output _has to be transformed in some way_ to make it appropriate for usage as an input - even if we were to go with solution 2 or 3. Examples of these transformations would be trimming off any fields not appropriate for the input (required for all solutions); plus for solution 1 you'd change the `__typename` value (maybe from `Foo` to `FooInput`), or for solution 5 you'd wrap in a container object.

For nested forms these transformations will have to be recursive in case the objects themselves have references to other union types that need converting to inputs. Assuming that we allow for this (and why wouldn't we?) no matter your programming paradigm (object-oriented, functional, etc) you'll need to add some methods/functions to aid with doing a recursive transform in order to send the objects output from GraphQL back as an input to GraphQL (after any modifications you've made in the client). None of the solutions escape this need (unless we apply overbearing restrictions on how you can use input unions).

Now we've established that a transform is pretty much always required except in the simplest of cases, here's what marshalling may look like for solutions 1, 3 and 5:

```js
function marshallFoo1(obj) {
  const { __typename } = obj;
  const filtered = pick(obj, validInputKeysForOutputType(__typename));
  return { ...filtered, bar: marshallBar1(obj.bar), __typename: __typename + "Input" };
}
function marshallFoo3(obj) { // Also valid for solution 2
  const { __typename } = obj;
  const filtered = pick(obj, validInputKeysForOutputType(__typename));
  return { ...filtered, bar: marshallBar1(obj.bar) };
}
function marshallFoo5(obj) {
  const { __typename } = obj;
  const filtered = pick(obj, validInputKeysForOutputType(__typename));
  return { [__typename]: { ...filtered, bar: marshallBar1(obj.bar) } };
}
```

They all follow the pattern: find out the `__typename`; filter the fields to just those that are valid as input for this specific type, mix in the recursive marshalling of any child objects (`bar` in this case), and finally format an object for output. It's only this last step that differs between the solutions; solution 1 adds the new `__typename`; solution 3 just returns the filtered object and hopes structural analysis is sufficient; and solution 5 uses a wrapper object. These are all very similar and simple transforms.

TL;DR:

1. This check isn't very valuable: it's clear transformation is required whatever solution is adopted, so arguably no solution can ever pass without the constraints in paragraph 1 (and more that I edited out for brevity(!))
1. Solutions 2 and 3 still need transforms (and care must be applied when querying the data from the API to ensure all required fields are requested), so they shouldn't get a straight pass - they should have warnings too
1. I don't think 1 and 4 somehow pass this check with a warning whereas 5 does not - the marshall functions are extremely similar, and the resulting structures are similarly human readable: `{ __typename: "CatInput", livesLeft: 3 }` vs `{ Cat: { livesLeft: 3 } }`.

---

I believe my main concern with this is how it's been evaluated due to the word "mirrored" in the requirement; it's being evaluated based on aesthetics rather than functionality. I challenge the need for it to be an explicit reflection, but I recognise the need for inputs to be capable of expressing similar concepts to outputs (even if the format/metadata/etc is slightly different).

rfcs/InputUnion.md
benjie: I believe this distinction is acceptable, and it's clear to me the functionality is desired. I don't think our existing output polymorphism should place unnecessary restraints on our input polymorphism solution - we may yet find a way to solve this with output polymorphism (e.g. with value boxing `{__typename: "Int", value: 3}`).

rfcs/InputUnion.md
benjie: I feel quite strongly that not _somehow_ indicating the concrete type would be a very serious mistake we'll come to regret.

Would you like me to open a separate PR with just this change?

binaryseed: Sure, we can merge that and continue discussing the others

benjie: 👍  https://github.com/graphql/graphql-spec/pull/678

binaryseed: merged

benjie:
Closing since discussions have moved on since then.

PR #686, opened 2020-02-07
make clear that some String escape sequences are optional

andimarek:
Escaping is not always required, but rather optional if the code point can be represent directly inside a String literal. This small change makes it clear.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -821,10 +821,10 @@ be interpreted as the beginning of a block string. As an example, the source
 {`""""""`} can only be interpreted as a single empty block string and not three
 empty strings.
 
-Note: Unicode characters are allowed within standard quoted strings, however
-{SourceCharacter} must not contain some ASCII control characters so escape
-sequences must be used to represent these characters. Also {`\`}, {`"`} and {LineTerminator}
-must be escaped. All other escape sequences are optional.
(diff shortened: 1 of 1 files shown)


spec/Section 2 -- Language.md
```diff
@@ -823,7 +823,8 @@ empty strings.
 
 Note: Unicode characters are allowed within String value literals, however
 {SourceCharacter} must not contain some ASCII control characters so escape
-sequences must be used to represent these characters.
+sequences must be used to represent these characters. Also {`\`}, {`"`} and {LineTerminator}
```
IvanGoncharov: @andimarek Block strings can contain unescaped `{`"`} and {LineTerminator}`

andimarek: Block string don't allow any escape sequences except `\"""` so this whole sections only applies to non block strings.

andimarek: But it is a bit confusing because it refers to "String value literals` which according to the grammar are actually normal Strings and Block Strings. I guess this was written before Block Strings was a thing.

andimarek: I updated the section to talk about standard quoted string, to make it more clear.

IvanGoncharov: @andimarek Thanks for the clarification 👍

PR #692, opened 2020-02-25
Add discussion on client re-renders, batched responses

robrichard:
Adding some potential approaches to solving the issue of clients re-rendering in quick succession due to `@defer` or `@stream` payloads. This was discussed previously in this comment: https://github.com/graphql/graphql-spec/pull/679#issuecomment-580506294.


rfcs/DeferStream.md
```diff
@@ -17,4 +17,110 @@ A `@defer` directive would allow clients to annotate fragments that are non-crit
...
 - [2] [[Proposal] Introducing @defer in Apollo Server](https://blog.apollographql.com/introducing-defer-in-apollo-server-f6797c4e9d6e)
-- [3] [Active development on @defer and @stream in Relay at Facebook](https://github.com/graphql/graphql-wg/issues/329)
\ No newline at end of file
+- [3] [Active development on @defer and @stream in Relay at Facebook](https://github.com/graphql/graphql-wg/issues/329)
+
+## Potential concerns, challenges, and drawbacks
+
+### Client re-renders
+
+With incremental delivery, where multiple responses are delivered in one request, client code could re-render it's UI multiple times in a short period of time. This could degrade performance of the application, negating the performance gains from using `@defer` or `@stream`. There are a few approaches that could be taken to mitigate this. Each of these approaches are orthogonal to one another, i.e. the working group could decide that more than one of these should be included in the spec or be labeled as best practices.
```
benjie:
```suggestion
With incremental delivery, where multiple responses are delivered in one request, client code could re-render its UI multiple times in a short period of time. This could degrade performance of the application, negating the performance gains from using `@defer` or `@stream`. There are a few approaches that could be taken to mitigate this. Each of these approaches are orthogonal to one another, i.e. the working group could decide that more than one of these should be included in the spec or be labeled as best practices.
```

rfcs/DeferStream.md
```diff
@@ -17,4 +17,110 @@ A `@defer` directive would allow clients to annotate fragments that are non-crit
...
+            }
+        }
+    }
+}
+```
+
+It could then process each of these before triggering a re-render.
+
+2. __Client side debounce.__ GraphQL clients can debounce the processing of responses before triggering a re-render. For a query that contains `@defer` or `@stream`, the client will wait a predetermined amount of time starting from when a response is received. If any additional responses are received in that time, it can process the results in one batch. This has the downside of adding latency; if no additional responses are receiving in the timeout period, the processing of the initial response is delayed by the length of the debounce timeout.
+3. __Server sends batched responses.__ This approach changes the spec to allow GraphQL to return either the current GraphQL Response map, or a list of GraphQL Response maps. This gives the server the flexibility to determine when it is beneficial to group incremental responses together. If several responses are ready at the same time, the server can deliver them together. The server may also have knowledge of how long resolvers will take to resolve and could choose to debounce.
```
Keweiqu: My intuition is that this would be the direction we want to pursue precisely for the flexibility you mention here. When experimenting at Facebook, we tried client side debouncing first, and due to the complexity of figuring out a reasonable interval, we ended up with a naive debouncing which is always rendering twice, once for the initial payload and once after all incremental payloads have arrived. 
That turns out to be not granular enough, so we implemented server side batching. One complication is that the GraphQL executor actually doesn't have a good sense of how expensive a resolver is. Therefore, it's hard for the executor to make a decision on the fly while executing a query. It's more applicable if we outsource the decision of how to batch to the underlying @stream and @defer resolvers. In the end, we only implemented batching for @stream since a global batching at the query level means @deferred and @streamed resolvers need to communicate with each other in some sense. 

I wonder if we can offer practical suggestions for servers to implement server side batching. If not, should we instead propose something like @stream level batching?

robrichard:
@Keweiqu thanks for the comments, I added some more information about debouncing as you suggested.

IvanGoncharov:
@robrichard Can you please rebase so I can merge it?

PR #694, opened 2020-03-04
Add me to champions

binaryseed:


rfcs/InputUnion.md
```diff
@@ -754,6 +756,8 @@ input DogInput {
 
 ## 💡 5. One Of (Tagged Union)
 
+**Champion:** @binaryseed
```
binaryseed: @benjie Do you want to join the champion list for this one?

binaryseed: I went ahead and added you here instead of me, even though I'm a fan of this idea as well :)

benjie: Yeah! Thanks; I thought I'd done whatever was necessary back in December - seems I missed adding myself :man_facepalming: 

https://github.com/graphql/graphql-spec/issues/627#issuecomment-562504735

PR #702, opened 2020-03-30
Add a caveat to input union solution 2 criteria B

eapache:
This caveat was mentioned on solution 1 but also applies to solution 2. @binaryseed

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -216,7 +216,7 @@ Any data structure that can be modeled with output type polymorphism should be a
 
 | [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
 |----|----|----|----|----|
-| ✅⚠️ | ✅ | ✅ | ✅⚠️ | 🚫 |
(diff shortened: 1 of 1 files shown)


rfcs/InputUnion.md
```diff
@@ -555,6 +555,7 @@ input DogInput {
   * ✅
 * [B. Input polymorphism matches output polymorphism][criteria-b]
   * ✅ Data structures can mirror eachother.
+  * ⚠️ `__typename` can not match discriminator since Input & Output types are distinct (ex: `Cat` vs `CatInput`).
```
binaryseed: I'm not sure this is true - this solution is suggesting a pre-defined "literal" value as the discriminator, not the GraphQL type name.

Perhaps the confusion is that the first "variation" here is actually using the type name, I should probably just remove that in favor of the string / enum literal variations.

I think that's the main point of the solution, to enable the same data to work on query &. mutation sides...

eapache: Ah, yeah I definitely saw the type names in the example and made that assumption, it would be good to clarify.

PR #703, opened 2020-03-31
[Editorial] Fix example for Variable Uniqueness

PascalSenn:
I think the fragment definition in this example was wrong


IvanGoncharov:
@PascalSenn Thanks a lot 👍

PR #707, opened 2020-04-07
Contributing: issue or PR clarification

francisu:
The contributing document was not clear on whether you should start with an issue or PR, and in looking through them, I see examples of both. So I assumed you must start with an issue. However then I was making the issue, it suggested I do a PR if I could.

So I'm providing this small update to the contributors document to clarify this point.


IvanGoncharov:
@francisu Sound reasonable, thanks for PR 👍

PR #712, opened 2020-04-25
[grammar] add "the " to Type Condition section

micimize:
4 character PR 👍 

There's also a reference to an earlier example in the "Fragments" section that reads a lot like the example was accidentally omitted:
> Fragments must specify the type they apply to. In this example, `friendFields`
can be used in the context of querying a `User`.


IvanGoncharov:
@micimize Thanks for PR 👍

PR #716, opened 2020-05-02
Add criteria: Unconstrained combination of input types to unions

spawnia:
This added criteria captures an aspect of input unions that i had in the back of my mind for a while, i just got around to nailing it down and articulating it.

Solutions 1. and 5. place little constraints on schema design. While they do add some general complexity, those don't bother me once they are dealt with in a generic way.

Solution 2., 3. and 4. might force me to name, arrange or constrain the input types and combinations thereof in ways that are quite unnatural. They constantly force me to think about how the field names and non-nullability of fields might affect the composition of input unions.

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -229,7 +229,7 @@ Adding a new member type to an Input Union or doing any non-breaking change to e
 
 | [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
 |----|----|----|----|----|
-| ✅ | ✅ | 🚫 | ⚠️ | ✅ |
+| ✅ | ⚠ | 🚫 | ⚠️ | ✅ |
 
 Criteria score: 🥇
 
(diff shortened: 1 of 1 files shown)


spawnia:
@binaryseed @benjie @leebyron @eapache @sungam3r @IvanGoncharov Champions, what do you think?

sungam3r:
I have been watching this thread for a long time and I hope that at least some decision will be made. It is already clear that there will be no ideal.

spawnia:
> I have been watching this thread for a long time and I hope that at least some decision will be made. It is already clear that there will be no ideal.

I understand that sentiment, it has taken a long time. However, i think that solution 2., 3. and 4. have major problems that are not evaluated in this document yet. Deciding to incorporate them means we basically have to live with them forever.

Their design seems simple enough and they seem to fit the criteria somewhat well. I would even say that they feel more simple, more natural at first glance.

In practice, they constrain schema design in ways that are far from intuitive. All of them influence how input types have to be structured in order to be used as union types. They require (or in the case of 3. incentivize) input types to be unique in some aspect, be it through the presence of a particular field (2.) or structurally (4.).

Those restrictions might forbid certain schema designs or force them to be modeled in a certain way. Going from simple input types to input unions is not just a matter of defining that association, but might require changes or upfront consideration in order to work.

The structural uniqueness constrains of 3. and 4. might lead to a emerging best practice of giving all input type fields unique prefixes, in order to allow all possible unions between them.

Solution 2. makes it naturally hard for input types to be a part of multiple unions.

benjie:
I agree with the intent of this 👍 Input unions should allow arbitrary inputs to be added to the union without requiring modification to the underlying input types.

I think 3 passes this requirement by virtue of just picking the first match and not worrying about specifics.

leebyron:
Well framed. Solution 3 does pass this criteria.

I think this concern was captured in our subcommittee meetings but wasn't cleanly captured in the RFC doc, so thank you for doing this

PR #717, opened 2020-05-05
Updates to defer/stream RFC

robrichard:
This builds on top of @Keweiqu's changes in https://github.com/graphql/graphql-spec/pull/679

### DIFF
--- a/rfcs/DeferStream.md
+++ b/rfcs/DeferStream.md
@@ -3,18 +3,100 @@ RFC: GraphQL Defer and Stream Directives
 
 *Working Draft - January 2020*
 
-**Introduction**
+# Introduction
 
-While the definition for "fast" and "slow" varies across applications, a universal concept they share is the existence of critical elements. These are parts of an application that should be immediately available to the end-user. Prioritizing these critical elements often involves deprioritizing their non-critical counterparts. This solution usually entails splitting out non-critical fragments and fields into separate queries, which incur the cost of optimizing critical elements while worsening non-critical ones. This document proposes directives `@defer` and `@stream` that introduce the ability to within a query selectively fetch fragments and items in a list, respectively, without impacting the response time of the query or of the deferred fields.
+One of the disadvantages of GraphQL’s request/response model is that applications which retrieve large datasets may suffer from latency. However not all requested data may be of equal importance, and in some use cases it may be possible for applications to act on a subset of the requested data.  
 
-**Background**
+Today applications which seek to ensure that data of high importance can be retrieved quickly have a few options:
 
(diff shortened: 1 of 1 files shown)


mike-marcacci:
This is just a nit, but for the OSS RFC, we _probably_ want to camelCase the identifiers `initial_count` and `is_final` to match the convention used elsewhere in GraphQL.

rfcs/DeferStream.md
```diff
@@ -3,18 +3,146 @@ RFC: GraphQL Defer and Stream Directives
...
+  * When `true` fragment may be deferred, if omitted defaults to `true`.
+* `label: String!`
+  * A unique label across all `@defer` and `@stream` directives in an operation.
+  * This `label` should be used by GraphQL clients to identify the data from patch responses and associate it with the correct fragment.
+
+## `@stream`
+
+The `@stream` directive may be provided for a field of `List` type so that the backend can leverage technology such asynchronous iterators to provide a partial list in the initial response, and additional list items in subsequent responses. `@include` and `@skip` take presedence over `@stream`.
+
+### `@stream `arguments
```
fotoetienne, line 50: Should there also be a way for clients to limit the number items that would be returned in a stream? i.e. `maxCount`

rfcs/DeferStream.md
```diff
@@ -3,18 +3,146 @@ RFC: GraphQL Defer and Stream Directives
...
+## Splitting queries
 
+Given a query where some fields are expensive and non-essential, non-essential fields can be fetched in a separate query issued after an initial query. In the context of lists, this may mean fetching only the first few items in a list and issuing a follow-up pagination request for additional items. By separating the requests based on data priority, we have achieved a faster delivery time to essential data. However, query splitting imposes several **trade-offs** on applications:
+
+**Increased latency for lower priority fields,** as now these fields have to wait for the original query to complete and have added network round trips
+**Client resource contention.** Issuing multiple requests can increase contention on scarce resources like battery and antenna.
```
fotoetienne, line 22: Would this be true for requests that are multiplexed over HTTP/2 connections?

rfcs/DeferStream.md
robrichard: @fotoetienne is this different from the pagination arguments that would typically be used on list or connection fields? Would `maxCount` affect the streaming mechanism itself or would it change the data that is returned?

rfcs/DeferStream.md
robrichard: I'm not sure. Splitting queries usually requires data from the first query to be used as input to the second query, so they are not able to run in parallel.

leebyron:
Including these updates to keep the RFC moving, thank you!

PR #718, opened 2020-05-08
[Input Union]: small grammar fix in RFC

bgentry:
Hi there, I was skimming through the input unions RFC and noticed a small grammar issue. Figured I might as well take a few seconds to fix it. Cheers :v:


binaryseed:
thanks!

PR #721, opened 2020-05-26
[RFC] Allow @deprecated directive for input fields

Andy2003:
The PR fixes the problem that in the current specification it is not allowed to use the `@deprecated` directive for input fields.


sungam3r:
Relates to #525 and #197

IvanGoncharov:
@Andy2003 Thanks for taking the time to make this PR 👍 
This is a subset of changes proposed in #525 so closing this as duplicate.
BTW We are planing to move #525 to Stage2 on WG next week.

PR #723, opened 2020-05-30
Small fixes and clarifications to contributing

mike-marcacci:
Small grammar and clarity improvements to `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,7 @@ move forward. See editor Lee Byron talk about
 
 * **Performance is a feature**
 
-  GraphQL typically avoids syntax or behaviors that could jepordize runtime
+  GraphQL typically avoids syntax or behaviors that could jeopardize runtime
   efficiency, or that make demands of GraphQL services which cannot efficiently
   be fulfilled.
 
@@ -76,15 +76,15 @@ move forward. See editor Lee Byron talk about
 
 * **Simplicity and consistency over expressiveness and terseness**
 
-  There are plenty of behaviors and patterns found in other languages that are
-  intentionally absent from GraphQL. "Possible but awkward" is often favored
-  over more complex alternatives. Simplicity (e.g. fewer concepts) is
-  more important than expressing more sophisticated ideas or writing less.
+  Plenty of behaviors and patterns found in other languages are intentionally
+  absent from GraphQL. "Possible but awkward" is often favored over more complex
+  alternatives. Simplicity (e.g. fewer concepts) is more important than
+  expressing more sophisticated ideas or writing less.
 
 * **Preserve option value**
 
-  It's hard to know what the future brings, so whenever possible, decisions
-  should be made that allow for more options in the future. Sometimes this is
+  It's hard to know what the future brings; whenever possible, decisions should
+  be made that allow for more options in the future. Sometimes this is
   unintuitive: spec rules often begin more strict than necessary with a future
   option to loosen when motivated by a real use case.
 
@@ -177,9 +177,9 @@ libraries *may* implement *proposals*, though are encouraged to not enable the
 ## Stage 2: *Draft*
 
 An RFC at the *draft* stage is a fully formed solution. There is working group
-consensus that the problem identified should be solved, and that this particular
-solution is preferred. A *draft's* goal is to precisely and completely describe
-the solution and resolve any concerns through library implementations. A *draft*
+consensus the problem identified should be solved, and this particular solution
+is preferred. A *draft's* goal is to precisely and completely describe the
+solution and resolve any concerns through library implementations. A *draft*
 must be a pull request.
 
 *Entrance criteria:*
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
Not a native speaker so can't review grammar change.
@benjie Can you please take a look?

CONTRIBUTING.md
```diff
@@ -56,9 +56,9 @@ move forward. See editor Lee Byron talk about
 
 * **Performance is a feature**
 
-  GraphQL typically avoids syntax or behaviors which could place burden on
-  runtime efficiency or make demands of a GraphQL service it cannot
-  efficiently fulfill.
+  GraphQL typically avoids syntax or behaviors that could jepordize runtime
```
eapache, line 59: typo: jeopardize

CONTRIBUTING.md
```diff
@@ -70,21 +70,21 @@ move forward. See editor Lee Byron talk about
...
   intentionally absent from GraphQL. "Possible but awkward" is often favored
   over more complex alternatives. Simplicity (e.g. fewer concepts) is
   more important than expressing more sophisticated ideas or writing less.
 
 * **Preserve option value**
 
-  It's hard to know what the future brings, so whenever possible decisions
-  should be made which allow for more options in the future. Sometimes this is
+  It's hard to know what the future brings, so whenever possible, decisions
+  should be made that allow for more options in the future. Sometimes this is
```
benjie, line 87: This comma isn't quite right - it creates a parenthesised text which cannot be safely deleted. Effectively you want the "whenever possible" to be parenthesized, as in:

```
  It's hard to know what the future brings so, whenever possible, decisions
  should be made that allow for more options in the future. Sometimes this is
```

Personally, I'd rewrite it as:

```suggestion
  It's hard to know what the future brings; whenever possible, decisions
  should be made that allow for more options in the future. Sometimes this is
```

CONTRIBUTING.md
```diff
@@ -70,21 +70,21 @@ move forward. See editor Lee Byron talk about
...
-  examples are always more interesting than theoretical ones, and common
-  scenarios are more interesting than rare ones. RFCs should do more than offer
+  examples are always more compelling than theoretical ones, and common
+  scenarios are more compelling than rare ones. RFCs should do more than offer
   a different way to reach an already achievable outcome.
 
 * **Simplicity and consistency over expressiveness and terseness**
 
-  There are plenty of behaviors and patterns found in other languages
+  There are plenty of behaviors and patterns found in other languages that are
```
benjie, line 79: I think this is fine with or without `that are`; without it is the more condensed form, in general I try to avoid putting the word `that` into my prose as it's rarely necessary.

https://english.stackexchange.com/questions/50962/what-is-a-noun-modifying-clause/50971#50971
```suggestion
  There are plenty of behaviors and patterns found in other languages
```

CONTRIBUTING.md
```diff
@@ -177,7 +177,7 @@ libraries *may* implement *proposals*, though are encouraged to not enable the
 ## Stage 2: *Draft*
 
 An RFC at the *draft* stage is a fully formed solution. There is working group
-consensus that the problem identified should be solved, and this particular
+consensus that the problem identified should be solved, and that this particular
```
benjie, line 180: Yes, either this `that` needs to be added, or both `that`s need to be removed (which would be my preference):

```suggestion
consensus the problem identified should be solved, and this particular
```

CONTRIBUTING.md
benjie, line 59:
```suggestion
  GraphQL typically avoids syntax or behaviors that could jeopardize runtime
```

eapache:
> I'm not certain of the rules of that vs which.

https://www.quickanddirtytips.com/education/grammar/which-versus-that-0

AFAICT all of the `s/which/that` changes made in this PR are correct. Maybe a bit pedantic at this point, but technically correct.

CONTRIBUTING.md
mike-marcacci: I switched around the sentence to avoid awkwardness without "that".

mike-marcacci:
Thanks for all the extra eyes on this! I had recently become re-familiarized with the "which vs that" rules, so when I was looking at this file to make sure it was appropriate for folks who are new to the project, I couldn't help but notice the "mistake" that I often make.

These are TOTALLY pedantic, and I'm not particularly passionate about any of these changes, but figured I'd submit my edits nonetheless.

I've taken all the suggestions above: feel free to suggest any more.

CONTRIBUTING.md
```diff
@@ -70,21 +70,21 @@ move forward. See editor Lee Byron talk about
 * **Enable new capabilities motivated by real use cases**
 
   Every change should intend on unlocking a real and reasonable use case. Real
-  examples are always more interesting than theoretical ones, and common
-  scenarios are more interesting than rare ones. RFCs should do more than offer
+  examples are always more compelling than theoretical ones, and common
+  scenarios are more compelling than rare ones. RFCs should do more than offer
```
mike-marcacci, line 74: This is one of the more "meaningful" changes (if you will), because IMO theoretical use-cases are often _extremely_ interesting... they're just not compelling justifications for making a costly change. Real-world use cases, even if banal and boring, have much more weight, as the implementation/change cost is justified by usefulness.

benjie: I agree with this change and its reasoning :+1:

PR #725, opened 2020-06-02
Fix missing section headers for "Formal Specifications"

benjie:
All the `**Formal Specification**` headers in the document currently do not have permalinks because they are not detected by `spec-md` to be sections due to indentation of the bulleted list.

Example: [This explanatory text](https://spec.graphql.org/draft/#sec-Operation-Name-Uniqueness.Explanatory-Text) is a section and is linkable, but the "Formal Specification" above it is not.

To solve this, I've unindented these bulleted lists.

What follows is a full diff of the changes this causes in the compiled spec:

```
--- before.html	2020-06-02 12:15:35.955790531 +0100
+++ after.html	2020-06-02 12:18:08.307718120 +0100
@@ -4053,11 +4053,13 @@
 <h3><span class="spec-secid" title="link to this section"><a href="#sec-Documents">5.1</a></span>Documents</h3>
 <section id="sec-Executable-Definitions" secid="5.1.1">
 <h4><span class="spec-secid" title="link to this section"><a href="#sec-Executable-Definitions">5.1.1</a></span>Executable Definitions</h4>
-<p><strong>Formal Specification</strong></p>
+<section id="sec-Executable-Definitions.Formal-Specification" class="subsec">
+<h6><a href="#sec-Executable-Definitions.Formal-Specification" title="link to this subsection">Formal Specification</a></h6>
 <ul>
 <li>For each definition <var data-name="definition">definition</var> in the document.</li>
 <li><var data-name="definition">definition</var> must be <span class="spec-nt"><a href="#OperationDefinition" data-name="OperationDefinition">OperationDefinition</a></span> or <span class="spec-nt"><a href="#FragmentDefinition" data-name="FragmentDefinition">FragmentDefinition</a></span> (it must not be <span class="spec-nt"><a href="#TypeSystemDefinition" data-name="TypeSystemDefinition">TypeSystemDefinition</a></span>).</li>
 </ul>
+</section>
 <section id="sec-Executable-Definitions.Explanatory-Text" class="subsec">
 <h6><a href="#sec-Executable-Definitions.Explanatory-Text" title="link to this subsection">Explanatory Text</a></h6>
 <p>GraphQL execution will only consider the executable definitions Operation and Fragment. Type system definitions and extensions are not executable, and are not considered during execution.</p>
@@ -4084,7 +4086,8 @@
 <h4><span class="spec-secid" title="link to this section"><a href="#sec-Named-Operation-Definitions">5.2.1</a></span>Named Operation Definitions</h4>
[...]

PR #726, opened 2020-06-02
Formatting: use markdown lists in 'one of' lists

benjie:
`spec-md` recently [added support for bulleted "one of" productions](https://github.com/leebyron/spec-md/pull/51); this PR leverages this features to make the "one of" productions render similarly in regular markdown parsers (such as GitHub's) as they do in `spec-md`.

This is a precursor to formatting the spec with `prettier` (since, without this formatting, prettier will remove these "significant" line breaks).

---

Originally attempted with Markdown linebreaks, but there was significant pushback. ~~These whitespace changes make the `one of` lists render similarly in regular markdown parsers (such as GitHub's) as they do in `spec-md`.~~

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -59,13 +59,13 @@ NameContinue ::
   - Digit
   - `_`
 
-Letter :: one of
-  `A` `B` `C` `D` `E` `F` `G` `H` `I` `J` `K` `L` `M`
-  `N` `O` `P` `Q` `R` `S` `T` `U` `V` `W` `X` `Y` `Z`
-  `a` `b` `c` `d` `e` `f` `g` `h` `i` `j` `k` `l` `m`
+Letter :: one of  
+  `A` `B` `C` `D` `E` `F` `G` `H` `I` `J` `K` `L` `M`  
+  `N` `O` `P` `Q` `R` `S` `T` `U` `V` `W` `X` `Y` `Z`  
+  `a` `b` `c` `d` `e` `f` `g` `h` `i` `j` `k` `l` `m`  
   `n` `o` `p` `q` `r` `s` `t` `u` `v` `w` `x` `y` `z`
 
-Digit :: one of
+Digit :: one of  
   `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`
 
 IntValue :: IntegerPart [lookahead != {Digit, `.`, NameStart}]
@@ -313,25 +313,25 @@ DirectiveLocation :
   - ExecutableDirectiveLocation
   - TypeSystemDirectiveLocation
 
-ExecutableDirectiveLocation : one of
-  `QUERY`
-  `MUTATION`
-  `SUBSCRIPTION`
-  `FIELD`
-  `FRAGMENT_DEFINITION`
-  `FRAGMENT_SPREAD`
-  `INLINE_FRAGMENT`
+ExecutableDirectiveLocation : one of  
+  `QUERY`  
+  `MUTATION`  
+  `SUBSCRIPTION`  
+  `FIELD`  
+  `FRAGMENT_DEFINITION`  
+  `FRAGMENT_SPREAD`  
+  `INLINE_FRAGMENT`  
   `VARIABLE_DEFINITION`
 
-TypeSystemDirectiveLocation : one of
-  `SCHEMA`
-  `SCALAR`
-  `OBJECT`
-  `FIELD_DEFINITION`
-  `ARGUMENT_DEFINITION`
-  `INTERFACE`
-  `UNION`
-  `ENUM`
-  `ENUM_VALUE`
-  `INPUT_OBJECT`
+TypeSystemDirectiveLocation : one of  
+  `SCHEMA`  
+  `SCALAR`  
+  `OBJECT`  
+  `FIELD_DEFINITION`  
+  `ARGUMENT_DEFINITION`  
+  `INTERFACE`  
+  `UNION`  
+  `ENUM`  
+  `ENUM_VALUE`  
+  `INPUT_OBJECT`  
   `INPUT_FIELD_DEFINITION`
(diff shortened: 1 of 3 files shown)


IvanGoncharov:
@benjie I tried to run prettier on spec couple years ago and reported few issues:
https://github.com/prettier/prettier/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+author%3AIvanGoncharov+label%3Alang%3Amarkdown

I really like to use prettier but I don't want to add brittle and non obvious workarounds.
What I propose is to change `spec-md` to use code blocks for grammar rules:
image
Since I think it a generally bad idea to allow prettier to run on grammar rules.

benjie:
@IvanGoncharov I saw your eariler work, that's what inspired me to put the time in. I now have prettier running fully functionally against the GraphQL spec with a very tight and carefully reviewed diff thanks to some changes to `spec-md`.

Whether or not those changes get merged, this change helps to ensure the code is output in a similar way through Markdown as it is through spec-md, which is one of the goals of `spec-md`, I think, and makes the spec source easier to view.

I challenge the assertion that this is a "brittle and non obvious workaround"; this is [part of the original Markdown spec](https://daringfireball.net/projects/markdown/syntax#p):

> _When you do want to insert a `<br />` break tag using Markdown, you end a line with two or more spaces, then type return._

And is supported by all markdown processors I know of, including prettier. It's standard markdown formatting.

IvanGoncharov:
> I challenge the assertion that this is a "brittle and non obvious workaround";

It not visible in editors, for example here is your file in GitHub editor:
image
So how we can expect people to figure it out from a first try?

I still think grammar shouldn't be written in a pure markdown.
Can we trust that prettier doesn't change the current behavior in some other way that will break grammar?

benjie:
> Can we trust that prettier doesn't change the current behavior in some other way that will break grammar?

Yes. I've literally diffed the output HTML and it's identical except for a few small changes unrelated to grammar - you can read some about it in #727.

**However** this PR isn't specifically about prettier, just like #725 wasn't. I believe it has value in its own right - it is a formatting concern rather than a grammar concern.

To me it's a clear enhancement to the GraphQL spec source code -

[Spec output](https://spec.graphql.org/draft/#Letter):

Screenshot_20200602_151412

[Spec text in markdown before this change](https://github.com/graphql/graphql-spec/blob/master/spec/Appendix%20B%20--%20Grammar%20Summary.md#lexical-tokens):

Screenshot_20200602_151521

[Spec text in markdown after this change](https://github.com/benjie/graphql-spec/blob/format-one-ofs/spec/Appendix%20B%20--%20Grammar%20Summary.md#lexical-tokens):

Screenshot_20200602_151633

This last screenshot is clearly a lot closer to the intent. And the HTML generated by `spec-md` is byte-for-byte identical before and after this change. It's a small change that gives a significant enhancement at virtually zero cost.

IvanGoncharov:
>> Can we trust that prettier doesn't change the current behavior in some other way that will break grammar?

> Yes. I've literally diffed the output HTML and it's identical except for a few small changes unrelated to grammar - you can read some about it in #727.

I don't mean at the moment I meant that I worried about prettier changing formating in future releases.

As for the readability, I think code block is the closes option because it keeps indentation:
```
Letter :: one of  
  `A` `B` `C` `D` `E` `F` `G` `H` `I` `J` `K` `L` `M`  
  `N` `O` `P` `Q` `R` `S` `T` `U` `V` `W` `X` `Y` `Z`  
  `a` `b` `c` `d` `e` `f` `g` `h` `i` `j` `k` `l` `m`  
  `n` `o` `p` `q` `r` `s` `t` `u` `v` `w` `x` `y` `z`
```

And for readability inside PRs, I personally don't want to review changes where I need to pay attention to the trailing spaces.

Also, in the long run, we probably want to run some markdown linter (e.g. Vale) and we will have the same situation because it can't distinguish syntax rules from the rest of the text.

IvanGoncharov:
@benjie Pragmatically speaking without changes to `spec-md` your current solution is the best one.
So we need to find out if such changes in line with `spec-md` design and would be accepted.

@leebyron Can you please comment on what do you think is the best solution?

benjie:
I hear your concerns about whitespace, but it's a technical spec so I'm sure technical people can be disciplined enough to be careful with whitespace changes. When reviewing a diff, whitespace changes are shown in GitHub:

Screenshot_20200602_181419

When editing, whitespace can be seen via text selection:

Screenshot_20200602_181433

> I don't mean at the moment I meant that I worried about prettier changing formating in future releases.

Prettier, as far as I know, won't deliberately break the markdown rules (of which this is a fundamental one, defined before even headers are defined); but if it does then we can pin an earlier version.

> As for the readability, I think code block is the closes option because it keeps indentation [...] Pragmatically speaking without changes to `spec-md` your current solution is the best one. So we need to find out if such changes in line with `spec-md` design and would be accepted.

I personally prefer the more markdown-native approach over the codeblock escape-hatch, but you're much more familar with the GraphQL spec than I. Formatted productions like this seem to be a minority case, most productions can wrap happily* without any change to the output HTML. As you say, it's up to @leebyron to state what he wants for spec-md.

* _with my changes to `spec-md` which enable text to wrap_

alshamiri5 (approved):

leebyron:
I know we discussed this in person months ago, but for posterity (now that the solution is up as part of a spec-md release):

While trailing space line breaks are officially part of markdown, it's one of my very least favorite features for the reasons Ivan raised. Many tools and linters actually highlight trailing spaces as an error or editors that simply remove them (mine strips trailing whitespace on file save).

The fix as part of spec-md allows gridded one of productions to optionally use a bulleted list. It's not an ideal solution since the bullets could imply each row is its own entity which might be slightly confusing in raw markdown view, but the spec rendered result is the same. @benjie if you're up for reviving this one to use bullet lists instead, I think that will solve

benjie:
@leebyron Done; and have confirmed with diff that the spec renders identically (byte for byte) on `main` as on this branch. This should be suitable for merging :+1:

benjie:
(Musing: I bet if Markdown had used `[space][backslash]` or similar for their linebreak syntax rather than `[space][space]` this discussion would have gone a different direction.)

leebyron:
Nice! Happy to see this prettier-ready

PR #737, opened 2020-06-21
fix __Type explanation

dugenkui03:
Add 'input object types' in ` __Type` explanation.


IvanGoncharov:
@dugenkui03 Thanks 👍 
Merged 🎉

PR #740, opened 2020-06-25
Invalid JSON in @oneOf input union query

acelot:
I noticed that there are not enough curly brackets in the JSON.


binaryseed:
thanks!

PR #743, opened 2020-07-02
Deprecated input values

fotoetienne:
This PR rebases https://github.com/graphql/graphql-spec/pull/525 onto master with a bit of additional cleanup.

The intent is that this revision should now be suitable for promoting `Deprecated Input Values` to [Stage 2](https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md#stage-2-draft) RFC. To be discussed at the next working group meeting, [2020-07-02](https://github.com/graphql/graphql-wg/blob/master/agendas/2020-07-02.md).

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -321,24 +321,14 @@ input Point {
 
 Fields
 
-<<<<<<< HEAD
-- `kind` must return `__TypeKind.INPUT_OBJECT`.
-- `name` must return a String.
-- `description` may return a String or {null}.
-- `inputFields`: a list of `InputValue`.
-  - Accepts the argument `includeDeprecated` which defaults to {false}. If
-    {true}, deprecated enum values are also returned.
-- All other fields must return {null}.
-=======
 * `kind` must return `__TypeKind.INPUT_OBJECT`.
 * `name` must return a String.
 * `description` may return a String or {null}.
 * `inputFields`: a list of `InputValue`.
   * Accepts the argument `includeDeprecated` which defaults to {false}. If
-    {true}, deprecated fields are also returned.
+    {true}, deprecated enum values are also returned.
 * All other fields must return {null}.
 
->>>>>>> Update with suggested fixes
 
 #### List
 
@@ -374,13 +364,9 @@ Fields
 - `description` may return a String or {null}
 - `args` returns a List of `__InputValue` representing the arguments this
   field accepts.
-<<<<<<< HEAD
-- `type` must return a `__Type` that represents the type of value returned by
-=======
   * Accepts the argument `includeDeprecated` which defaults to {false}. If
     {true}, deprecated arguments are also returned.
 * `type` must return a `__Type` that represents the type of value returned by
->>>>>>> Update with suggested fixes
   this field.
 - `isDeprecated` returns {true} if this field should no longer be used,
   otherwise {false}.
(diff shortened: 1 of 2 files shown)


spec/Section 4 -- Introspection.md
```diff
@@ -143,7 +143,7 @@ type __Type {
   enumValues(includeDeprecated: Boolean = false): [__EnumValue!]
 
   # should be non-null for INPUT_OBJECT only, must be null for the others
-  inputFields: [__InputValue!]
+  inputFields(includeDeprecated: Boolean = false): [__InputValue!]
```
jturkel, line 146: I'm not sure if you're looking for feedback on this or #525 but won't defaulting to excluding deprecated input fields/field arguments result in problematic introspection results for older clients? For example, if I deprecate a field argument in my API, then the field argument will disappear in older versions of GraphiQL hitting my API which is not what I want and is inconsistent with how GraphiQL handles deprecated fields/enum values. 

Possible alternatives I can think of:

1. Default `includeDeprecated` to `true` for `__Type.inputFields` and `__Field.args` - not ideal since it's inconsistent with the default behavior of `__Type.fields` and `__Type.enumValues`.
1. Default `includeDeprecated` to `true` for `__Type.inputFields`, `__Field.args`, `__Type.fields` and `__Type.enumValues` which is a breaking change but won't actually break clients that explicitly specify `includeDeprecated`. Introspection queries generated by graphql-js explicitly set `includeDeprecated` it to true (see [here](https://github.com/graphql/graphql-js/blob/master/src/utilities/getIntrospectionQuery.js#L70)) but I'm not sure about other clients.
1. Don't support server-side filtering to include/exclude deprecated input fields/arguments and use this new functionality to deprecate the `includeDeprecated` argument for `__Type.inputFields` and `__Field.args`.
1. Introduce a new argument (e.g. `excludeDeprecated`) to  `__Type.inputFields`, `__Field.args`, `__Type.fields` and `__Type.enumValues` with defaulting that doesn't filter out deprecated members and deprecate the old `includeDeprecated` argument. Queries that specify both the old and new arguments would results in runtime errors.

IvanGoncharov:
@fotoetienne Thanks for rebasing 👍 
I think it's better to have all discussion in one PR so I just squash all the changes into a single commit and updated #525

spec/Section 4 -- Introspection.md
IvanGoncharov: @jturkel Don't think it's a problem since it will solve itself after you update GraphiQL.
If you have a concrete scenario in mind, can you please describe it here?

IvanGoncharov: Moreover, we constantly add new features that break old GraphiQL if they are used there, e.g. repeatable directives, support for interfaces implementing other interfaces, etc.

jturkel: In the GraphiQL scenario I was thinking of (i.e. an argument disappears when marked as deprecated in older version of GraphiQL) we only control the server. The API is public so we don't control the clients (which may include things like GraphiQL or Apollo Client) so we can't easily force them to upgrade. Ideally in that scenario older clients would continue to see deprecated arguments/input field but they just wouldn't know they were deprecated. 

I could be missing something but it looks like repeatable directives and interfaces implementing other interfaces made backwards compatible changes to the introspection system (i.e. older clients wouldn't know whether a directive was repeatable or if an interface implemented another interface) but servers using those new features wouldn't break existing clients that didn't know about them.

IvanGoncharov:
Superseded by #525

PR #744, opened 2020-07-02
Alternative input union workaround

jturkel:
This PR documents another input union workaround in the Input Union RFC. We use this approach in our GraphQL API at [Salsify](https://www.salsify.com/) and it looks like others do too based on the presence of the [graphql-union-input-type](https://github.com/Cardinal90/graphql-union-input-type) library with 3,430 weekly NPM downloads. I don't know how common this solution is in the wild vs. alternatives like "tagged unions" but I think it's relevant to consider a variety of existing workarounds in the "Input unions should be easy to upgrade from existing solutions" solution criteria.

/cc @binaryseed

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -65,7 +65,7 @@ input AnimalDropOffInput {
 }
 ```
 
-This allows non-sensical mutations to pass GraphQL validation, for example representing an animal that is both a `Cat` and a `Dog`.
+This allows nonsensical mutations to pass GraphQL validation, for example representing an animal that is both a `Cat` and a `Dog`.
 
 ```graphql
 mutation {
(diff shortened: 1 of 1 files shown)


rfcs/InputUnion.md
```diff
@@ -103,6 +103,30 @@ In addition, relying on this layer of abstraction means that this domain must be
...
+      {type: DOG, name: "Ripple", age: 2, breed: WHIPPET}
+    ]
+  )
+}
+
+input AnimalDropOffInput {
+  type: AnimalType!
+  name: String!
+  age: Int!
+  breed: DogBreed # only applies to dogs
```
binaryseed, line 123: How is the mapping between the field `breed` and `type: DOG` communicated?

Seems like this particular example is already possible, right?

jturkel: The information that the `breed` input field only applies when `type = DOG` and the `livesLeft` input field only applies when `type = CAT` is only conveyed via human readable documentation which is the major shortcoming with this approach. This schema information isn't conveyed in a machine, readable format that can be used by tooling to ensure clients only submit valid combinations of input fields for a given `type. It's really the same drawbacks as using "tagged unions" but with a slightly different literal shape.

On a somewhat related note, I'll update this example input type to include snake input fields to be consistent with the earlier example.

binaryseed: My second question is: how is this different than what you can already do with GraphQL as-is?

Tagged unions _do_ introduce something you can't do today & has a direct machine understandable mapping from type to fields...

jturkel: You can absolutely do this in GraphQL as-is! The proposed content is in the "Problem Sketch" section of the RFC which describes how people solve input polymorphism problems with GraphQL today and the shortcomings of these workarounds. The "tagged union" I was referring to is the current tagged union [workaround](https://github.com/graphql/graphql-spec/blame/master/rfcs/InputUnion.md#L61) and not the proposed [first class tagged union](https://github.com/graphql/graphql-spec/blame/master/rfcs/InputUnion.md#L778). Does that clear up any confusion?

binaryseed: oops, yeah it does clear it up - sorry the doc has gotten big and i assumed this was in the solutions section...

Issue #735, opened 2020-06-15
Standardize naming for "type/field pairs"?


magicmark:
_Note: This is a relatively minor nit, but thought it still worth a discussion_

For example, given this schema:

```graphql
type Business {
    reviews: [Reviews]
    name: String
}

type Review {
    body: String
    ...
}
```

- `Business.name` uniquely identifies the "name" field on the "Business" type
- `Business.reviews` uniquely identifies the "reviews" field on the "Business" type
- `Review.body` uniquely identifies the "body" field on the "Review" type

## Why do we want this?

Various tooling wants to show a "type/field pair" - usually for documentation or metrics/tracking purposes.

e.g. here we can see GraphiQL's documentation tab search bar showing type/field pairs to uniquely identify fields:

![](https://i.imgur.com/4BLxTZA.png)

Internally at Yelp, we want to show a list of "most popular fields in the schema" - using this key/value pair is a good way to track this!

## The ask

Being able to communicate what this strange dot separated string means would be useful to folks onboarding, and it'd be great to standardize this and point to something in the spec/official docs.

Does this worth standardizing and including in the docs?

Thanks!

benjie:
I agree with the drive for this. However, bikeshedding, we need to ensure that it wouldn't conflict with features we might add in future, such as [namespaces](#163). Of course we could just choose a different symbol for these relevant features, but the dot is particularly convenient.

Alternative symbol choices that might be suitable for type/field pair (or other features):

- `Foo::bar`
- `Foo#bar`
- `Foo->bar`
- `Foo~bar`

One advantage of `Foo.bar` is that it's already used in GraphiQL and thus effectively already the standard, and that may be sufficient to make it the official way.

andimarek:
As a meta comment: I think helping to communicate GraphQL concepts across ecosystems and implementation is a core responsibility of the GraphQL spec. Therefore standardizing this concept by giving it a clear name is a good goal. 

In GraphQL Java for example we call it "Field coordinates" and we print it as "Foo:bar". In another project the notation "Foo.bar" is used.

leoloso:
I very much like this proposal. I think using `Foo.bar` makes a lot of sense, since it is quite understandable. However, I wonder if the `.` symbol should be reserved for separating fields only, taking into account the Flat chain syntax (#174) feature, if it is ever approved. Then, using another symbol to separate the type from the field, maybe `:`, would be more understandable: 

`
Client:programs.shortName
`

Over:

`
[...]

PR #746, opened 2020-07-04
RFC: Schema Coordinates

magicmark:
Here's a stab at the RFC for https://github.com/graphql/graphql-spec/issues/735 - field coordinates

👉 Read in fancy format: https://github.com/magicmark/graphql-spec/blob/add_field_coordinates/rfcs/SchemaCoordinates.md
👉 See the proposed spec: https://github.com/graphql/graphql-spec/pull/794

## What does this RFC propose? 

As a quick reminder, given the following schema:

```graphql
type Person {
  name: String
}

type Business {
  name: String
  owner: String
}
```

- `Person.name` uniquely identifies the "name" field on the "Person" type
- `Business.name` uniquely identifies the "name" field on the "Business" type
- `Business.owner` uniquely identifies the "owner" field on the "Business" type

3rd party tooling uses this notation as a convention - however there is no official standardization (yet!) on what these are called, or how they're defined. We'll henceforth refer to these as "schema coordinates".

---

Looking forward to working with y'all to figure out if this seems worth advancing / hearing feedback / figuring out next steps.

Thanks!


rfcs/FieldCoordinates.md
```diff
@@ -0,0 +1,122 @@
...
+  ![](https://i.fluffy.cc/g78sJCjCJ0MsbNPhvgPXP46Kh9knBCKF.png)
+
+## 🗳️ Alternatives considered
+
+### Naming
+
+- "type/field pairs" was the original suggestion
+
+However, "field coordinates" was chosen as it is already understood and used by
+by the popular [GraphQL Java](https://github.com/graphql-java/graphql-java)
```
sungam3r, line 87:
```suggestion
the popular [GraphQL Java](https://github.com/graphql-java/graphql-java)
```

rfcs/FieldCoordinates.md
```diff
@@ -0,0 +1,122 @@
...
+project.
+
+### Seperator
```
sungam3r, line 90:
```suggestion
### Separator
```

rfcs/FieldCoordinates.md
```diff
@@ -0,0 +1,122 @@
...
+
+This RFC proposes using "`.`" as the seperator character. The following have also
```
sungam3r, line 92:
```suggestion
This RFC proposes using "`.`" as the separator character. The following have also
```

rfcs/FieldCoordinates.md
```diff
@@ -0,0 +1,122 @@
...
+been proposed:
+
+- `Foo::bar`
+- `Foo#bar`
+- `Foo->bar`
+- `Foo~bar`
+- `Foo:bar`
+
+"`.`" is already used in the existing implemenations of field coordinates, hence
```
sungam3r, line 101:
```suggestion
"`.`" is already used in the existing implementations of field coordinates, hence
```

rfcs/FieldCoordinates.md
```diff
@@ -0,0 +1,122 @@
...
+the suggested usage in this RFC. However, we may wish to consider one of the
+alternatives above, should this conflict with existing or planned language
+features.
+
+## 🤔 Drawbacks / Open questions
+
+- https://github.com/graphql/graphql-spec/issues/735 discusses potential
+  conflicts with the upcoming namspaces proposal - would like to seek clarity on
```
sungam3r, line 109:
```suggestion
  conflicts with the upcoming namespaces proposal - would like to seek clarity on
```

rfcs/FieldCoordinates.md
```diff
@@ -0,0 +1,122 @@
...
+  this
+
+- **Is this extensible enough?** The above issue discusses adding arguments as
+  part of this specifcation - we haven't touched on this here in order to keep
```
sungam3r, line 113:
```suggestion
  part of this specification - we haven't touched on this here in order to keep
```

spec/Appendix C -- Tooling Conventions.md
```diff
@@ -0,0 +1,103 @@
...
+This specification document formalizes naming and conventions that may be used by
+GraphQL tooling and third party libraries.
+
+
+## Field Coordinates
+
+FieldCoordinate : TypeName `.` FieldName
+
+A field coordinate may be used to uniquely identify a field in a GraphQL schema.
+It is encoded as a string containing a dot seperated parent type and field pair.
```
sungam3r, line 12:
```suggestion
It is encoded as a string containing a dot separated parent type and field pair.
```

spec/Appendix C -- Tooling Conventions.md
```diff
@@ -0,0 +1,103 @@
...
+
+**Examples**
```
sungam3r, line 14: Do you mind to write example for array?

magicmark: Could you elaborate? I think this should look the same for nullability/arrays etc - but I will add another example for clarity and avoidance of doubt. Is there something specific you were looking for?

thanks!

(What we do for arrays and nullability will get more complicated for serialization - still working on this section)

magicmark:
Thanks @sungam3r for the review! Sorry for the typos, have installed a spellcheck in vscode now :)

spec/Appendix C -- Tooling Conventions.md
sungam3r: It seemed to me that for arrays field coordinate should include an index. I might be wrong.

magicmark:
@IvanGoncharov Hey! This is my rodeo with the GQL RFC process - are there any more steps I can take for now? What should I be expecting at this stage? Thanks!

fotoetienne:
Would `InterfaceName.fieldName` also be a valid field coordinate, or would it need to be the concrete implementing type?

alshamiri5 (approved):

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,306 @@
...
+
+- https://github.com/graphql/graphql-spec/issues/735 discusses potential
+  conflicts with the upcoming namespaces proposal - would like to seek clarity on
+  this
+
+- Should we specify an algorithm for doing the query -> set of schema
+  coordinates? Or just hint/imply that this mapping theoretically exists? Is this
+  out of scope?
+
+### Answered quetsions
```
sungam3r, line 285:
```suggestion
### Answered questions
```

magicmark: good catch again, thanks!

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,310 @@
+# RFC: Schema Coordinates
+
+**Proposed by:** [Mark Larah](https://twitter.com/mark_larah) - Yelp
+
+This RFC proposes formalizing "Schema Coordinates" - a human readable syntax to
+uniquely identify a type, field, or field argument defined in a GraphQL Schema.
```
jturkel, line 6: Should enum value coordinates be addressed by this RFC too? Similar to the use cases discussed for fields and field arguments, we (Salsify) have tooling to understand how clients are using enum values (in addition to fields and field arguments) to support schema evolution. It should fit in pretty well with the proposed dot syntax (e.g. `Direction.NORTH` refers the `NORTH` enum value in the `Direction` enum) so hopefully it's not a big lift.

magicmark: yep - makes sense! thanks for this, incorporated this into the selector grammar.

magicmark:
I've started to turn this into a proper grammar defined in the spec. Just put this in a separate appendix file for now - we can discuss exactly where this should go tho.

The current draft gives a strong sense of direction of where i'm headed with this.

Still TODO:

- [ ] Document interfaces
- [ ] Figure out what to do about unions
- [ ] Figure out what to do about directives

Bikeshedding questions:

- How many examples should we show per selector type? I've shown 2 for most so far. We're showing how schema selectors disambiguate things, so showing both `Business.name` and `User.name` helps to communicate this. Also, I personally always appreciate more examples, and the cost here is cheap. Happy to pull this back to 1 tho - thoughts?
- Where should the schema selectors spec go? In an existing file? The new thing I made? Should we rename "Schema Selectors.md" to something more generic?

magicmark:
👋 @leebyron @benjie @IvanGoncharov howdy!

I think I'm happy with this as a first draft - ready for a review!

Cheers :)

spec/Appendix C -- Schema Selectors.md
```diff
@@ -0,0 +1,168 @@
...
+
+A Schema Selector is a human readable string that uniquely identifies an element defined in a GraphQL Schema.
+
+## Definition 
+
+SchemaSelector :
+  - TypeName FieldSelector?
+  - InterfaceName FieldSelector?
+  - EnumName EnumValueSelector?
+  - @ DirectiveName ArgumentSelector?
```
sungam3r, line 11: Why is the `@` symbol needed here? If I am not mistaken, then from the point of view of the GraphQL type system the directive should have a unique name among all other types. What is the point of using an additional syntax element in the selector for the directive?

magicmark: Good question - I had considered this, and i'm certainly open to changing this.

This was a purely stylistic/aesthetics choice. Given the use cases, seeing "@private" in the output of some tooling would more obviously indicate "this relates to a directive" to me.

(I realize we didn't make any such distinction for enums or interfaces, but we're making up the rules here.)

There's an existing built in hint for directive syntax (the @ symbol), so using it here too to mirror.

Hope this makes sense!

Again, if folks feel strongly, I'm open to changing this?

Thanks!

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+   **search results** when a user searches for a type or field name. We can
+   display a list of schema coordinates that match the search term. A schema
+   coordinate can also be used in the hyperlink to form a permalink for
+   documentation for a particular field.
+
+   _Existing implementations: GraphiQL, Apollo Studio (see "Prior Art")_
+
+1. A developer may want to perform **analytics** on all known
+   [persisted queries][apq] - e.g. what are the most commonly used fields across
+   all documents. Schema selectors may be used as the index/lookup keys when
```
sungam3r, line 43: You entered the `Schema Coordinates` term but here jumped to `Schema selectors`.

magicmark: 😱 yikes yes I had "selectors" littered about everywhere! (maybe that's a signal that selectors is a better name?! :P)

Thanks! Fixed!

sungam3r (commented):
> https://github.com/graphql/graphql-spec/pull/746#discussion_r502825198

I expected something like this. Let's hear what the others have to say.

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+    searchBusinesses(name: "El Greco Deli") {
+      name
+      owner {
+        name
+      }
+    }
+  }
+  ```
+
+  Person.name` in the response may as written as the following "field path":
```
sungam3r, line 235:
```suggestion
  `Person.name` in the response may as written as the following "field path":
```

sungam3r:
@magicmark  Because of force-push it is impossible to see what exactly you changed. 3 large files are displayed each time. This is the second (and last) time I read this large text.

sungam3r (approved):

magicmark:
> @magicmark Because of force-push it is impossible to see what exactly you changed. 3 large files are displayed each time. This is the second (and last) time I read this large text.

Oh yikes, sorry - was using `--amend` to keep the history clean. Wasn't thinking that that made things awkward to review. (Surprised github doesn't support this tbh? I'm used to reviewboard which does...)

Will switch to normal commits for future revisions.

magicmark:
@sungam3r actually looks like github does support this, you just have to punch it in manually

- https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons
- https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/comparing-commits#comparing-commits



e.g. for my last revision:

Screen Shot 2020-10-10 at 2 53 47 PM

you can see the incremental diff here https://github.com/graphql/graphql-spec/compare/a2c11ad..72438bf

sungam3r:
Nice to know but probably you understand that very few people will burden themselves with manually making a link each time to see the diff?

magicmark:
@sungam3r I think this is a 2 spaces vs 4 spaces thing. I'll go by whatever the repo's standard is, but in the absence of any, I default to rebasing and stuff to keep a clean history.

If you're regularly reviewing stuff, you may wish to find a browser extension (I'm sure one must exist) to linkifiy this for you.

In general, let's try an keep a positive attitude :)

benjie:
There’s no need to “keep a clean git history”; whilst reviewing it’s important to see the evolution of changes over time, and once approved it will be merged with “squash merge” so it’ll all be rolled up into one commit anyway 👍

sungam3r:
> it will be merged with “squash merge” so it’ll all be rolled up into one commit anyway

🤦  I had to say this from the very beginning, so that there was no misunderstanding.

magicmark:
That works too :)

Thanks!

magicmark:
Moved the sec edit to a new PR as suggested: https://github.com/graphql/graphql-spec/pull/794

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
+# RFC: Schema Coordinates
+
+**Proposed by:** [Mark Larah](https://twitter.com/mark_larah) - Yelp
+
+This RFC proposes formalizing "Schema Coordinates" - a human readable syntax to
+uniquely identify a type, field, or field argument defined in a GraphQL Schema.
```
benjie, line 6:
```suggestion
This RFC proposes formalizing "Schema Coordinates" - a human readable syntax to
uniquely identify a type, field, field argument, directive or directive argument defined in a GraphQL Schema.
```

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+
+This should be listed as a non-normative note in the GraphQL specification to
+serve as an official reference for use by third party tooling.
+
+## 📜 Problem Statement
+
+Third party GraphQL tooling and libraries may wish to refer to a field, or set of
+fields in a schema. Use cases include documentation, metrics and logging
+libraries.
```
benjie, line 15: GraphiQL isn't third party, and still wants to refer to this, so I've removed "Third party". Also made the intention broader.

```suggestion
GraphQL tooling and libraries may wish to refer to the various components of a GraphQL schema. Use cases include documentation, metrics and logging
libraries.
```

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+
+![](https://i.fluffy.cc/5Cz9cpwLVsH1FsSF9VPVLwXvwrGpNh7q.png)
+
+_(Example shown from GraphiQL's documentation search tab)_
+
+There already exists a convention used by some third party libraries for writing
+out fields in a unique way for such purposes. However, there is no formal
+specification or name for this convention.
```
benjie, line 23:
```suggestion
There already exists a convention used by some libraries and tools for writing
out fields in a unique way for such purposes. However, there is no formal
specification or name for this convention.
```

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+   display a list of schema coordinates that match the search term. A schema
+   coordinate can also be used in the hyperlink to form a permalink for
+   documentation for a particular field.
+
+   _Existing implementations: GraphiQL, Apollo Studio (see "Prior Art")_
+
+1. A developer may want to perform **analytics** on all known
+   [persisted queries][apq] - e.g. what are the most commonly used fields across
+   all documents. Schema coordinates may be used as the index/lookup keys when
+   storing this information in the database.
```
benjie, line 44: In this use case, it may be useful to differentiate between `Query.forums` (all, unfiltered), `Query.forums(about:)` (filtered) and `Query.forums(about:,first:,after:)` (filtered, paginated). No change required, just noting because this adds weight to the desire to add more than one argument to a field, though I understand that this is not technically a "schema coordinate" at this point but instead a usage summary.

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+```
+
+From the query above, we may calculate the following list of schema coordinates:
+
+- `Query.searchBusinesses`
+- `Business.name`
+- `Business.owner`
+- `Person.name`
+
+_`Query.searchBusinesses(name)` is also a valid member of the output set. The
```
benjie, line 176:
```suggestion
_`Query.searchBusinesses(name:)` is also a valid member of the output set. The
```

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+  type User {
+      name: String
+      reviewCount: Int
+      friends: [User]
+      email: String @private(scope: 'loggedIn')
+  }
+  ```
+
+  You _can_ select the definition of the `private` directive and its arguments
+  (with `@private` and `@private(scope)` respectively), but you cannot select the
```
benjie, line 304:
```suggestion
  (with `@private` and `@private(scope:)` respectively), but you cannot select the
```

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+
+## 🚫 What this RFC does _not_ propose
+
+- This does not cover "selectors" or "wildcard" syntax - e.g. `User.*`. _(See
+  alternatives considered.)_
+- There are **no proposed GraphQL language/syntax changes**
+- There are **no proposed GraphQL runtime changes**
+- [Schema coordinate non-goals](#-syntax-non-goals)
+
+## ✨ Worked Examples
```
benjie, line 77: Above this, we really should list the proposed syntaxes concretely rather than extracting them from examples. Maybe something like:

```suggestion
## Proposed syntax

### `Type`

Refers to a named type (e.g. something represented by `__Type` in GraphQL introspection).

### `Type.field`

Refers to a named "field" on the named type. Not all types support this. For object types and interface types this is a normal field, for input objects this would be input fields, for future GraphQL types this will relate to their concept of a field if they have one (e.g. for the [proposed "tagged" type](https://github.com/graphql/graphql-spec/pull/733) it would refer to the "member field".

### `Type.field(argName:)`

Refers to a named argument on the named field of the named type.

### `@directive`

References the given named directive

### `@directive(argName:)`

References the named argument of the named directive.

## ✨ Worked Examples
```

benjie (changes requested):
Looking good; I've gone through and tried to make the document more consistent after all the edits it's received :+1:

rfcs/SchemaCoordinates.md
benjie: (I'm not sure about "attribute", but couldn't think of a better word. Field isn't quite right because of enums.)

rfcs/SchemaCoordinates.md
eapache: or enum, as joel suggested

Also "type" is unclear whether that means specifically the formal output `type` or is a generic term that captures input types as well. Perhaps there's a broader category we could capture here rather than trying to list the specifics exhaustively?

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+out fields in a unique way for such purposes. However, there is no formal
+specification or name for this convention.
+
+### Use cases
+
+1. A GraphQL server wants to **log how often each field in the schema is
+   requested**. This may be implemented by incrementing a counter by the name of
+   the schema coordinate for each field executed in a request.
+
+   _Existing implementations: Yelp (internal), Facebook (internal)_
```
eapache, line 31: Shopify publicly via https://shopify.dev/concepts/about-apis/versioning/api-health.

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,372 @@
...
+
+- `Person` uniquely identifies the the "Person" type
+- `Business` uniquely identifies the the "Business" type
+- `Person.name` uniquely identifies the "name" field on the "Person" type
+- `Business.name` uniquely identifies the "name" field on the "Business"
+  type
+- `Business.owner` uniquely identifies the "owner" field on the "Business" type
+- `Query.searchBusinesses` uniquely identifies the "searchBusinesses" field on
+  the "Query" type
+- `Query.searchBusinesses(name:)` uniquely identifies the "name" argument on the
```
eapache, line 106: I'm still not keen on this syntax for arguments. `Query.searchBusinesses.name` feels a bit more natural to me, and is less likely to be mistaken for an actual query fragment or something. But I don't have a compelling argument here, it's mostly just personal preference.

rfcs/SchemaCoordinates.md
benjie: "type" is a very broad term, the SDL `type Foo { ... }` is called an "object type" in most places. I think "type" is sufficient to cover enums (and scalars, input object types, unions, interfaces, tagged, etc), but yes "enum values" does need to be added :+1: I've edited the above.

rfcs/SchemaCoordinates.md
benjie: If we stick with this syntax we can extend it in future to reference a position in a GraphQL document; e.g.

```
Query.businesses:searchBusinesses(name:).owner:personByOwnerId.email
```

might refer to:

```graphql
{
  businesses: searchBusinesses(name: "Automotive") {
    id
    name
    owner: personByOwnerId {
      id
      name
      email # <<< HERE
    }
  }
}
```

benjie: Actually that's a poor example (probably you'd just say `businesses.owner.email`); but you could write something like the above expression and have it automatically expand to a similar GraphQL document, similar to how you can use CSS shortcuts to expand to DOM structure using emmet or similar.

magicmark: @eapache Happy to go either way! I'm pretty 50/50 on `Foo.bar.baz` vs `Foo.bar(baz:)`. How do we decide here? Toss a coin? Dance-off battle?

@benjie  On the topic of something like `Query.businesses:searchBusinesses(name:).owner:personByOwnerId.email`, just to clarify:
1) Schema coordinates are supposed to "uniquely identify a type". Given our use cases, it's ideal if there's no normalization that has to be done to get to the "simplest" schema coordinate for the same thing.
2) Tracing the path to a member of a query/document is kinda already a thing with [field paths](https://spec.graphql.org/June2018/#example-90475) (`"path": [ "hero", "heroFriends", 1, "name" ]`).

Perhaps it's possible one day we unify this syntax and expand schema coordinates to support nesting, but I'll gingerly suggest that this is out of scope for this particular RFC? (Worth bearing in mind for a future extension of this syntax tho, along with selecting multiple arguments)

Thoughts?

eapache: > Dance-off battle?

I'm in 😂 

Seriously though, I think it's probably something worth chatting about at the IRL working group meetings. There might be an obvious consensus there, and if not then we'll probably want to do a slightly more formal pros/cons writeup to justify making a final call (basically the process we're following for input unions, but not nearly so heavy).

> businesses.owner.email

Yeah, the schema-coordinates vs query-path vs result-path distinction is still a little weird... they're definitely distinct but they *feel* like they ought to be related somehow.

Case in point, in the Shopify example I shared above we're technically sharing a weird combination of coordinate and path. For use of deprecated fields we share the plain coordinate `Type.field` but for use of deprecated inputs we append the full input path (so not `InputType.field` but actually `Type.field.argument.inputFieldPath...`.

magicmark:
(Just signed the CLA)

rfcs/SchemaCoordinates.md
magicmark: Makes sense - thanks!

rfcs/SchemaCoordinates.md
magicmark: ack! Ok leaving this open for now. I'll proceed with the `Foo.bar(baz:)` style just so we have something written down, but we'll make sure to talk about it next WG meeting. Happy to ctrl-f search and replace it all to `Foo.bar.baz` if that's where we land - should be an easy to change to make now that we're prepared for it.

I'll add this as an open question to the RFC.

Thanks!

benjie: > Perhaps it's possible one day we unify this syntax and expand schema coordinates to support nesting, but I'll gingerly suggest that this be out of scope for this particular RFC? (Worth bearing in mind for a future extension of this syntax tho, along with selecting multiple arguments)

100%; I was never suggesting adding this to this spec, more that we should make sure that the spec enables future innovation including using it for things _other than_ schema coordinates. To my mind the `(foo:)` syntax is more flexibile in this regard. For example, I can imagine referring to:

1. `Foo.bar(baz:.qux)`: the `qux` field of the input object referred to from the `baz` argument of the `bar` field on the `Foo` type.
2. `Foo.bar(baz:).qux`: the `qux` field on the return type of the `bar` field (with `baz:` argument) of the `Foo` type.
3. `Foo.bar.baz.qux`: the `qux` field of the return type of the `baz` field on the return type of the `bar` field on type `Foo`.

If we were to only use periods then all of these would come out the same as `Foo.bar.baz.qux`, and this ambiguity precludes this kind of reusal of the schema-coordinates syntax for this use case (which is outside the scope of the schema coordinates spec, for sure, but is still a potential use-case for the syntax).

Ultimately what I'm saying is I feel it would be a shame if the syntax could not be re-used outside of schema coordinates.

benjie:
@brianwarner Should the EasyCLA comment have been automatically updated in this case?

rfcs/SchemaCoordinates.md
eapache: > `Foo.bar(baz:).qux`

I would rewrite this all-periods as `Foo.bar.quz`. I don't see why the argument is relevant if the thing you're trying to reference is the `qux` field. Assuming a well-formed query then you don't even need the argument to uniquely identify the path.

benjie: Sure, if we're using them as coordinates rather than doing Emmet-style auto-completion or some other use case someone has come up with. Even as simply coordinates though, 1 and 3 are still distinct yet would have the same representation with periods?

Imagine you had the following schema

```graphql
type Foo {
  bar(baz: BazInput): Bar
}

type Bar {
  baz(qux: Int): Baz
}

type Baz {
  qux: Int
}

input BazInput {
  qux: Int
}

type Query {
  foo: Foo
}
```

With periods you might write `Foo.bar.baz.qux`, and that could refer to (using argument syntax) `Foo.bar(baz:.qux)` (input field), `Foo.bar.baz(qux:)` (argument) or `Foo.bar.baz.qux` (field) - three very different things. GraphQL query language looks like the JSON it returns, it makes sense to me that schema coordinates should look like the GraphQL entities they represent.

magicmark: I've added a couple notes in the RFC about this syntax:

1. ```Should arguments look like `Foo.bar.baz` instead of `Foo.bar(baz:)`?``` (we'll discuss in a WG meeting)
2. As an RFC goal: `Schema coordinate syntax is open for extension in the future.` To @benjie (and other's) points, whilst we're not currently supporting fancy extended syntax as described, we could all imagine extending schema coordinate syntax to support multiple arguments, wildcards etc. So if we can do something now even just to give us more flexibility and wiggle room in the future, then that's probably a good thing.

rfcs/SchemaCoordinates.md
magicmark: added, thanks!

rfcs/SchemaCoordinates.md
benjie: Following yesterday's WG (and the thumbs up above) I believe that @eapache is happy to go ahead with the parenthesized syntax.

Additional note; I wrote `Foo.bar(baz:.qux)` above; but I think `Foo.bar(baz.qux:)` has better aesthetics.

benjie:
Slightly off topic; other ideas for extensions that might build on this syntax:

- Permalinks for GraphiQL Documentation sidebar `?docs=User.friends>latestMedia>Post.title` (the `Post.` here might represent picking a type from a union/interface; we could use any number of syntaxes for this so don't focus too much on that :wink: )
- Linking from a field's deprecation reason to the new path to use for this property "The `Query.branchesFromFork` field is being removed; please use the following path instead: `Query>repositories>forks>branches(matching:)`"
- Indicating how to navigate to a specific schema coordinate via a path; e.g. when viewing the documentation for `User.firstName` GraphiQL might indicate the shortest paths it can be accessed from:
  - `Query.me>firstName`
  - `Query.articles>author>firstName`
  - `Query.searchMedia>Book.author>firstName`
- Tracking usage counters for each schema coordinate, and the path that's used to access it: `counters['Query.cities>libraries>findBook(isbn:)']++`

rfcs/SchemaCoordinates.md
cheapsteak: Bit of a tangent (although it seems like using parens is resolved so it may be safe) - 

In the example of `Foo.bar(baz:).qux`, is the contents of the parens carrying useful information for the coordinates?  

Assuming arguments require parens, would `Foo.bar(baz:).qux` point to the same place as `Foo.bar.qux`?

excitedbox: Is it too late to add my opinion? I agree with @eapache the C struct.member notation is much better for several reasons. 
- 1. Exactly that, it is C style syntax so it comes naturally for many devs who work with C style languages such as php. 
- 2. A period is easier and quicker to type than (:). 
- 3. It is 3 non alpha numeric characters instead of 1, which makes a 3 byte difference AND requires a parser while the struct syntax could be implemented in C as a struct without parsing. 
 
This also makes a big difference in embedded and IOT systems which can benefit immensely from graph data structures if they are implemented efficiently. This could literally keep me from using graphQL in a project.

Lastly why call it coordinates? It doesn't have anything to do with position or am I missing something. This constant reinvention of terms in web frameworks has always been a major issue for newcomers. Things should be named for clarification and not because they sound fancy. It is not descriptive of it's purpose. It is used to identify just as it was described in the OP. So why not call it a Field Identifier? Field Descriptor? It may not sound hip or new but I know what it means.

If you go up to non GraphQL literate person you should be able to explain a query without having to explain what a "field coordinate" is after using the word. If you were to ask me to identify the coordinate in those queries I would think you are trolling me and I have been a web dev for 15+ years. I just asked my GF what the Descriptor was and she said "the name after the period". She is a English teacher and doesn't know what a div tag is.

acao:
worth noting that the GraphiQL docs component already uses a simple dot notation internally for state, so we added url arguments for this to the 2.x rewrite this spring!

also, we authored the hover dot notation that apollo uses in studio, it comes from monaco-graphql/the LSP

rfcs/SchemaCoordinates.md
magicmark: I've had some time to chew this over - given discussions last WG meeting, I'm personally pretty settled on `Foo.bar(baz:)`. That's the option I'm proposing in the RFC.

I've captured this discussion in the RFC here: https://github.com/magicmark/graphql-spec/blob/add_field_coordinates/rfcs/SchemaCoordinates.md#field-arguments

Please let me know if anyone feels I missed anything, or if there's more points to add (or existing points to contest!)

Thanks everyone who has helped so far in thinking through this :)

@excitedbox to answer some stuff:

1. On the topic of argument syntax, it is a bit of a tradeoff! I've tried to capture the reasoning in a [new section in the RFC](https://github.com/magicmark/graphql-spec/blob/add_field_coordinates/rfcs/SchemaCoordinates.md#field-arguments). I welcome your followup feedback!
2. For naming, I'll point again to the RFC, as there's [a whole section on this](https://github.com/magicmark/graphql-spec/blob/add_field_coordinates/rfcs/SchemaCoordinates.md#naming). I will say that in the last WG meeting, I literally bought up the point "schema coordinates sounds a bit fancy", but I think (1) ultimately it's a more accurate name than the alternatives I had, and (2) it's reusing prior art (graphql java already used the "coordinates" naming). If you have an alternate suggestion, I'd love to hear your thoughts.

magicmark:
@acao could you clarify what is "hover dot notation"?

magicmark:
I've updated the RFC to capture discussion about field arguments.

I think this was the main thing we had to square away from the last WG meeting.

I think we're in the "dotting the I's" phase at this point, so if there's any lingering feedback (high level, or on the specifics of what's being proposed), I'd love to hear any thoughts, so we can move towards consensus and ultimately merge this.

(cc @benjie @eapache @IvanGoncharov as folks who have specifically expressed interest in this)

Thanks to everyone who has looked at this and given feedback so far :)

rfcs/SchemaCoordinates.md
magicmark: ++ something like `Query.forums(about:,first:,after:)` sounds like a great use case / future syntax extension.

Going to close this for now as it looks like there's no change needed, but let's certainly bear this in mind for the future!

rfcs/SchemaCoordinates.md
magicmark: (I'm going to leave this conversation thread open till at least the next WG to make sure this discussion remains visible for folks)

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,524 @@
...
+Using dots only as a separator may overload the meaning of elements in schema
+coordinates in the future. (If we capture new schema node types, or nested
+paths.)
+
+> We should make sure that the spec enables future innovation including using it
+> for things other than schema coordinates. To my mind the (foo:) syntax is more
+> flexible in this regard. For example, I can imagine referring to:
+>
+> 1. `Foo.bar(baz:.qux)`: the qux field of the input object referred to from the
+>    baz argument of the bar field on the Foo type.
```
benjie, line 375:
```suggestion
> 1. `Foo.bar(baz.qux:)`: the qux field of the input object referred to from the
>    baz argument of the bar field on the Foo type.
```

Would retro-editing this to the slightly modified syntax I now propose be appropriate? I've edited the linked comment with a note.

rfcs/SchemaCoordinates.md
```diff
@@ -0,0 +1,524 @@
...
+- Should we specify an algorithm for doing the query -> set of schema
+  coordinates? Or just hint/imply that this mapping theoretically exists? Is this
+  out of scope?
+
+### Answered questions
+
+- **Is this extensible enough?** The above issue discusses adding arguments as
+  part of this specification - we haven't touched on this here in order to keep
+  this RFC small, but we may wish to consider this in the future (e.g.
+  `Query.searchBusiness:name`).
```
benjie, line 508: Let's strikethrough this since the update below notes it's no longer relevant. Alternatively remove this bullet entirely.

```suggestion
- **Is this extensible enough?** ~~The above issue discusses adding arguments as
  part of this specification - we haven't touched on this here in order to keep
  this RFC small, but we may wish to consider this in the future (e.g.
  `Query.searchBusiness:name`).~~
```

benjie (approved):
LGTM :+1:

rfcs/SchemaCoordinates.md
magicmark: ack - it's under the sercion "Answered questions" which implies it's old news, but i'll just remove this point entirely as you suggest, since i guess it's superfluous as it's now explicitly now covered in the spec

benjie (approved):

PR #751, opened 2020-07-21
RFC: Scalar serialize as built-in scalar type

akshatj355:
Currently, Custom Scalars only describe their name and no other behavior about them can be programmatically determined. However in practice most custom scalar types are specializations of one of the existing built-in scalar types (String, Int, Float, etc.).

This proposes adding one additional piece of metadata to custom scalars, a "serializes as" type which must be a built-in scalar type. This serialize as type describes what kinds of values are allowed to be provided as input to that type and what kind of value is expected to serialize from that type. This metadata is useful for code-generation systems which expect to generate code in well typed environments. It can also improve the validation accuracy and IDE ergonomics by providing a similar level of information about custom scalars as exist for the built-in scalars.

To provide this metadata to those kinds of tools, this also extends the introspection system to return the serialize as type in the `ofType` field.
I can try for the same.

!!! IMPORTANT !!!

Please Read https://github.com/facebook/graphql/blob/master/CONTRIBUTING.md before creating a Pull Request.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -76,13 +76,7 @@ response format being used, there may be a more appropriate primitive for the
 given scalar type, and server should use those types when appropriate.
 
 GraphQL provides a number of built-in scalars, but type systems can add
-additional scalars with semantic meaning. For example, a GraphQL system could
-define a scalar called `Time` which, while serialized as a string, promises to
-conform to ISO-8601. When querying a field of type `Time`, you can then rely on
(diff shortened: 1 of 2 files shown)


eapache:
Whether a custom scalar serializes like an existing scalar type or not seems highly dependent on the transport layer, which can vary.

sungam3r:
I think that #649 more fully solves the problem of describing scalars.

IvanGoncharov:
Duplicate of #521

PR #761, opened 2020-07-29
Add missing VARIABLE DEFINITION

dugenkui03:
add missing `VARIABLE DEFINITION`.


sungam3r:
Interesting. There may be problems with parsers.

IvanGoncharov:
@sungam3r It already exists in grammar: http://spec.graphql.org/draft/#ExecutableDirectiveLocation
we just forgot to add it to instrospection.

PR #767, opened 2020-08-14
first commit

taru-svg:
First contribution: README.md file update


IvanGoncharov:
@taru-svg Thanks for taking the time to prepare this PR 👍 
But it changes the meaning without providing explanation for such changes.

PR #733, opened 2020-06-12
[RFC] Tagged type

benjie:
**THIS RFC HAS BEEN SUPERSEDED** by `@oneof`, for now at least... See: https://github.com/graphql/graphql-spec/pull/825

---

This is an RFC for a new "Tagged type" to be added to GraphQL. It replaces the ["@oneField directive"](https://github.com/graphql/graphql-spec/pull/586) proposal following feedback from the Input Unions Working Group. Please note that "Tagged type" is the working name, and may change if we come up with a better name for it.

A Tagged type defines a list of named members each with an associated type (like the fields in Object types and Input Object types), but differs from Object types and Input Object types in that exactly one of those members must be present.

The aim of the Tagged type is to introduce a form of polymorphism in GraphQL that can be symmetric between input and output. In output, it can generally be used as an alternative to Union (the differences will be outlined below). It goes beyond interfaces and unions in that it allows the same type to be specified more than once, which is particularly useful to represent filters such as this pseudocode `{greaterThan: Int} | {lessThan: Int}`.

If merged, Tagged would be the first non-leaf type kind (i.e. not a Scalar, not an Enum) that _could_ be valid in both input and output. It is also the first kind of type where types of that kind may have different input/output suitability.

In SDL, a tagged type could look like one of these:

```graphql
# suitable for input and output:
tagged StringFilter {
  contains: String!
  lengthAtLeast: Int!
  lengthAtMost: Int!
}

# output only:
tagged Pet {
  cat: Cat!
  dog: Dog!
  colony: ColonyType!
}

# input only:
tagged PetInput {
  cat: CatInput!
  dog: DogInput!
  colony: ColonyType!
}
```

(Note a [number of alternative syntaxes](https://gist.github.com/benjie/e45540ad25ce9c33c2a1552da38adb91) were mooted by the Input Unions working group; the one above was chosen to be the preferred syntax.) 

If we queried a `StringFilter` with the following selection set:

```graphql
{
  contains
  lengthAtLeast
  lengthAtMost
}
```

then this could yield one of the following objects:

- `{ "contains": "Awesome" }`
- `{ "lengthAtLeast": 3 }`
- `{ "lengthAtMost": 42 }`

Note that each of these objects specify exactly one key.

Similarly the above JSON objects would be valid input values for the `StringFilter` where it was used as an input.

## Tagged vs Union for output

Tagged does not replace Union; there are things that Union can do that tagged cannot:

```graphql
{
  myUnionField {
    ... on Node {
[...]

PR #773, opened 2020-09-02
Make 'Field Selections' header stable

benjie:
In #733, @leebyron [suggested renaming this header](https://github.com/graphql/graphql-spec/pull/733#discussion_r466619908) to simply 'Field Selections' with the line below it. I assume this is to increase the stability of this header such that permalinks to it will be stable.

It makes sense to me to introduce this change as early as possible, so I've separated it out into this PR.

Note: this also fixes a gramatical issue, it should read `on Objects, Interfaces, and Unions` or `on Object, Interface and Union Types`; the mixed plural currently seems incorrect.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -324,7 +324,9 @@ must provide the operation name as described in {GetOperation()}.
 
 ## Fields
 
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
Thanks @benjie 👍

PR #774, opened 2020-09-02
Add FAQ about @defer on fragments vs fields

robrichard:


rfcs/DeferStream.md
```diff
@@ -159,6 +159,11 @@ The GraphQL spec does not currently support object identification or consistency
 ### Can @defer/@stream increase risk of a denial of service attack?
 This is currently a risk in GraphQL servers that do not implement any kind of query limiting as arbitrarily complex queries can be sent. Adding `@defer` may add some overhead as the server will now send parts of the query earlier than it would have without `@defer`, but it does not allow for any additional resolving that was not previously possible.
 
+## Frequently Asked Questions
+
+### Why is `@defer` supported on fragments instead of fields?
+If there is a UI component that renders many fields which are deferred, it could be cumbersome to coordinate the loading state of all of those fields. By deferring all of the fields on a fragment, the component can render its fallback loading state until the fragment is loaded, without having to manage the individual state of many fields. If only a single field needs to be deferred, it can be wrapped in an inline fragment. Since there is an easy workaround, we do not plan to support `@defer` on an individual field as part of this proposal.
```
IvanGoncharov, line 165: @robrichard Same can be said for `@skip` and `@include`. 
I actually think we need to support it in fields to be consistent.
Especially since nothing prevents you from wrapping every individual component into separate fragments.

robrichard:
I'm going to add more background here as discussed in the WG meeting

rfcs/DeferStream.md
josephsavona: Note that it isn't possible to defer delivery of a field _value_ - to have the key present in the response but not its value. Therefore, the actual payload that has to be sent in this case - along with the label and path - must always be the same as if the field selection was wrapped in a deferred inline fragment. For example consider the following query, where both fields (id, name) are typed as non-null in the schema:

```
// HYPOTHETHICAL 
{
  user {
    id
    name @defer(...)
  }
}
```

There is no way to defer only the _value_ of `name`, ie it is invalid to send a set of responses such as that below, since there is no legal value to send in the first payload for `name`:

```
{
  "data": {
    "user": {
      "id": "<userid>"
      "name": ???? // <--- no legal value here, name is non-nullable
    },
}
{
  "data": "<username>",
  "path": ["user", "name"],
  "label": "...."
}
```

Instead, this must be treated as if the query was:

```
{
  user {
    id
    ... @defer(...) {
     name
    }
  }
}
```

in which case the server can return a valid set of responses - note that `data` and `path` are different!:

```
{
  "data": {
    "user": {
      "id": "<userid>"
      // note 'name' not present here
    },
}
{
  "data": {
    "name": "<username>",
  },
  "path": ["user"],
  "label": "...."
}
```

I'm not necessarily opposed to allowing `@defer` on fields, but i do think it's worth considering that implementation-wise they must always behave as if the user had deferred a parent inline fragment. If you're trying to work with a raw response, this likely isn't obvious and could lead to confusion.

benjie (approved):

andimarek:
@benjie @robrichard can we merge that?

robrichard:
@andimarek yes this can be merged

PR #775, opened 2020-09-04
__Type represents all types in the system

benjie:
In https://github.com/graphql/graphql-spec/pull/733#discussion_r466615699 @leebyron noted:

> maybe we should just say: "It represents all types in the system."

I wasn't quite comfortable with what "all types" meant - whether it was inclusive or exclusive of modified types. The second paragraph implied exclusive due to the `also` but this didn't sit quite right with me. At first I edited it to say "all _named_ types in the system", but felt it made the second paragraph feel a bit disjoint.

I've further modified the text and feel it's much clearer now.

I'm trying to move discussions like this out of the Tagged type spec itself, to keep the signal to noise ratio high (and figure they're useful mods for any future type additions anyway, keeping maintenance burden down).

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -218,10 +218,10 @@ enum __DirectiveLocation {
 ### The __Type Type
 
 `__Type` is at the core of the type introspection system, it represents all
-types in the system: both named types and type modifiers.
-
-Type modifiers, which are used to modify a referenced type (`ofType: __Type`),
-are how we represent lists, non-nullable types, and the combinations thereof.
+types in the system: both named types (e.g. Scalars and Object types) and 
+type modifiers. Type modifiers are used to modify a referenced type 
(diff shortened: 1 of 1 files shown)


benjie:
Incidentally we consistently refer to "type modifier" not as the concept of `List`/`NonNull` itself, but as a specific instance of these concepts applied to a specific type; I've tried to maintain this in this PR. I'd prefer to call `List`/`NonNull` themselves "type modifiers", and `List(NamedType)` a "modified type" - if this seems reasonable I'm happy to make the relevant edits.

benjie:
@leebyron Could I get your eyes on this?

leebyron:
Thanks for the reminder. I tagged it so I don't lose it :)

Let me give this a closer read when I get a free moment

Muftaa:
-0--popk

Muftaa (approved):

benjie:
@leebyron Gentle nudge :wink:

benjie:
LGTM @leebyron :+1:

PR #776, opened 2020-09-14
RFC: __typename is not valid at subscription root

benjie:
I'm writing up some of the GraphQL "query ambiguity" work currently and noticed this issue in the spec. In the case of a subscription it seems that

```graphql
subscription {
  __typename
}
```

does not correctly get evaluated by the reference implementation during subscriptions (though you can query it with `graphql(...)` you can't access it via `subscribe(...)`). It doesn't really make sense to request `__typename` here since it's not ever going to change during the life of the schema, and subscriptions only support one root-level field, however I thought it was worth highlighting.

Reproduction with GraphQL-js (toggle which `doc` is commented to see a functioning subscription):

```js
const {
  GraphQLSchema,
  GraphQLObjectType,
  GraphQLInt,
  GraphQLFloat,
  parse,
  validate,
  subscribe,
} = require("graphql");

const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));

async function* everySecond() {
  for (let i = 0; i < 1000; i++) {
    yield i;
    await sleep(1000);
  }
}

  name: "Query",
  fields: {
    a: {
      type: GraphQLInt,
    },
  },
});

const Subscription = new GraphQLObjectType({
  name: "Subscription",
  fields: {
    ts: {
      type: GraphQLFloat,
      subscribe() {
        return everySecond();
      },
      resolve() {
        return Date.now();
      },
    },
  },
});

  query: Query,
  subscription: Subscription,
});

async function main() {
  const doc = parse("subscription { __typename }");
  // const doc = parse("subscription { ts }");
  const errors = validate(schema, doc);
  if (errors && errors.length) {
    console.dir(errors);
    throw new Error("Errors occurred");
  }
  const result = await subscribe(schema, doc);
  console.dir(result);
  for await (const r of result) {
    console.dir(r);
  }
}

main().catch(e => {
  console.dir(e);
  process.exit(1);
});
```

Produces this error:

```
Error: Subscription field must return Async Iterable. Received: undefined
    at [...]/node_modules/graphql/subscription/subscribe.js:169:13
    at async main ([...]/test.js:57:18)
```

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -98,7 +98,8 @@ identify which actual type of the possible types has been returned.
 
 This field is implicit and does not appear in the fields list in any defined type.
 
-Note: `__typename` may not be used as a root field in a subscription operation.
+Note: `__typename` may not be included as a root field in a subscription
+operation.
 
 ## Schema Introspection
 
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -89,9 +89,10 @@ warnings.
 
 ## Type Name Introspection
 
-GraphQL supports type name introspection at any point within a query by the
+GraphQL supports type name introspection at any point within an operation by the
 meta-field `__typename: String!` when querying against any Object, Interface,
-or Union. It returns the name of the object type currently being queried.
+or Union; with the single exception of the subscription root operation type. It
+returns the name of the object type currently being queried.
 
 This is most often used when querying against Interface or Union types to
 identify which actual type of the possible types has been returned.
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -98,8 +98,7 @@ identify which actual type of the possible types has been returned.
 
 This field is implicit and does not appear in the fields list in any defined type.
 
-Note: `__typename` may not be used as a root field in a subscription operation
-because it does not return an async iterable.
+Note: `__typename` may not be used as a root field in a subscription operation.
 
 ## Schema Introspection
 
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -89,16 +89,17 @@ warnings.
 
 ## Type Name Introspection
 
-GraphQL supports type name introspection at any point within an operation by the
+GraphQL supports type name introspection at any point within a query by the
 meta-field `__typename: String!` when querying against any Object, Interface,
-or Union; with the single exception of the subscription root operation type. It
-returns the name of the object type currently being queried.
+or Union. It returns the name of the object type currently being queried.
 
 This is most often used when querying against Interface or Union types to
 identify which actual type of the possible types has been returned.
 
 This field is implicit and does not appear in the fields list in any defined type.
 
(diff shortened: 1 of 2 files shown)


AugustoCalaca (approved):

leebyron:
Exception doesn't need to be as prominent since this sentence first introduces the idea of `__typename`. Maybe a `Note:` at the bottom of the section you're editing

Validation rule should change to "one *non introspection* field"

Example that needs to change https://spec.graphql.org/draft/#example-2353b

benjie:
GraphQL.js PR here: https://github.com/graphql/graphql-js/pull/2861

yaacovCR:
It would be helpful I think if this would be valid to allow for a completely typed result to be passed to result transformers. Otherwise result transformers have to be passed additional state in terms of which operation type was sent and treat the data field differently than inner fields during processing, at least for subscriptions.

yaacovCR:
Of course, you can save the request document and add __typename to all abstract fields and then combine request and result to know types of everything, including subscription root. That is another way. But seems like this way should also be allowed to work in terms of symmetry of the root types.

yaacovCR:
Of course, then you need to repeat the same field collection algorithm that happened during field resolution on the server when processing the result, so we currently do that in graphql-tools only when we need to know the types of scalar fields. Perhaps a generic solution would be to annotate fields using aliases that would include the typename, that is yet another option, just would be easier I think if all of these options worked...

eapache:
FWIW I checked how graphql-ruby handles this, and it Just Works (I suppose in contravention of the spec). If you put a `__typename` in a subscription request you get that back (value `"Subscription"`) beside your payload in the initial response and in every subsequent update.

benjie:
@IvanGoncharov I've written up a solution that follows your suggestion (and might be how `graphql-ruby` works, @eapache?) in https://github.com/graphql/graphql-spec/pull/806.

I'm on the fence between these two solutions. In general I'm leaning towards the simplicity of this one ("there can only be one field, and it can't be __typename") but #806's expansion of `__typename` to all contexts is also attractive, particularly for tooling as @yaacovCR notes.

My main hesitation against #806, beyond complexity, is that it feels like it's going to make the "subscriptions must only have one root-level field" rule muddier ("if we can have a normal field and an introspection field, why can't we have another normal field?").

Interestingly if we merged this PR, we could later merge #806 and it would be a non-breaking change (an expansion of capabilities). Both invalidate the operation `subscription { __typename }`, so if there's no clear consensus that #806 is superior I feel we should merge this PR to address the immediate issue, and then continue discussions about potentially merging #806 later.

leebyron:
Looks like some of this language already leaked in via editorial changes in #844 - https://spec.graphql.org/draft/#sel-FAJZDABzCBjF8wX

benjie:
I think this is fine to merge, it doesn't prevent us changing our mind in future :+1: 

One caveat: I think graphql-ruby (?) allows `__typename` in addition to the subscription field, but I think this just means it has "more lenient validation" rather than being specifically non-compliant?

leebyron:
I know PR is still in review, but I think we're good here to go to RFC 3:

- [x] Consensus the solution is complete (via *editor* or working group)
- [x] Complete spec edits, including examples and prose
- [x] Compliant implementation in GraphQL.js (fully tested and merged or *ready to merge*)

Issue #715, opened 2020-04-29
The term "query" is overloaded and can be ambiguous


benjie:
We have a general issue in the GraphQL ecosystem that the term "query" is overloaded - sometimes it refers to the query operation type or a query operation, sometimes it refers to a GraphQL request or a GraphQL document containing operations, sometimes it refers to the contents of selection sets (as in "query reuse"), and sometimes it used as a verb to refer to the generic act of querying a server for information.

This is exacerbated by us using the term "query" in the communication with the server via `{query, variables, operationName}`. Arguably `query` should be `document` here since it may contain multiple operations, and those operations might be of different operation types (query, mutation, subscription). I believe this ship may have already sailed, however, and may not be worth the churn of changing this particular usage now.

---

The spec tends to be fairly disciplined in referencing query currently; it uses the following phrases:

- 'query operation' (an operation of the query type)
  - >   * If {operation} is a **query operation**:
- 'query error' (an error during execution of an operation, applies to all operation types)
  - > If an input value does not match a coercion rule, a **query error** must be raised.
- 'query reuse' (refering to reusing selection sets, e.g. within fragments)
  - > A document may contain operations (queries, mutations, and subscriptions) as well as fragments, a common unit of composition allowing for **query reuse**.
- 'query keyword' (literally the keyword `query` in a query operation)
  - > If a Document contains only one operation, that operation may be unnamed or represented in the shorthand form, which omits both the **query keyword** and operation name.
- 'query execution' (executing an operation of any operation type)
  - > GraphQL services which only seek to provide GraphQL **query execution** may choose to only include {ExecutableDefinition} and omit the {TypeSystemDefinition} and {TypeSystemExtension} rules from {Definition}.
  - > Since the result of evaluating a selection set is ordered, the serialized Map of results should preserve this order by writing the map entries in the same order as those fields were requested as defined by **query execution**.

---

The term `query` or `GraphQL query`, when used as a noun outside the above phrases, typically refers to an operation (within a document). In this use it generally applies to all operation types:

- > To achieve congruence with the structure of these applications, a **GraphQL query** itself is structured hierarchically.
- > The **query** is shaped just like the data it returns.
- > A **GraphQL query** can be parameterized with variables, maximizing query reuse, and avoiding costly string building in clients at runtime.
- > Given a **query**, tools can ensure that the query is both syntactically
   correct and valid within the GraphQL type system before execution, i.e. at
   development time, and the server can make certain guarantees about the shape
   and nature of the response.
[...]

PR #777, opened 2020-09-14
Spec edits to reduce "query ambiguity"

benjie:
This PR addresses the "query ambiguity" issue raised in #715 and [discussed in the May 2020 GraphQL Working Group](https://github.com/graphql/graphql-wg/blob/master/notes/2020-05-07.md#query-a-query-query-query-query-ambiguity-10m-benjie).

I went through every usage of the word `query`, `queries` and `queried` in the spec and attempted to discern the different terminology we have going on. I've come up with the following definitions which are broadly in line with the terminology the spec already uses in places.

What follows are notes I took whilst doing this work, which also constitutes the beginnings of a glossary. This glossary is probably sufficient for WG members who already understand these concepts, but I'll need to tidy them up and define them more concretely if we want to add them as a glossary to the spec itself. (Also need to populate it with non-`query` terms.) I think if we do decide to add a glossary we should do that as a separate PR.

## Glossary

When usage of these terms may cause ambiguity\* the normally optional
parenthesized "(GraphQL)" should be included for clarity.

_\* example of an ambiguous phrase: "sending a request to a server" may refer to
an HTTP request, network request, or a GraphQL request._

### (GraphQL) operation

**Definition**: an action you wish GraphQL to perform, for example retrieving
data, mutating data or state, or subscribing to events.

Examples:

> A query **operation** should not mutate data or state, for that a mutation
> **operation** should be used.

Used in the GraphQL spec:

> Let {operation} be the result of {GetOperation(document, operationName)}.

### (GraphQL) operation type

**Definition**: a type of operation supported by GraphQL - "query", "mutation", or
"subscription".

Examples:

> GraphQL currently supports 3 **operation types**: `query`, `mutation` and
> `subscription`.

Not to be confused with (GraphQL) root operation type.

Used in the GraphQL spec:

> Schema extensions are used to represent a schema which has been extended from
> an original schema. For example, this might be used by a GraphQL service which
> adds additional operation types, or additional directives to an existing schema.

### (GraphQL) query operation

A GraphQL operation of type `query`.

Used in the GraphQL spec:

> If {operation} is a query operation: ...

### (GraphQL) mutation operation

A GraphQL operation of type `mutation`.

Used in the GraphQL spec:

> Otherwise if {operation} is a mutation operation: ...

### (GraphQL) subscription operation

A GraphQL operation of type `subscription`.

Used in the GraphQL spec:

> Otherwise if {operation} is a subscription operation: ...

### (GraphQL) root operation type

The GraphQL object type associated with a given GraphQL operation type within
the schema.

Used in the GraphQL spec:

> A schema defines the initial root operation type for each kind of operation it
> supports: query, mutation, and subscription; this determines the place in the
> type system where those operations begin.

### (GraphQL) document

**Definition**: the textual representation (using GraphQL query language) of an
operation (or operations) you wish to perform, including selection sets and
fragments as appropriate.

Examples:

> Once a **GraphQL document** is written, it should always mean the same
> thing and return the same shaped result.

> For example, this **document** will receive the name of the user
> with id 4 from the Facebook implementation of GraphQL.

> Clients use the GraphQL query language to make requests to a GraphQL service.
> We refer to these request sources as **GraphQL documents**. A **document**
> may contain operations (queries, mutations, and subscriptions) as
> well as fragments, a common unit of composition allowing for **data
> requirement reuse**.

[Inspired by](https://github.com/graphql/graphql-js/blob/7a327a6e2045e75c577be924694e6642d3e28e14/src/language/source.js#L15).

Alternatives considered:

- GraphQL operation document (could also be expected to include IDL)
- Request document (too confusing with "request")

### (GraphQL) operation variables

**Definition**: variables declared within an operation.

Previously "query variables".

Example: `$devicePicSize` is an operation variable in the following operation:

```graphql example
query getZuckProfile($devicePicSize: Int) {
  user(id: 4) {
    id
    name
    profilePic(size: $devicePicSize)
  }
}
```

Why not "request variables"? The variables are declared in the operation and
there doesn't seem to be a need to push the definition up to request level.

### (GraphQL) request

**Definition**: the full description of what you wish GraphQL to execute,
including the GraphQL schema, document, variables, operation name and initial
value.

> When using GraphQL over HTTP, it's common to encode the **GraphQL request** as
> JSON.

Used in the GraphQL spec:

> ExecuteRequest(schema, document, operationName, variableValues, initialValue):

### (GraphQL) request error

**Definition**: an error which occurs during validating a GraphQL request
(including coercing the variables) resulting in the entire GraphQL request
being aborted.

Previously: consistently "query error" throughout the spec.

Why not "query error"? It also occurs for mutations, subscriptions, and invalid
documents.

Why not "operation error"? It can occur when operationName is not specified,
and thus the operation is indeterminate. Further it may result from invalid
variables which are a GraphQL request concern, not a GraphQL operation concern.

## Unmodified terms

The verb "to query" is left in may places, as is the term "query language". A few of these are noted below:

> GraphQL is a query language designed to build client applications...

> A GraphQL server's type system must be queryable by the GraphQL language itself...

> It describes the language and its grammar, the type system and the introspection system used to query it...

### query reuse

"query reuse" w.r.t fragments. "query" seems like the wrong term here (because
of how overloaded/vague it is) but I can't come up with a succinct alternative.
Best I've come up with is "data requirement reuse," but I'm not super happy
with it.

###  queried

> By default, the key in the response object will use the field name
> **queried**. However, you can define a different name by specifying an alias.

### querying

> Inline Fragments can be used directly within a selection to condition upon a
> type condition when **querying** against an interface or union.

> Fragments must specify the type they apply to. In this example, `friendFields`
> can be used in the context of **querying** a `User`.

Can be replaced with 'selecting' / 'selecting fields from', using 'select' as
in 'selection set'.

### query of

> A query of an object value must select at least one field.

### request

In many cases 'request' can be replaced with 'fetch' to avoid ambiguity with
'GraphQL request'.

> If providing JSON for the variables' values, we could run this operation and
> **request** profilePic of size `60` width:

### query root object type

Also mutation/subscription root object type and simply "query root type". These
have the same meaning as query "root operation type", but "root object type" is
more explicit that the root operation type is an _object_ type. We should pick
between them and stick to that throughout.

> The type system, as described in the "Type System" section of the spec, must
> provide a query root object type. If mutations or subscriptions are supported,
> it must also provide a mutation or subscription root object type, respectively.

### query execution

Replaced with "request execution" due to the function being called `ExecuteRequest`

### Fetch

> There are three types of operations that GraphQL models:
> 
>   * query - a read-only fetch.
>   * mutation - a write followed by a fetch.
>   * subscription - a long-lived request that fetches data in response to source
>     events.

### Response key

> A field's response key is its alias if an alias is provided, and it is
> otherwise the field's name.

## Ethos

Rather than being conservative with edits and leaving ambiguity I was quite heavy

## Status

This is not ready for merging, it's only a first edit and there's likely inconsistencies where the final rules I applied weren't reapplied to earlier edits. Mostly this is a discovery exercise.

## Next steps

I'll add this to the next GraphQL WG. I have broadly the following questions:

1. in general are we happy with my chosen terms?
1. "root operation type", "root object type" and "root type" term - need to pick one.
1. Is there a better term than "querying" to use when querying a type, e.g. relating to field validity, fragment spreads, etc. Perhaps "selecting" or a derivative, inspired by "selection set"?

### DIFF
--- a/spec/Section 1 -- Overview.md
+++ b/spec/Section 1 -- Overview.md
@@ -26,7 +26,7 @@ Which produces the resulting data (in JSON):
 ```
 
 GraphQL is not a programming language capable of arbitrary computation, but is
-instead a language used to query application services that have
+instead a language used to make requests to application services that have
 capabilities defined in this specification. GraphQL does not mandate a
 particular programming language or storage system for application services that
 implement it. Instead, application services take their capabilities and map them
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -3,7 +3,7 @@
 Clients use the GraphQL query language to make requests to a GraphQL service.
 We refer to these request sources as documents. A document may contain
 operations (queries, mutations, and subscriptions) as well as fragments, a
-common unit of composition allowing for query reuse.
+common unit of composition allowing for data requirement reuse.
 
 A GraphQL document is defined as a syntactic grammar where terminal symbols are
 tokens (indivisible lexical units). These tokens are defined in a lexical
@@ -337,7 +337,7 @@ under-fetching data.
 }
 ```
 
-In this query, the `id`, `firstName`, and `lastName` fields form a selection
+In this query operation, the `id`, `firstName`, and `lastName` fields form a selection
 set. Selection sets may also contain fragment references.
 
 
@@ -408,7 +408,7 @@ Fields are conceptually functions which return values, and occasionally accept
 arguments which alter their behavior. These arguments often map directly to
 function arguments within a GraphQL service's implementation.
 
-In this example, we want to query a specific user (requested via the `id`
+In this example, we want to fetch a specific user (requested via the `id`
 argument) and their profile picture of a specific `size`:
 
 ```graphql example
@@ -438,7 +438,7 @@ Many arguments can exist for a given field:
 Arguments may be provided in any syntactic order and maintain identical
 semantic meaning.
 
-These two queries are semantically identical:
+These two operations are semantically identical:
 
 ```graphql example
 {
@@ -546,7 +546,7 @@ query noFragments {
 ```
 
 The repeated fields could be extracted into a fragment and composed by
-a parent fragment or query.
+a parent fragment or operation.
 
 ```graphql example
 query withFragments {
@@ -567,8 +567,8 @@ fragment friendFields on User {
 }
 ```
 
-Fragments are consumed by using the spread operator (`...`). All fields selected
-by the fragment will be added to the query field selection at the same level
+Fragments are consumed by using the spread operator (`...`). All fields
+selected by the fragment will be added to the field selection at the same level
 as the fragment invocation. This happens through multiple levels of fragment
 spreads.
 
@@ -597,7 +597,7 @@ fragment standardProfilePic on User {
 }
 ```
 
-The queries `noFragments`, `withFragments`, and `withNestedFragments` all
+The operations `noFragments`, `withFragments`, and `withNestedFragments` all
 produce the same response object.
 
 
@@ -616,7 +616,7 @@ Fragments can be specified on object types, interfaces, and unions.
 Selections within fragments only return values when the concrete type of the object
 it is operating on matches the type of the fragment.
 
-For example in this query on the Facebook data model:
+For example in this operation using the Facebook data model:
 
 ```graphql example
 query FragmentTyping {
@@ -1049,7 +1049,7 @@ literal representation of input objects as "object literals."
 Input object fields may be provided in any syntactic order and maintain
 identical semantic meaning.
 
-These two queries are semantically identical:
+These two operations are semantically identical:
 
 ```graphql example
 {
@@ -1113,7 +1113,7 @@ query getZuckProfile($devicePicSize: Int) {
 
 Values for those variables are provided to a GraphQL service along with a
 request so they may be substituted during execution. If providing JSON for the
-variables' values, we could run this query and request profilePic of
+variables' values, we could run this operation and fetch profilePic of
 size `60` width:
 
 ```json example
@@ -1124,7 +1124,7 @@ size `60` width:
 
 **Variable use within Fragments**
 
-Query variables can be used within fragments. Query variables have global scope
+Operation variables can be used within fragments. Operation variables have global scope
 with a given operation, so a variable used within a fragment must be declared
 in any top-level operation that transitively consumes that fragment. If
 a variable is referenced in a fragment and is included by an operation that does
@@ -1146,7 +1146,7 @@ NonNullType :
   - NamedType !
   - ListType !
 
-GraphQL describes the types of data expected by query variables. Input types
+GraphQL describes the types of data expected by operation variables. Input types
 may be lists of another input type, or a non-null variant of any other
 input type.
 
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -337,8 +337,8 @@ under-fetching data.
 }
 ```
 
-In this query operation, the `id`, `firstName`, and `lastName` fields form a selection
-set. Selection sets may also contain fragment references.
+In this query operation, the `id`, `firstName`, and `lastName` fields form a
+selection set. Selection sets may also contain fragment references.
 
 
 ## Fields
@@ -408,7 +408,7 @@ Fields are conceptually functions which return values, and occasionally accept
 arguments which alter their behavior. These arguments often map directly to
 function arguments within a GraphQL service's implementation.
 
-In this example, we want to fetch a specific user (requested via the `id`
+In this example, we want to query a specific user (requested via the `id`
 argument) and their profile picture of a specific `size`:
 
 ```graphql example
@@ -1096,7 +1096,9 @@ If not defined as constant (for example, in {DefaultValue}), a {Variable} can be
 supplied for an input value.
 
 Variables must be defined at the top of an operation and are in scope
-throughout the execution of that operation.
+throughout the execution of that operation. Values for those variables are
+provided to a GraphQL service as part of a request so they may be substituted
+in during execution.
 
 In this example, we want to fetch a profile picture size based on the size
 of a particular device:
@@ -1111,10 +1113,8 @@ query getZuckProfile($devicePicSize: Int) {
 }
 ```
 
-Values for those variables are provided to a GraphQL service along with a
-request so they may be substituted during execution. If providing JSON for the
-variables' values, we could run this operation and fetch profilePic of
-size `60` width:
+If providing JSON for the variables' values, we could request a `profilePic` of
+size `60`:
 
 ```json example
 {
@@ -1124,11 +1124,10 @@ size `60` width:
 
 **Variable use within Fragments**
 
-Operation variables can be used within fragments. Operation variables have global scope
-with a given operation, so a variable used within a fragment must be declared
-in any top-level operation that transitively consumes that fragment. If
-a variable is referenced in a fragment and is included by an operation that does
-not define that variable, the operation cannot be executed.
+Variables can be used within fragments. Variables have global scope with a given operation, so a variable used within a fragment must be declared in any
+top-level operation that transitively consumes that fragment. If a variable is
+referenced in a fragment and is included by an operation that does not define
+that variable, that operation is invalid (see [All Variable Uses Defined](#sec-All-Variable-Uses-Defined)).
 
 
 ## Type References
@@ -1146,9 +1145,9 @@ NonNullType :
   - NamedType !
   - ListType !
 
-GraphQL describes the types of data expected by operation variables. Input types
-may be lists of another input type, or a non-null variant of any other
-input type.
+GraphQL describes the types of data expected by arguments and variables.
+Input types may be lists of another input type, or a non-null variant of any
+other input type.
 
 **Semantics**
 
@@ -1188,8 +1187,8 @@ including or skipping a field. Directives provide this by describing additional 
 Directives have a name along with a list of arguments which may accept values
 of any input type.
 
-Directives can be used to describe additional information for types, fields, fragments
-and operations.
+Directives can be used to describe additional information for types, fields,
+fragments and operations.
 
 As future versions of GraphQL adopt new configurable execution capabilities,
 they may be exposed via directives. GraphQL services and tools may also provide
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1,8 +1,8 @@
 # Type System
 
 The GraphQL Type system describes the capabilities of a GraphQL service and is
-used to determine if a query is valid. The type system also describes the
-input types of query variables to determine if values provided at runtime
+used to determine if a request is valid. The type system also describes the
+input types of operation variables to determine if values provided at runtime
 are valid.
 
 TypeSystemDocument : TypeSystemDefinition+
@@ -79,7 +79,7 @@ schema {
 }
 
 """
-Root type for all your queries
+Root type for all your query operations
 """
 type Query {
   """
@@ -191,7 +191,7 @@ When using the type system definition language, a document must include at most
 one {`schema`} definition.
 
 In this example, a GraphQL schema is defined with both query and mutation
-root types:
+root operation types:
 
 ```graphql example
 schema {
@@ -424,7 +424,7 @@ raised (input values are validated before execution begins).
 
 GraphQL has different constant literals to represent integer and floating-point
 input values, and coercion rules may apply differently depending on which type
-of input value is encountered. GraphQL may be parameterized by query variables,
+of input value is encountered. GraphQL may be parameterized by operation variables,
 the values of which are often serialized when sent over a transport like HTTP. Since
 some common serializations (ex. JSON) do not discriminate between integer
 and floating-point values, they are interpreted as an integer input value if
@@ -601,14 +601,14 @@ FieldsDefinition : { FieldDefinition+ }
 
 FieldDefinition : Description? Name ArgumentsDefinition? : Type Directives[Const]?
 
-GraphQL queries are hierarchical and composed, describing a tree of information.
-While Scalar types describe the leaf values of these hierarchical queries, Objects
+GraphQL operations are hierarchical and composed, describing a tree of information.
+While Scalar types describe the leaf values of these hierarchical operations, Objects
 describe the intermediate levels.
 
 GraphQL Objects represent a list of named fields, each of which yield a value of
 a specific type. Object values should be serialized as ordered maps, where the
-queried field names (or aliases) are the keys and the result of evaluating
-the field is the value, ordered by the order in which they appear in the query.
+requested field names (or aliases) are the keys and the result of evaluating
+the field is the value, ordered by the order in which they appear in the operation.
 
 All fields defined within an Object type must not have a name which begins with
 {"__"} (two underscores), as this is used exclusively by GraphQL's
@@ -687,8 +687,8 @@ type Person {
 }
 ```
 
-Valid queries must supply a nested field set for a field that returns
-an object, so this query is not valid:
+Valid operations must supply a nested field set for a field that returns
+an object, so this operation is not valid:
 
 ```graphql counter-example
 {
@@ -722,7 +722,7 @@ And will yield the subset of each object type queried:
 **Field Ordering**
 
 When querying an Object, the resulting mapping of fields are conceptually
-ordered in the same order in which they were encountered during query execution,
+ordered in the same order in which they were encountered during request execution,
 excluding fragments for which the type does not apply and fields or
 fragments that are skipped via `@skip` or `@include` directives. This ordering
 is correctly produced when using the {CollectFields()} algorithm.
@@ -920,10 +920,10 @@ type Person {
 }
 ```
 
-GraphQL queries can optionally specify arguments to their fields to provide
+GraphQL operations can optionally specify arguments to their fields to provide
 these arguments.
 
-This example query:
+This example operation:
 
 ```graphql example
 {
@@ -948,7 +948,7 @@ Object, Interface, or Union type).
 ### Field Deprecation
 
 Fields in an object may be marked as deprecated as deemed necessary by the
-application. It is still legal to query for these fields (to ensure existing
+application. It is still legal to fetch these fields (to ensure existing
 clients are not broken by the change), but the fields should be appropriately
 treated in documentation and tooling.
 
@@ -1062,7 +1062,7 @@ type Contact {
 }
 ```
 
-This allows us to write a query for a `Contact` that can select the
+This allows us to write a selection set for a `Contact` that can select the
 common fields.
 
 ```graphql example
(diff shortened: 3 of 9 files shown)


eapache:
Thank you for tackling this! 🙏
- The "Questionable Terms" section appears twice in your explanation, I think just copy-pasta?
- For "query reuse", perhaps "selection reuse"?

benjie:
1. Thanks; fixed!
2. maybe :+1: 

Also I accidentally included an earlier decision which was to use "operation document" which I've since rethought and gone with simply "document"; so have edited that into the description too.

sungam3r:
> (GraphQL) request
Definition: the full description of what you wish GraphQL to execute,
including the GraphQL schema, document, variables, operation name and initial
value.

Why do you include schema and initial value in graphql request? This is purely server side implementation concept.

benjie:
@sungam3r good question. The answer’s straight-forward: because that’s how the GraphQL spec defines it; this isn’t definitions for accessing a GraphQL schema over HTTP (for example), a request in schema spec terms relates to the items listed:

https://spec.graphql.org/June2018/#ExecuteRequest()

sungam3r:
GraphQL spec defines implementation part of executing request with some **additional server-side context** - schema and initialValue. In this paragraph, I do not see that the specification describes the content of the request. It says
> To execute a request, the executor must have

Executor can have any number of things to execute the request, not related to the request itself.

> this isn’t definitions for accessing a GraphQL schema over HTTP

It always seemed to me that the request **should** be defined as a (document, operationName, variables) for an arbitrary transport. To be honest, I find it strange if its definition would include something else.

benjie:
I'm not sure; all requests indicate a schema - sometimes by explicitly having a schema instance, sometimes by referencing the schema via URL endpoint. You can't validate a GraphQL request without knowing to which GraphQL schema it applies. (Not true for the initial value, sure.)

But anyway; we're currently discussing something that's _not actually part of the changes proposed by this PR_ - you're currently commenting on my notes on the terms that I added so that we as GraphQL Working Group members would know what's being referenced. An actual RFC to add a glossary would be a separate PR which can and should have this kind of discussion on this and many other topics.

sungam3r:
Ok

spec/Section 2 -- Language.md
```diff
@@ -431,7 +432,7 @@ Many arguments can exist for a given field:
 Arguments may be provided in any syntactic order and maintain identical
 semantic meaning.
 
-These two queries are semantically identical:
+These two operations are semantically identical:
```
vwkd, line 435: “Query operations”, to be consistent.

spec/Section 2 -- Language.md
```diff
@@ -593,7 +594,7 @@ fragment standardProfilePic on User {
 }
 ```
 
-The queries `noFragments`, `withFragments`, and `withNestedFragments` all
+The operations `noFragments`, `withFragments`, and `withNestedFragments` all
```
vwkd, line 597: “Query operations”, to be consistent.

spec/Section 2 -- Language.md
```diff
@@ -251,13 +251,13 @@ and validated to allow client tools to represent many GraphQL uses which may
 appear across many individual files.
 
 If a Document contains only one operation, that operation may be unnamed or
-represented in the shorthand form, which omits both the query keyword and
-operation name. Otherwise, if a GraphQL Document contains multiple
-operations, each operation must be named. When submitting a Document with
-multiple operations to a GraphQL service, the name of the desired operation to
-be executed must also be provided.
+represented in the shorthand form, which omits both the query keyword (for
```
vwkd, line 254: I find this backwards formulation confusing. First the shorthand is defined, and only afterwards it’s mentioned that it only works for queries. Why not say that first? Something like this maybe

“[...] If that single operation is a query, both the query keyword and the name may be omitted.”

But maybe this section could even be deleted, because it’s mentioned below in “Query shorthand” again. But more importantly, here it’s even inaccurate, because it doesn’t mention that the query must also not have variables or directives, otherwise the shorthand can’t be used.

spec/Section 2 -- Language.md
```diff
@@ -294,9 +294,10 @@ mutation {
 
 **Query shorthand**
 
-If a document contains only one query operation, and that query defines no
-variables and contains no directives, that operation may be represented in a
-short-hand form which omits the query keyword and query name.
+If a document contains only one operation and the operation is query operation,
```
vwkd, line 297: “a query operation”

spec/Section 2 -- Language.md
```diff
@@ -1045,7 +1046,7 @@ literal representation of input objects as "object literals."
 Input object fields may be provided in any syntactic order and maintain
 identical semantic meaning.
 
-These two queries are semantically identical:
+These two operations are semantically identical:
```
vwkd, line 1049: “Query operations”, to be consistent.

spec/Section 3 -- Type System.md
```diff
@@ -660,8 +661,8 @@ type Person {
 }
 ```
 
-Valid queries must supply a nested field set for a field that returns
-an object, so this query is not valid:
+Valid operations must supply a nested field set for a field that returns
+an object, so this operation is not valid:
```
vwkd, line 665: “Query operation”, to be consistent.

spec/Section 3 -- Type System.md
```diff
@@ -893,10 +894,10 @@ type Person {
 }
 ```
 
-GraphQL queries can optionally specify arguments to their fields to provide
+GraphQL operations can optionally specify arguments to their fields to provide
 these arguments.
 
-This example query:
+This example operation:
```
vwkd, line 900: “Query operation”, to be consistent.

spec/Section 4 -- Introspection.md
```diff
@@ -16,7 +16,7 @@ type User {
 }
 ```
 
-The query
+The operation
```
vwkd, line 19: “Query operation”, to be consistent.

spec/Section 5 -- Validation.md
```diff
@@ -1587,7 +1587,7 @@ extend type Query {
 }
 ```
 
-The following queries are valid:
+The following operations are valid:
```
vwkd, line 1590: “Query operations”, to be consistent.

spec/Section 5 -- Validation.md
```diff
@@ -1607,7 +1607,7 @@ query TakesListOfBooleanBang($booleans: [Boolean!]) {
 }
 ```
 
-The following queries are invalid:
+The following operations are invalid:
```
vwkd, line 1610: “Query operations”, to be consistent.

spec/Section 6 -- Execution.md
```diff
@@ -687,7 +687,7 @@ When more than one field of the same name is executed in parallel, their
 selection sets are merged together when completing the value in order to
 continue execution of the sub-selection sets.
 
-An example query illustrating parallel fields with the same name with
+An example operation illustrating parallel fields with the same name with
```
vwkd, line 690: “Query operation”, to be consistent.

spec/Section 7 -- Response.md
```diff
@@ -89,10 +89,10 @@ response and ending with the field associated with the error. Path segments
 that represent fields should be strings, and path segments that
 represent list indices should be 0-indexed integers. If the error happens
 in an aliased field, the path to the error should use the aliased name, since
-it represents a path in the response, not in the query.
+it represents a path in the response, not in the request.
 
 For example, if fetching one of the friends' names fails in the following
-query:
+operation:
```
vwkd, line 95: “Query operation”, to be consistent.

vwkd (commented):
I like the clarifications very much! 👍

I saw that you introduced the term “query operation” for an operation of type “query”, but didn’t use it consistently in the examples. To be consistent, the examples should then also be “query operations” to distinguish them from mutation / subscription operations. Of course the alternative would be to change “query operation” to “operation” in the other places instead.

As another thought, maybe for the glossary, now that “query” has been de-overloaded. What does the naked term “query” actually mean and when is it appropriate to use it? Should it just be used for “query language” and “querying”?

spec/Section 2 -- Language.md
benjie: I think this should be proposed via a separate PR; let's keep this one focussed on just the task at hand :+1:

benjie:
@vwkd Thanks for your review. There's a reason that I used `operation` over `query operation` in those cases: though they _are_ query operations, the text applies for all types of operations, so the fact it's a `query` operation is incidental - part of the intention was to broaden the meaning that the text implies to make it clear that the text does not only apply to queries even though that's generally what we use in the examples.

vwkd:
@benjie Thanks for the reply. I intendedly left out the places where an operation was defined or mentioned in general, and commented only where there was a concrete example showing a query operation. I don't think explicitly mentioning the type of the operation in the example is implying that whatever was defined before it couldn't be used for operations of other types. On the contrary, I think being explicit about the type of operation used in the example helps clear confusion.

vwkd:
Is it only me, or does `SchemaDefinition` not have a `Description` anymore? I can't seem to find references in the code that indicate any changes regarding this, but on my fresh local build the `Description` in the `SchemaDefinition` is gone. Weird, since I can repeat this with a fresh clone and running `npm i && npm run build`.

mcohen75:
Would it make sense to formalize the glossary you've created in the description, even in just for "internal" use? It would seem useful for those creating documentation as part of formal working group activities. I think members of the community would also benefit from something like this as it would help to further align usage of these terms.

benjie:
Absolutely; it’s one of my action items to push that forward (it’s just currently queued up behind a few higher priority items).

spec/Section 3 -- Type System.md
```diff
@@ -68,7 +68,7 @@ schema {
 }
 
 """
-Root type for all your queries
+Root type for all your query operations
```
mcohen75, line 71: Is this a valid use of the term `query operation`? According to the spec and the glossary I believe a query operation is an `OperationDefinition` with the type `query`.

If this is not a valid use of the term, what term should be used to describe these fields on the Query type?

mcohen75: I recently had a conversation with a colleague here at Indeed and use (misused?) the term `query operation` in a similar context.

I said something like "when a response includes data for multiple query operations...". He [pointed me to the spec](http://spec.graphql.org/draft/#GetOperation()) and pointed out that only one `query operation` can be executed.

What I was trying to communicate was that it's possible to select multiple fields on the top level Query type.

benjie: `Query` is the root type for all query operations; you issue a selection set against the Query root type, this selection set may select more than one field like most selection sets (the exception being the selection set at the root of the `subscription` operation).

Here's a (single) query operation issuing a selection set against a root Query type:

```graphql
query {
  __typename
  myField
  myOtherField { id }
}
```

So yes, I believe the usage here is correct.

>  According to the spec and the glossary I believe a query operation is an OperationDefinition with the type query.

Correct.

> He pointed me to the spec and pointed out that only one query operation can be executed.

He's correct :+1: 

> What term should be used to describe these fields on the Query type?

You could call them "Query fields" I guess; just like fields on the User type would by "User fields." I'm not sure we have an established name for them, the root level fields on the `mutation` operation would generally be called "mutation fields" and for `subscription` "subscription fields" so I think "query fields" would make sense, but is potentially ambiguous.

benjie:
I'll review this in the next few hours :+1:

spec/Section 4 -- Introspection.md
```diff
@@ -16,7 +16,7 @@ type User {
 }
 ```
 
-The operation
+The request:
```
benjie, line 19: I feel that calling this a request explicitly may be confusing - technically it's sufficient, but it doesn't provide space for providing variables/etc which are commonly part of a request, making it harder for the reader to reason about what a request really is. Perhaps change it to something like: `A request containing the following operation:`?

spec/Section 4 -- Introspection.md
```diff
@@ -319,7 +319,7 @@ Fields
 
 #### Input Object
 
-Input objects are composite types used as inputs into operations defined as a
+Input objects are composite types used as inputs to operations defined as a
 list of named input values.
```
benjie: Perhaps:

```
Input objects are composite types used as inputs to some arguments and variables, they are defined as a list of named input values.
```

Riffing off of the previous text `GraphQL describes the types of data expected by arguments and variables.`

benjie (commented):
Generally good improvements to the text; I've called out a few things that I don't fully agree with and some minor issues which I will fix shortly.

spec/Section 1 -- Overview.md
```diff
@@ -38,30 +38,30 @@ GraphQL has a number of design principles:
...
    the client that is responsible for specifying exactly how it will consume
-   those published capabilities. These operations are specified at field-level
-   granularity. In the majority of client-server applications written
-   without GraphQL, the service determines the data returned in its various
-   scripted endpoints. A GraphQL operation, on the other hand, returns exactly
-   what a client asks for and no more.
+   those published capabilities. These requests are specified at field-level
+   granularity. In the majority of client-server applications written without
+   GraphQL, the service determines the data returned from from its various
+   endpoints. A GraphQL response, on the other hand, returns exactly what a
```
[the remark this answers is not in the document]
benjie: This isn't strictly true either. E.g. you can add additional information in the `extensions` property of the response body.

spec/Section 3 -- Type System.md
```diff
@@ -133,37 +135,40 @@ A schema defines the initial root operation type for each kind of operation it
...
+For example, this example operation:
 
 ```graphql example
 query {
   myName
 }
 ```
 
-Is valid when the `query` root operation type has a field named "myName".
+Is valid when the {`query`} root operation type has a field named "myName".
```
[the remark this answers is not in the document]
benjie: I'm going to add `only` as discussed below.

spec/Section 1 -- Overview.md
```diff
@@ -38,30 +38,30 @@ GraphQL has a number of design principles:
...
    publishes the capabilities that its clients are allowed to consume. It is
    the client that is responsible for specifying exactly how it will consume
-   those published capabilities. These operations are specified at field-level
-   granularity. In the majority of client-server applications written
-   without GraphQL, the service determines the data returned in its various
-   scripted endpoints. A GraphQL operation, on the other hand, returns exactly
-   what a client asks for and no more.
+   those published capabilities. These requests are specified at field-level
+   granularity. In the majority of client-server applications written without
+   GraphQL, the service determines the data returned from from its various
```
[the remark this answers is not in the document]
benjie: Fixed in 544cbaff75c8babcecb8b3f63eb489007c5405e6

spec/Section 1 -- Overview.md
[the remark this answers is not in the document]
benjie: Fixed in 544cbaff75c8babcecb8b3f63eb489007c5405e6

spec/Section 2 -- Language.md
```diff
@@ -251,16 +251,17 @@ However documents which do not contain {OperationDefinition} or do contain
...
+operations, each operation must be named. When submitting a Document with
+multiple operations to a GraphQL service, the name of the desired operation to
+be executed must also be provided.
 
-GraphQL services which only seek to provide GraphQL request execution may choose
-to only include {ExecutableDefinition} and omit the {TypeSystemDefinition} and
-{TypeSystemExtension} rules from {Definition}.
+GraphQL services which only seek to execute GraphQL requests and not construct
+new GraphQL schema may choose to only permit {ExecutableDefinition} and omit the
+{TypeSystemDefinition} and {TypeSystemExtension} rules from {Definition}.
```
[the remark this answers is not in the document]
benjie: Fixed in 199d2f18cf5dcbf797162873e259443bb6e45d32

spec/Section 3 -- Type System.md
```diff
@@ -15,12 +15,14 @@ The GraphQL language includes an
 describe a GraphQL service's type system. Tools may use this definition language
 to provide utilities such as client code generation or service boot-strapping.
 
-GraphQL tools which only seek to provide GraphQL request execution may choose not
-to parse {TypeSystemDefinition}.
+GraphQL tools or services which only seek to execute GraphQL requests and not
+construct new GraphQL schema may choose not to parse {TypeSystemDefinition} or
+{TypeSystemExtension}.
```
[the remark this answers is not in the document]
benjie: Fixed in 199d2f18cf5dcbf797162873e259443bb6e45d32

spec/Section 3 -- Type System.md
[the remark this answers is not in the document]
benjie: Fixed in b49462d67e7e1e2be420a620edcf560086c72c84

spec/Section 3 -- Type System.md
```diff
@@ -133,37 +135,40 @@ A schema defines the initial root operation type for each kind of operation it
...
 type Query {
   myName: String
 }
 ```
 
-Similarly, the following mutation is valid if a `mutation` root operation type
-has a field named "setName". Note that the `query` and `mutation` root types
-must be different types.
+Similarly, the following mutation is valid if a {`mutation`} root operation type
+has a field named "setName".
```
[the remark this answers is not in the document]
benjie: Fixed in b49462d67e7e1e2be420a620edcf560086c72c84

spec/Section 4 -- Introspection.md
benjie: Fixed in 324a6bfd898b1a6572bc810c3ce0a5c65e5b721e

benjie:
I've added 4 more tiny commits which are justified in the resolved comments above (including references to the commit). There's a couple other changes that I think may warrant exploration (these are still open above); but I'm happy with this being merged as is.

spec/Section 1 -- Overview.md
```diff
@@ -38,30 +38,30 @@ GraphQL has a number of design principles:
 
  * **Hierarchical**: Most product development today involves the creation and
    manipulation of view hierarchies. To achieve congruence with the structure
-   of these applications, a GraphQL operation itself is structured hierarchically.
-   The operation is shaped just like the data it returns. It is a natural
-   way for clients to describe data requirements.
+   of these applications, a GraphQL request itself is structured
+   hierarchically. The request is shaped just like the data it returns. It is
+   a natural way for clients to describe data requirements.
```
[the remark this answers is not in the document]
leebyron: > I imagine you're trying to avoid introducing the concept of operation too early

That's exactly right - and here we're looking for a common term as we introduce GraphQL's principles as we compare it to other API systems

spec/Section 3 -- Type System.md
leebyron: FWIW the historical official name for these were "root fields" - when we generalized roots to be no different from other types, we dropped an official name. I've also heard "top level fields" which makes sense to me, but "Query fields" or "Fields on the Query type" both would be perfectly clear

spec/Section 1 -- Overview.md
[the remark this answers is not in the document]
benjie: Sure; resolving 👍

spec/Section 4 -- Introspection.md
benjie: This has been addressed.

benjie:
The glossary entries in this PR description have been extracted to https://github.com/graphql/graphql-spec/pull/846 (and they've also been improved).

PR #784, opened 2020-09-30
Update criterion match in input union RFC

dfreeman:
Currently for **Criterion F** (`Migrating a field to a polymorphic input type is non-breaking`), the tagged unions option is listed as meeting that criterion on the grounds "No migration required, as this pattern is already possible".

This seems to actually be an answer to **Criterion I** (`Input unions should be easy to upgrade from existing solutions`), which lists the same explanation.

As I understand criterion F (and what the responses for the other proposals seem to be addressing) isn't about migrating from other input union approximations today. Rather, in a world where _some_ input union proposal is a part of the spec, it examines migrating an input that currently only accepts a single type to one that accepts a union.

For example, given these types:

```gql
input CatInput {
  name: String!
  age: Int!
  livesLeft: Int
}

input DogInput {
  name: String!
  age: Int!
  breed: DogBreed
}

input AnimalInput @oneOf {
  cat: CatInput
  dog: DogInput
}
```

Consider migrating to an input union for the `animals` argument:

```gql
type Mutation {
  # From this:
  logAnimalDropOff(location: String, animals: [CatInput!]!): Int

  # To this:
  logAnimalDropOff(location: String, animals: [AnimalInput!]!): Int
}
```

This kind of situation is what the other proposals appear to be addressing in their explanations of how they do (or don't) meet Criterion F. With tagged unions, a valid `CatInput` is not a valid `AnimalInput`, so changing `pet` from a simple input type to a polymorphic one is a breaking change.

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -902,7 +902,7 @@ A quick glance at the evaluation results. Remember that passing or failing a spe
 | [C][criteria-c] 🥇 | ✅ | ✅⚠️ | 🚫 | ⚠️ | ✅ |
 | [D][criteria-d] 🥇 | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [E][criteria-e] 🥉 | 🚫 | 🚫 | ✅⚠️ | 🚫 | ✅ |
-| [F][criteria-f] 🥉 | ✅⚠️ | ✅⚠️ | ✅ | ⚠️ | ✅ |
+| [F][criteria-f] 🥉 | ✅⚠️ | ✅⚠️ | ✅ | ⚠️ | 🚫 |
 | [G][criteria-g] 🥉 | ❔ | ❔ | ❔ | ❔ | ❔ |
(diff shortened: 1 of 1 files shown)


benjie (changes requested):
:+1: 

Please also update line 905 (in the "Evaluation Overview" table)

dfreeman:
@benjie Thanks for catching that! Updated.

benjie (approved):

binaryseed:
I think this isn't so clearly a failing grade, I'd suggest it's both passing and failing. But I'm going to merge this anyway :)

dfreeman:
> I'd suggest it's both passing and failing. But I'm going to merge this anyway :)

@binaryseed Can you unpack that a little? It seems like a clear 🚫 to me on this specific criterion, but more to the point, if there's not consensus on whether this was a correct change, I'd honestly rather have seen it not get merged.

The point of laying out the criteria in this RFC document is to help figure out objective metrics to weigh the various options against, right? This kind of ambiguity feels like it undermines that effort, and given how complex the design constraints are here, I'd hate to be contributing to that 🙂

PR #793, opened 2020-11-13
[RFC] Default value coercion rules

benjie:
[Coercing Field Arguments](http://spec.graphql.org/draft/#sec-Coercing-Field-Arguments) states:

> 5.c. Let defaultValue be the default value for argumentDefinition.
> [...]
> 5.h. If hasValue is not true and defaultValue exists (including null):
> 5.h.i. Add an entry to coercedValues named argumentName with the value defaultValue.
> [...]
> 5.j.iii.1. If value cannot be coerced according to the input coercion rules of argumentType, throw a field error.
> 5.j.iii.2. Let coercedValue be the result of coercing value according to the input coercion rules of argumentType.
> 5.j.iii.3. Add an entry to coercedValues named argumentName with the value coercedValue.

Here we note that there is no run-time coercion of `defaultValue`, which makes sense (why do at runtime that which can be done at build time?). However, there doesn't seem to be a rule that specifies that `defaultValue` must be coerced at all, which leads to consequences:

1. you could use any value for defaultValue and it could break the type safety guarantees of GraphQL
2. [nested defaultValues are not applied](https://github.com/graphql/graphql-js/issues/385)

When building the following GraphQL schema programmatically (code below) with GraphQL.js:

```graphql
type Query {
  example(inputObject: ExampleInputObject! = {}): Int
}

input ExampleInputObject {
  number: Int! = 3
}
```

And a resolver for `Query.example`:

```js
resolve(source, args) {
  return args.inputObject.number;
}
```

You might expect the following queries to all gave the same result:

```graphql
query A {
  example
}
query B {
  example(inputObject: {})
}
query C {
  example(inputObject: { number: 3 })
}
query D($inputObject: ExampleInputObject! = {}) {
  example(inputObject: $inputObject)
}
```

However, it turns out that query A's result differs:

```json
{"example":null}
{"example":3}
{"example":3}
{"example":3}
```

This is because `defaultValue` for `Query.example(inputObject:)` was not coerced, so none of the defaultValues of `ExampleInputObject` were applied.

This is extremely unexpected, because looking at the GraphQL schema definition it looks like there's no circumstance under which `ExampleInputObject` may not have `number` as an integer; however when the `defaultValue` of `Query.example(inputObject:)` is used, the value of `number` is `undefined`.

<details>
<summary>Example runnable with GraphQL.js</summary>

```js
const {
  graphqlSync,
  printSchema,
  GraphQLSchema,
  GraphQLObjectType,
  GraphQLInputObjectType,
  GraphQLInt,
  GraphQLNonNull,
} = require("graphql");

const ExampleInputObject = new GraphQLInputObjectType({
  name: "ExampleInputObject",
  fields: {
    number: {
      defaultValue: 3,
    },
  },
});

  name: "Query",
  fields: {
    example: {
      args: {
        inputObject: {
          type: new GraphQLNonNull(ExampleInputObject),
          defaultValue: {},
        },
      },
      type: GraphQLInt,
      resolve(source, args) {
        return args.inputObject.number;
      },
    },
  },
});

  query: Query,
});

// All four of these should be equivalent?
const source = /* GraphQL */ `
  query A {
    example
  }
  query B {
    example(inputObject: {})
  }
  query C {
    example(inputObject: { number: 3 })
  }
  query D($inputObject: ExampleInputObject! = {}) {
    example(inputObject: $inputObject)
  }
`;
const result1 = graphqlSync({ schema, source, operationName: "A" });
const result2 = graphqlSync({ schema, source, operationName: "B" });
const result3 = graphqlSync({ schema, source, operationName: "C" });
const result4 = graphqlSync({ schema, source, operationName: "D" });

console.log(JSON.stringify(result1.data));
console.log(JSON.stringify(result2.data));
console.log(JSON.stringify(result3.data));
console.log(JSON.stringify(result4.data));
```

</details>

This was [raised against GraphQL.js back in 2016](https://github.com/graphql/graphql-js/issues/385), but @IvanGoncharov closed it early last year stating that GraphQL.js conforms to the GraphQL Spec in this regard.

My proposal is that when a `defaultValue` is specified, the GraphQL implementation should coerce it to conform to the relevant type just like it does for runtime values as specified in [Coercing Variable Values](https://spec.graphql.org/draft/#sec-Coercing-Variable-Values) and [Coercing Field Arguments](https://spec.graphql.org/draft/#sec-Coercing-Field-Arguments).

---

This is validated for query documents (and schema defined as SDL), because:

> Literal values must be compatible with the type expected in the position they are found as per the coercion rules defined in the Type System chapter.
> -- http://spec.graphql.org/draft/#sel-FALXDFDDAACFAhuF

But there doesn't seem to be any such assertion for GraphQL schemas defined in code.

---

GraphQL.js implementation: https://github.com/graphql/graphql-js/pull/3814


benjie:
An alternative approach is to apply the coercion during execution in the same way it is done for user-supplied values (the spec mods for this would look like: https://github.com/graphql/graphql-spec/commit/cb29f6e995b1d43235ba9f69ee4ad0ad3d072e46) but this could of course be done once and cached at schema build time.

benjie:
NOTE: we'd also have to prevent infinite recursion in default values; for example the following should throw a schema validation error:

```graphql
input O {
  o: O = {}
  a: Int
}
```

but the following should be fine:

```graphql
input O {
  o: O = { o: null }
  a: Int
}
```

eapache:
Whether a server does the coercion at execution time or build time is largely an implementation detail, but I do think we should specify the more natural behaviour that comes from doing it at execution time: expected output from `print(schema)`, and the "free" infinite recursion guard that presumably already protects coercion in regular runtime inputs.

graphql-ruby at least already seems to try to coerce the default value, but if the coercion fails it just silently passes a null to the resolver even if the argument is required. It would be great to have this behaviour well-defined.

benjie:
> I do think we should specify the more natural behaviour that comes from doing it at execution time: expected output from print(schema), and the "free" infinite recursion guard that presumably already protects coercion in regular runtime inputs.

This is exactly what I'm thinking too; it's why I closed my graphql-js PR because I think the solution's going to be somewhat more involved.

spec/Section 6 -- Execution.md
```diff
@@ -586,8 +588,10 @@ CoerceArgumentValues(objectType, field, variableValues):
         name {variableName}.
     * Otherwise, let {value} be {argumentValue}.
     * If {hasValue} is not {true} and {defaultValue} exists (including {null}):
+      * Let {coercedDefaultValue} be the result of coercing {defaultValue} according to the
+        input coercion rules of {argumentType}.
```
benjie, line 592: NOTE: this is memoizeable and could be computed at schema build time.

benjie:
Put a little more time into this spec, I think it's much more viable now.

IvanGoncharov:
@benjie I'm all for adding validation on default values but I'm against applying coercion rules on them, because:
1. Input coercion rules are done to simplify passing input values for clients but the same simplicity makes it less obvious for readers of SDL (they need to figure out coercion rules). 
2. Coercion rules for scalars are server-specific (even for standard scalars) so we exposing this behavior through introspection. For example for `graphql-js` default values `"5"` and `5` is equal for `ID` scalar but you can't reliably know that from introspection and it can affect tooling (e.g. tool that generates changelog).
3. Introduces recursion during coercion and non-trivial validation rule.
4. So far we designed SDL in such a way that every type is self-descriptive (e.g. need to specify all fields from interfaces it implements).
5. It's a breaking change for current SDL files and more importantly it would be triggered by a simple update of GraphQL lib so it would be the first SDL breaking change we added that is not opt-in.

spec/Section 3 -- Type System.md
```diff
@@ -818,6 +818,9 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          characters {"__"} (two underscores).
       2. The argument must accept a type where {IsInputType(argumentType)}
          returns {true}.
+      3. If the argument has a default value, {defaultValue} must be compatible
+         with {argumentType} as per the coercion rules for that type, and
+         coercion of {defaultValue} must not cause an infinite loop.
```
andimarek, line 823: I am not sure we should state `not cause an infinite loop`: this is a general requirement,right? :-)

benjie: Letting people know of the risk is important ;) But seriously, this is the part I need to pick Lee's brains on, though if the RFC @IvanGoncharov referenced is merged then infinite recursion in defaultValues will be prevented elsewhere, and thus this won't be needed.

benjie:
### The infinite recursion during coercion problem

The following schema is (I believe) valid GraphQL _currently_:

```graphql
# SCHEMA 1

input A {
  b: B = {}
}

input B {
 a: A = {}
}

type Query {
  q(a: A = {}): Int
}
```

_(Note: we could also do this with a self-reference `input A { a: A = {} }` but I felt spreading it across two types make it clearer there's some walking to do.)_

In current implementations this will be fine because the default value is not coerced and thus there is no recursion. 

However, if we implement this RFC and ignore the "must not cause an infinite loop" part (highlighted by @andimarek) then this could cause infinite recursion during coercion.

However; the following GraphQL schema would be fine as there'd be no infinite recursion:


```graphql
# SCHEMA 2

input A {
  b: B = {}
}

input B {
 a: A = {b: null}
}

type Query {
  q(a: A = {}): Int
}
```

Once we reach a situation where `B.a`'s default value will be used the chain will end because the default for `B.a` explicitly provides that the next `A.b` will be `null`.

I think we need a schema validation rule that invalidates "SCHEMA 1". This would be a breaking change to the GraphQL spec, but I think it's an acceptable one and can be easily resolved by GraphQL schema designers by explicitly changing the defaults from `{}` to `{a: null}` / `{b: null}` as appropriate - a very minor fix that will not break client queries.

I do not know how to write this change into the spec (hence the "must not cause an infinite loop" placeholder); @leebyron could you give some hints on the approach you'd take?

benjie:
@IvanGoncharov mentioned that https://github.com/graphql/graphql-spec/pull/701 might help with the recursive defaults problem, but alas I don't think it will as currently stated - the RFC seems to relate to making sure fields are nullable to prevent expressing a state that's not representable via a JSON input, this doesn't solve the infinite recursion issue with default values above (note the fields above are all nullable).

benjie:
@IvanGoncharov in response to your feedback, we discussed this during the WG but I'll include a summary of my positions here:

> Input coercion rules are done to simplify passing input values for clients but the same simplicity makes it less obvious for readers of SDL (they need to figure out coercion rules).

For me, seeing `= {}` means that I would read it as "this would be the same as if I were to pass `{}` explicitly" which is _currently not the case_ so I'd actually argue the opposite of this: people reading the SDL have to know that the defaults will _not_ be applied, which is entirely unexpected.

Ultimately, it is up to the people who develop the schema to decide if this is an issue for them or not - if it is they can consider not using defaultValue at all.

> Coercion rules for scalars are server-specific (even for standard scalars) so we exposing this behavior through introspection. For example for graphql-js default values "5" and 5 is equal for ID scalar but you can't reliably know that from introspection and it can affect tooling (e.g. tool that generates changelog).

Can you expand on this, and perhaps suggest how the proposal can address it/minimise it? To my mind the `defaultValue` should be exactly an equivalent of the user passing a value explicitly, so anything a user can pass explicitly should be representable as defaultValue and vice versa.

> Introduces recursion during coercion and non-trivial validation rule.

Indeed, I think this is an acceptable trade-off.

> So far we designed SDL in such a way that every type is self-descriptive (e.g. need to specify all fields from interfaces it implements).

This was discussed in the WG, and I feel that default value / passed value equivalence (i.e. seeing `= {}` should be the same as passing `{}` explicitly) outweighs this concern by a significant degree.

> It's a breaking change for current SDL files and more importantly it would be triggered by a simple update of GraphQL lib so it would be the first SDL breaking change we added that is not opt-in.

Agree, however the fix is straightforward and it would be determined at schema validation time (not runtime) so should be quick to address for implementors. Further, if implementors are already doing what you say they should be doing (i.e. supplying the full expected default value, applying whatever defaults are needed manually) then their schemas will not be broken by this change.

spec/Section 3 -- Type System.md
andimarek: I think we must clearly distinguish several aspects here: I agree that `defaultValue` must be coerced.  
And I also agree that in the case of input objects there is the possibility of default values causing infinite recursion if evaluated statically. But this is a special case for input objects coercion and we should deal with it separately from the default value coercion (this doesn't mean separate PR, but just two problems).

benjie: Agreed; I think this needs to be a separate validation rule. This text is just a placeholder to remind us something needs to be written about this problem, I never intended this part to make it into the spec hence needing to pick Lee’s brains on the right approach. https://github.com/graphql/graphql-spec/pull/793#issuecomment-738736539

Alan-Cha:
During the January working group meeting, we discussed the idea of sharing example schemas so we have a better idea of how to move forward and I mentioned that I worked on a [research paper](https://arxiv.org/abs/1907.13012) where we analyzed a large corpus (8,399) of open source schemas that we collected from GitHub. 

I mentioned that there may be complications with sharing this corpus but it turns out I was misremembering some of the details. We were not able to directly publish the raw schemas and "recovered" schemas (schemas formed by merging small sub-schemas), but we were able to publish links to the commit-stamped files!

See [here](https://github.com/ErikWittern/graphql-schemas).

Additionally, we also published the scripts used to collect, clean, and refactor all the schemas in the corpus. This could be used to generate a new _up to date_ corpus from GitHub.

mmatsa:
@benjie asked for real examples of use.  Here’s an “anonymized” example from a real schema:

```graphql
input Z {
  x: Int = 1
  y: Int = 2
}

input A {
  b: Z = {x: 1, y: 2}
  c: Z = {x: 1, y: 3}
}

type T {
    f(a: A = {b: {x: 1, y: 2}, c: {x: 1, y: 3}}): F
```

Note that there is really use of multiple levels of default values here.  Given the spec today the `f()` definition is required to override them even though the values are identical.  This schema uses default input objects for arguments hundreds of times.  Obviously, it must overwrite the values, so I think that either of the changes proposed to the spec recently wouldn't break it. I'm sharing it as only one example, per @benjie 's request.

mcohen75:
Providing another example here from a real-world schema from Indeed's GraphQL API.

We have a number of query operations that define nested input types that specify default values. 

The most prominent example is the core job search operation:
```
jobSearch(
  cursor: String
  filters: [JobSearchFilterInput!]! = []
  grouping: JobSearchGroupingInput! = {type: EMPLOYER_TITLE}
  includeSponsoredResults: JobSearchIncludeSponsoredResults! = NONE
  lastSearchTime: Timestamp
  limit: Int! = 10
  location: JobSearchLocationInput! = {radius: 0, radiusUnit: KILOMETERS, where: ""}
  numNextPages: Int! = 1
  numPrevPages: Int! = 1
  sort: JobSearchSortOrder! = RELEVANCE
  what: String! = ""
): JobSearchPayload
```

Note that the default values are specified inline for nested inputs.

dwwoelfel:
GitHub uses objects as defaultValues a lot in their schema for their orderBy arguments. 

For example, `Query.securityAdvisories` looks something like:

```graphql
securityAdvisories(
   ...
  orderBy: GitHubSecurityVulnerabilityOrder = {field: UPDATED_AT, direction: DESC}
   ...
): GitHubSecurityVulnerabilityConnection!
```

benjie:
A developer from a large company (household name) DM'd me on Twitter to share parts of their schema, and they have a similar pattern to GitHub for ordering; the relevant parts of their schema looks a bit like (manually anonymised):

```graphql
enum SortDirection {
  ASC
  DESC
}

input FooOrder {
  sort: FooSort!
  direction: SortDirection! = ASC
}

type Query {
  foos(orderBy: FooOrder = {sort: DATE, direction: ASC}): FooConnection
}
```

They also use this for conditions:

```graphql
enum OperatorEnum {
  OR
  AND
}

input BarSearchInput {
  queries: [BarQuery]
  queryOperator: OperatorEnum = OR
}

type Query {
  barSearch(search: BarSearchInput = {queryOperator: OR}): BarConnection
}
```

And have defaults used for input objects used in mutations:

```graphql
enum BazType {
  CAT
  HEDGEHOG
  ZEBRA
}

input BazInput {
  type: BazType = CAT
  url: String!
  name: String
}

type Mutation {
  addBaz(quxId: ID!, baz: BazInput!): Baz
}
```

spec/Section 3 -- Type System.md
benjie: An argument can never cause an infinite loop anyway; so I've removed this - but I've also fixed the "must not cause an infinite loop" text that was below too.

benjie:
"The infinite recursion during coercion problem" - solved - new wording is included in this PR.

liliserf:
Here's an example from pagination in our real-world schema:
```
input PaginationConfig {
  cursor: String
  direction: PaginationDirection = AFTER
  limit: Int = 20
}
```

benjie:
@liliserf Thanks for sharing! Do any of the arguments (or other input object fields) that reference this type have a default?

backbone87:
regarding readability:
```gql
enum SortDirection {
  ASC
  DESC
}
enum FooSort {
  DATE
}

input FooOrder {
  sort: FooSort!
  direction: SortDirection! = ASC
}

type Query {
  foos(orderBy: FooOrder = { sort: DATE, direction: ASC }): FooConnection
  otherFoos(orderBy: FooOrder = { sort: DATE }): FooConnection
}
```

here `foos` makes it more clear what the default direction is, while for `otherFoos` the reader would have to look up the default sort direction in another place (`input FooOrder` could be "far away" in the schema).

but the resolver should get in both cases `{ sort: "DATE", direction: "ASC" }`.

liliserf:
> @liliserf Thanks for sharing! Do any of the arguments (or other input object fields) that reference this type have a default?

Nope, just the one.

spec/Section 6 -- Execution.md
benjie: TODO: add this as a non-normative note below.

leebyron:
The next step for this RFC is definitely a reference implementation change. It's hard to assert correct behavior and clarity in the corner cases without writing actual test cases. That will build confidence that this change is the right one.

spec/Section 3 -- Type System.md
```diff
@@ -1580,6 +1583,17 @@ Literal Value            | Variables               | Coerced Value
       characters {"__"} (two underscores).
    3. The input field must accept a type where {IsInputType(inputFieldType)}
       returns {true}.
+   4. If the input field has a non-null default value:
```
leebyron, line 1586: Should this just be:

```suggestion
   4. If the input field has a default value:
```

Since a `null` default value might fail the inner step 2, "must be compatible with {inputFieldType}" if that type might be a Non-Null

spec/Section 3 -- Type System.md
```diff
@@ -1580,6 +1583,17 @@ Literal Value            | Variables               | Coerced Value
...
+      1. If the input field references this Input Object either directly or
+         through referenced Input Objects, all input fields in the chain of
+         references which reference this Input Object must either:
+         1. have no default value; or
+         2. have a {null} default value; or
+         3. have a default value, {nestedDefaultValue}, such that the value for
+            this field within {nestedDefaultValue} is either {null} or an empty
+            list.
```
leebyron, line 1594: Could this possibly be incorporated into the existing step 3 below? 

Perhaps:

> at least one of the fields in the chain of references must be either a nullable or a List type.

adds something along the lines of "and have either no default value, or a default value of either {null} or an empty list"

spec/Section 3 -- Type System.md
benjie: The `non-null` is here because if the default for the field is `null` (assuming this is compatible with `inputFieldType`), then no further validation is required since no recursion through the default value can occur. Consider A vs B in the below scenario; both are fine, but the validation of A can stop at the `A.a` field immediately, whereas the validation of `B` has to traverse through C and D to get back to finding that the nested reference to `B` via `D.b` defaults to `null`.

```graphql
input A {
  a: A = null
}

input B {
  c: C = {}
}

input C {
  d: D = {}
}

input D {
  b: B = null
}
```

spec/Section 3 -- Type System.md
benjie: I don't think so, these constraints live at different levels. Step 3 below relates to the entire "default value" either being not present, null, or an empty list; however the text here (above) relates to the _fields_ of the object within the default value. E.g. the following example:

```graphql
input A {
  b: B = {}
}

input B {
  c: C = { a: null }
}

input C {
  a: A = {}
}
```

is fine since no matter whether we reference type A, B or C initially the chain will terminate at C.a since the default for B.c includes that `C.a` is null. Note: at no point is the default value for a field null or an empty list.

The wording could definitely do with some work though :sweat_smile:

benjie: Scratch that; I think you're right but the wording you've suggested doesn't quite work (see above example). Maybe:

3. If an Input Object references itself either directly or through referenced
   Input Objects:
   1. at least one of the fields in the chain of references must be
     either a nullable or a List type.
   2. at least one of the fields in the chain must have no default value, have a default value of either {null} or an empty list, or have a default value that results in all instances of one of the fields in the chain defaulting to {null} or an empty list)


Bit awkward wording, but hopefully you catch my drift. i.e. allow the following:

```
input A {
  b: B! = {}
}

input B {
  c: C = {}
}

input C {
  a: A! = { b: { c: null} }
}
```

andimarek:
@benjie what about the input object coercion section (in 3.10)? I don't think it currently states that default value of the input object should be coerced.

spec/Section 3 -- Type System.md
leebyron: I'm less concerned about the recursion (nested step 1) and more about the type correctness (nested step 2)

    4. If the input field has a non-null default value:
      ...
      2. {defaultValue} must be compatible with {inputFieldType} as per the
         coercion rules for that type.

This implies that for the example:

```
input Test {
  a: String! = null
}
```

when validating input field `a`, it would see that the default value is `null` and not progress to determine if it is compatible with `String!` (which it would and should fail)

spec/Section 3 -- Type System.md
leebyron: These are very helpful examples! I was definitely too restrictive in my rule set. I think what I currently have implemented in PR would reject all cycles, but would also incorrectly reject this example.

I agree that wording isn't quite right, but it's much closer. I like that this has just one rule about circular references but it makes it clear that there are multiple ways to have illegal circular references - and the above rule can specifically be about type checking the default value

andimarek: just to be clear here:
the coerced default value for `A` =  `{b: {c: {a: {b: {c: null}}}}}`.
the coerced default value for `B` =  `{c: {a: {b: {c: null}}}}`.
the coerced default value for `C` =  `{a: {b: {c: null}}}`

That correct?

andimarek: also: we would have to apply this validation rule also for default variable values, right?

leebyron: I don't think we also need it for variable values or argument defaults since those can only be the entry to a circular ref but not actually contained within a circular ref.

```graphql
input A {
  a: A = {}
}

query Test ($x: A = {}) {}
```

This is problematic because `A.a` creates a circular reference, but `$x` only references it, it isn't contained in it. I think more generally you can say that given a set of Input Objects that have no circular reference, there exists no default value for arguments or variable values that would create a circular reference.

leebyron: > just to be clear here:
> the coerced default value for `A` = `{b: {c: {a: {b: {c: null}}}}}`.
> the coerced default value for `B` = `{c: {a: {b: {c: null}}}}`.
> the coerced default value for `C` = `{a: {b: {c: null}}}`
> 
> That correct?

Coerced default values apply to fields, not types - so:

> the coerced default value for `A.b` = `{c: {a: {b: {c: null}}}}`.
> the coerced default value for `B.c` = `{a: {b: {c: null}}}`.
> the coerced default value for `C.a` = `{b: {c: null}}`

abernix:
We've done a meta-analysis of the schemas we track at Apollo, being careful to respect the privacy of schema owners, and we've found that:

- 6.5% graphs contain object fields with non-leaf argument defaults
- 2.6% graphs contain input objects with non-leaf arguments
- 1.4% graphs contain both

spec/Section 3 -- Type System.md
benjie: Okay I've had a go at writing an algorithm for this; I think the key learning whilst attempting this was that the _act of referencing the default value a second time_ is what triggers a cycle; so the algoritm only tracks when the default value of a field is referenced.

spec/Section 3 -- Type System.md
benjie: This is no longer relevant with the rewriting of the text I think.

benjie:
@leebyron Could you help me turn my new algoritm into more natural spec-algorithm speak? Also since we cannot check the value coerces (because doing so might cause an infinite loop) before we actually check that there's no infinite loop you may note there's a lot of "{blah} must be a {thing}" validation-style assertions... Is this okay? What's the right way to handle this?

benjie:
I've re-written the algorithm to detect cycles in input objects following @leebyron's implementation in https://github.com/graphql/graphql-js/commit/fceda32845ae9950c926cfe6b05af403da63ad60

martinbonnin (approved):

BoD (approved):

benjie (commented):
I've carefully reviewed this against the GraphQL.js implementation in https://github.com/graphql/graphql-js/pull/3814 and it lines up (I did find a mistake in the spec text where we were mutating something though!).

Here's a GraphQL.js schema that demonstrates the fix to the infinite recursion issue I outlined; if you delete the `b: null` then a schema validation error is raised.

```ts
import {
  GraphQLInt,
  GraphQLInputObjectType,
  GraphQLObjectType,
  GraphQLSchema,
  printSchema,
  validateSchema,
  graphql,
} from './npmDist/index.mjs';

const A = new GraphQLInputObjectType({
  name: 'A',
  fields: () => ({
    b: {
      type: B,
      default: { value: {} }, // triggers B's default, which uses A again but safely
    },
  }),
});

const B = new GraphQLInputObjectType({
  name: 'B',
  fields: () => ({
    a: {
      type: A,
      default: { value: { b: null } }, // breaks the cycle
    },
  }),
});

  name: 'Query',
  fields: {
    q: {
      type: GraphQLInt,
      args: {
        a: {
          type: A,
          default: { value: {} }, // kicks off the default chain
        },
      },
    },
  },
});

const schema = new GraphQLSchema({ query: Query });

if (errors.length > 0) throw errors[0];

console.log();
console.log();

  schema,
  source: `{q}`,
  rootValue: {
    q(a) {
      console.log('↓'.repeat(80));
      console.dir(a, { depth: 100 });
      console.log('↑'.repeat(80));
      return 1;
    },
  },
});

console.log();
console.log();

console.dir(result);
```

martinbonnin:
[This will be supported by Apollo Kotlin 5](https://github.com/apollographql/apollo-kotlin/pull/6440)

PR #816, opened 2021-01-18
State that built-in mandatory directives must be omitted

maraisr:
Following on from https://github.com/graphql/graphql-spec/pull/815#issuecomment-760944471, this aims to state that built-in directives must be omitted from the SDL.

cc @benjie @IvanGoncharov


maraisr:
friendly bump @IvanGoncharov @benjie

PR #817, opened 2021-01-28
Comments on InputUnion RFC

ericvergnaud:
Hi, thanks for this extensive and in-depth work.

I will try to keep my comments concise (which might make them sound agressive, apologies if that's the case, it's definitely not the intent)

This PR touches one of the most challengeable founding philosophical choices of GraphQL: having 2 different type systems for input and output.

I understand that the sender -> receiver relation is an unbalanced one: the sender has obligations, the receiver has options.

However:
 - ALL GraphQL users use programming languages (Javascript, Java, C#, Python...) which make no difference between input and output types
 - I actually don't know any other protocol that does that (is there one?)
 - having 2 different type systems for input and output solves (does it?) an implementer problem, not a user problem

The lack of polymorphism on input is only a side-effect of the aformentioned original choice.
In an unreal world, rather than tweak GraphQL to fix that, it would be time for GraphQL2, unifying input and output types (amongst other improvements).

That's very unlikely to happen, but saying so helps forming an opinion on the various proposals:
 - adding yet another polymorphic construct only available on input 'smells' like increasing confusion 
 - it would increase the gap between input and output type systems, rather than reduce it

From there, I think proposal #5 @oneOf is the most useful one:
 - it acts as a constraint on existing type constructs rather than yet another type construct
 - it expresses the required behavior much better than proposal #7

!!! IMPORTANT !!!

Please Read https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md before creating a Pull Request.

### DIFF
--- a/rfcs/DeferStream.md
+++ b/rfcs/DeferStream.md
@@ -25,9 +25,9 @@ Given a query where some fields are expensive and non-essential, non-essential f
 ## Prefetching
 This technique involves optimistically fetching data based on a prediction that a user will execute an action. Prefetching can be one of the most effective ways of reducing latency. However, a significant tradeoff with prefetching that a lot of applications cannot afford is **increased server cost due to incorrect predictions.** With an unsophisticated prefetch algorithm, applications can easily overfetch by a factor of 10 fold.
 
-# Proposal: Incrementally deliver data with @defer and @stream
+# Proposal: Incrementally deliver data with `@defer` and `@stream`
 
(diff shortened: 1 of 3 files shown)


benjie:
Solution 5 is the solution we’re currently pushing forwards; that was the conclusion of the latest input unions working group (which you can watch on the GraphQL Foundation’s YouTube) 👍

If you’d like this feedback factored into the RFC document to help add more reasoning behind solution 5, it should be done in the existing form; for example adding a goal that “No additional types are introduced” and then evaluating it against the various proposals.

rmwhite03 (approved):

leebyron:
Merging since this is RFC Doc and I'm about to move them

PR #828, opened 2021-03-12
Add definition of selectionSet to CreateSourceEventStream

benjie:
Seems like this line is missing (`selectionSet` is otherwise not defined).


leebyron:
Nice catch - good solve!

PR #836, opened 2021-04-06
Add CoerceResult()

leebyron:
This factors out a step from `CompleteValue()` about  "coercing" a value to a formal sub-algorithm `CoerceInternalValue()` that mirrors the language used in `ResolveAbstractType()` - both refer to "internal methods" provided for each type, which is in fact how the reference implementation applies this coercion behavior.

As editorial, this provides a place to expand on the behavior and purpose as well as link back to the relevant subsection from the type system section.

This also makes a *very important* clarification. The previous step read that if coercion failed then "otherwise null". This actually described a much older version of the reference implementation which did not always produce errors for failed coercion. Years back the reference implementation and the spec changed the "result coercion" subsections of all the built-in scalars to clearly state that field errors are thrown if coercion fails. This step now creates ambiguity about which behavior is correct - returning null or throwing a field error? In practice both may be correct because of the error handling behavior - but we should be clear that the returned null is a result of that error behavior rather than it being an explicit behavior that's part of `CompleteValue()`. The new pulled out sub-algo makes it clear that the return type must be valid or a field error is produced - which is exactly the behavior described by each built-in scalar.

Inspired by #780

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -656,8 +656,7 @@ CompleteValue(fieldType, fields, result, variableValues):
       {CompleteValue(innerType, fields, resultItem, variableValues)}, where
       {resultItem} is each item in {result}.
   * If {fieldType} is a Scalar or Enum type:
-    * Return the result of "coercing" {result}, ensuring it is a legal value of
-      {fieldType}, otherwise {null}.
+    * Return the result of {CoerceResult(fieldType, result)}.
   * If {fieldType} is an Object, Interface, or Union type:
     * If {fieldType} is an Object type.
       * Let {objectType} be {fieldType}.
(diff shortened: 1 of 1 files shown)


andimarek:
I agree with that change, but I am not sure about the naming:
- Instead of `CoerceInternalValue` can we name it `ResultCoercion` to make it clear what it is
- internal values sounds more complicated than it is: these are the results of the Resolver. Can we not use internal values?

spec/Section 6 -- Execution.md
```diff
@@ -666,6 +665,23 @@ CompleteValue(fieldType, fields, result, variableValues):
...
+
+See the Scalars [Result Coercion and Serialization](#sec-Scalars.Result-Coercion-and-Serialization)
+sub-section for more detailed information about how GraphQL's built-in scalars
+coerce internal values to GraphQL values.
+
+CoerceInternalValue(leafType, internalValue):
+  * Return the result of calling the internal method provided by the type
+    system for determining the "result coercion" of {leafType} given the value
+    {internalValue}. This internal method must return a valid value for the
+    type or otherwise throw a field error.
```
IvanGoncharov, line 683: I would add that "result coercion" can't be null, otherwise, we are breaking the symmetry of coercing values and completing values. It can happen in the following scenario we have scalar with `parseValue`/`serialize`(graphql-js terminology) so `serialize` is return null, but if you pass `null` as the input value of the same scalar it will be interpreted by input coercing algorithm and will never hit `parseValue`.

leebyron: Great feedback

benjie (approved):

PR #844, opened 2021-04-08
Editorial: Clarify meta-fields in introspection

leebyron:
Derived from #777

Renames "field" to "meta-field" consistently. Adds "with the single exception of the subscription root
operation type". Clarifies some language.

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -89,14 +89,17 @@ warnings.
 
 ## Type Name Introspection
 
-GraphQL supports type name introspection at any point within a query by the
-meta-field `__typename: String!` when querying against any Object, Interface,
-or Union. It returns the name of the object type currently being queried.
+GraphQL supports type name introspection at any selection set within an
+operation with the meta-field `__typename: String!` on any Object,
+Interface, or Union; with the single exception of the subscription root
+operation type. It returns the name of the concrete Object type at that point
+during execution.
 
 This is most often used when querying against Interface or Union types to
-identify which actual type of the possible types has been returned.
+identify which actual Object type of the possible types has been returned.
 
-This field is implicit and does not appear in the fields list in any defined type.
+This meta-field is implicit and does not appear in the fields list in any
+defined type.
 
 
 ## Schema Introspection
@@ -110,8 +113,8 @@ __schema: __Schema!
 __type(name: String!): __Type
 ```
 
-These fields are implicit and do not appear in the fields list in the root type
-of the query operation.
+These meta-fields are implicit and do not appear in the fields list in the root
+type of the query operation.
 
 The schema of the GraphQL schema introspection system:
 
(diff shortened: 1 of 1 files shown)


spec/Section 4 -- Introspection.md
```diff
@@ -89,14 +89,17 @@ warnings.
...
+Interface, or Union; with the single exception of the subscription root
+operation type. It returns the name of the concrete Object type at that point
+during execution.
 
 This is most often used when querying against Interface or Union types to
-identify which actual type of the possible types has been returned.
+identify which actual Object type of the possible types has been returned.
 
-This field is implicit and does not appear in the fields list in any defined type.
+This meta-field is implicit and does not appear in the fields list in any
```
benjie, line 101: Not just this one; all of them:

```suggestion
As a meta-field, `__typename` is implicit and does not appear in the fields list in any
```

spec/Section 4 -- Introspection.md
```diff
@@ -110,8 +113,8 @@ __schema: __Schema!
 __type(name: String!): __Type
 ```
 
-These fields are implicit and do not appear in the fields list in the root type
-of the query operation.
+These meta-fields are implicit and do not appear in the fields list in the root
```
benjie, line 116:
```suggestion
Like all meta-fields, these are implicit and do not appear in the fields list in the root
```

spec/Section 4 -- Introspection.md
```diff
@@ -89,14 +89,17 @@ warnings.
...
 ## Type Name Introspection
 
-GraphQL supports type name introspection at any point within a query by the
-meta-field `__typename: String!` when querying against any Object, Interface,
-or Union. It returns the name of the object type currently being queried.
+GraphQL supports type name introspection at any selection set within an
+operation with the meta-field `__typename: String!` on any Object,
+Interface, or Union; with the single exception of the subscription root
+operation type. It returns the name of the concrete Object type at that point
+during execution.
```
benjie, line 96: This feels a bit clunky; how about:

```suggestion
GraphQL supports type name introspection within any selection set in an
operation, with the single exception of selections at the root of a subscription
operation. Type name introspection is accomplished via the meta-field
`__typename: String!` on any Object, Interface, or Union. It returns the name of
the concrete Object type at that point during execution.
```

I carefully used the term `selections` rather than `selection set` in the above text in one location to make it clear it was transitive (i.e. all nested fragment spreads _also_ cannot query `__typename` on the subscription type).

I also carefully wrote `at the root of a subscription operation` because at least in the GraphQL.js implementation `__typename` can be executed against `Subscription` via `execute()`, just not via `subscribe()` (at least I think this was true last time I tried). I also think it would be awkward to imply that the `__typename` meta-field is omitted from the subscription root operation object type but no other object type - instead it's simply not _accessible_. Presumably you _could_ also reference the `Subscription` type within a `Query` field (in the same way that you can reference `Query` from e.g. the payload type of a mutation), at which point `__typename` would be expected to work on this type.

benjie (changes requested):
I like this improvement to consistency :+1: I've tried to edit the text to make it easier to read and more precise, hopefully this helps :crossed_fingers:

leebyron:
Great suggestions! Applied them all

PR #846, opened 2021-04-08
RFC: Adding a glossary to the GraphQL Specification

benjie:
Since #777 is near to merging, it's time to extract the glossary work into a separate PR.

At this stage, the glossary is incomplete; I suggest that we merge this as an RFC as-is, iterate on it with follow-up PRs (fixing existing text and adding new definitions) and then when it's ready we can propose to the GraphQL Working Group that it gets merged into the specification.

You can read more about the intent of this PR in the included [`rfcs/glossary/README.md` file](https://github.com/benjie/graphql-spec/blob/glossary/rfcs/glossary/README.md).

Relates to GraphQL Spec Working Group action item https://github.com/graphql/graphql-wg/issues/547.

### DIFF
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -1,0 +1,167 @@
+# C. Appendix: Glossary
+
+This appendix seeks to act as a quick reference for some of the technical terms
+used throughout this specification document.
+
+## Glossary is Non-Normative
+
+Where a glossary definition and the usage within the specification do not
+align, this likely indicates a mistake in the glossary - the glossary is
+non-normative and the specification should be treated as the "source of truth."
+
+## Parenthesized Terms
+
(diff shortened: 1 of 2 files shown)


eapache:
> I suggest that we merge this as an RFC as-is, iterate on it with follow-up PRs (fixing existing text and adding new definitions) and then when it's ready we can propose to the GraphQL Working Group that it gets merged into the specification.

👍

leebyron:
Ideally these terms are defined when introduced in the context in which they're most relevant, and the glossary could then be derived from these terms. It would also be really nice to have glossary terms linked in the rendered result.

How do you feel about adding term definition and automatic glossary generation to spec-md?

leebyron:
Merging since this is a RFC doc. I'm in strong support of having better term definition

benjie:
I’d definitely be up for chatting with you about how the spec-md changes might work and then potentially implementing them but I don’t currently know what changes would make sense so I think it warrants discussion. Automated > manual 👍

PR #847, opened 2021-04-09
Clarity in Section 3

sungam3r:
Just a typo.


leebyron:
I think the previous phrasing was correct, but this is slightly simpler to read so happy to take it

PR #848, opened 2021-04-12
fixed introspection field "specifiedBy" to "specifiedByURL"

Code-Hex:
- I've sent PR to graphql/graphql-js is modified introspection field according to "[[RFC] Custom Scalar Specification URLs](https://github.com/graphql/graphql-spec/pull/649)".
  - But I have got a reply from @IvanGoncharov and I understand `specifiedBy` should be `specifiedByURL` to be consistent with `deprecatedReason` representing `@deprecated(reason: "")` directive. (I agreed this direction)
  - https://github.com/graphql/graphql-js/pull/3032#issuecomment-817857967

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -395,7 +395,7 @@ URL via the `@specifiedBy` directive or the `specifiedByURL` introspection field
 This URL must link to a human-readable specification of the data format,
 serialization, and coercion rules for the scalar. For example, a GraphQL service
 providing a `UUID` scalar may link to RFC 4122, or some custom document defining
-a reasonable subset of that RFC. If a scalar `specifiedByURL` is present,
+a reasonable subset of that RFC. If a scalar specification URL is present,
 systems and tools that are aware of it should conform to its described rules.
 
 ```graphql example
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@Code-Hex Thanks for opening it 👍 

Context: https://github.com/graphql/graphql-spec/pull/649#issuecomment-565826631
So naming change from the above suggestion was accepted by the introspection field stayed unchanged.
The only argument I saw is:
> This is a personal feelings, specifiedByURL causes some associations with the Hungarian notation, which has its drawbacks.

But `graphql-js` PR added `specifiedByURL` and I always assumed that #649 is also used that name.
Moreover, other implementations (e.g. Python, graphql-java) followed reference implementation and did `specifiedByURL`.

Also, I think `specifiedByURL` is needed to be consistent with `deprecatedReason` and future directives exposed through introspection.
The original proposal is still `Draft - Stage3` so we can fix it before it gets released.

cc: @leebyron

leebyron:
Thanks for opening this - I thought we had handled this, but perhaps we missed something.

While you're at it, can you make sure prose is updated accordingly?

Example:

> When defining a custom scalar, GraphQL services should provide a specification URL via the `@specifiedBy` directive or the `specifiedBy` introspection field.

Code-Hex:
@leebyron I've done it. Could you review please 🙏

leebyron (approved):

nodkz:
I think you made a typo with `specifiedByURL`, should be `specifiedByUrl`.

GraphQLJSON https://github.com/taion/graphql-type-json/blob/master/src/index.js#L55-L56 and other scalars distributed via NPM now lost `specifiedByUrl` property with v16.0.0-alpha.2

PR #26, opened 2015-07-04
fix #24, a typo in introspection example

chentsulin:

PR #849, opened 2021-04-13
RFC: Allow full unicode range

### LINKED ISSUE #214: Allow unicode characters outside the Basic Multilingual Plane

### LINKED ISSUE #687: allow full unicode range
These are my proposed changes to the spec to allow for full unicode range (currently it is restricted to [BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) code points. See [SourceCharacter](http://spec.graphql.org/draft/#SourceCharacter))

### 1. Change SourceCharacter to allow also code points between 0xFFFF and 0x10FFFF (outside of the BMP):  
```
SourceCharacter ::
  - "U+0009"
  - "U+000A"
  - "U+000D"
  - "U+0020–U+10FFFF"
```
This does not cover all unicode code points: most of the [Control Characters](https://en.wikipedia.org/wiki/ASCII#Control_characters) are not allowed. This is the same behavior as now and I don't see a reason to change it: the only places where Control Characters could be allowed are inside comments or String literals. Inside Strings you can escape them and inside comments they don't really make sense or you can easily work around it.

Changing it to allow for Control Characters to be included would also add an additional burden on systems processing GraphQL documents. Most importantly JSON also requires Control Characters to be escaped (https://tools.ietf.org/html/rfc8259#section-7). 

### 2. Allow surrogate code pair escapes in standard quoted strings:

Currently standard quoted strings allow for BMP code points to be escaped. (Via `\u<4-digit-hex-value>`.) In order to align this with the `SourceCharacter` change above the spec should allow also code points outside of the BMP to be escaped. [Surrogate Pairs](https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Surrogates) are the most direct way to allow for that.

For example the unicode code point U+1F37A ( 🍺 ) which is outside of the BMP  can be escaped as `\ud83c\udf7a`

There are other escapes sequences used for code points outside of the BMP. For example JS and others allow for `\u{1F37A}`. But this would introduce a new syntax. I argue that surrogate code pairs are the most compatible and simplest option. JSON for example understands surrogate code pairs but not `\u{1F37A}`.

One small open question is how illegal surrogate pairs should be handled: 
For example `\ud83c\u0020` or `\uDEAD` is such an illegal pair. 

The [JS spec says](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-ecmascript-language-types-string-type): 

> A code unit that is a leading surrogate or trailing surrogate, but is not part of a surrogate pair, is interpreted as a code point with the same value.

The [JSON spec notes](https://tools.ietf.org/html/rfc8259#section-7):
> However, the ABNF in this specification allows member names and
   string values to contain bit sequences that cannot encode Unicode
   characters; for example, "\uDEAD" (a single unpaired UTF-16
   surrogate).  Instances of this have been observed, for example, when
   a library truncates a UTF-16 string without checking whether the
   truncation split a surrogate pair.  The behavior of software that
   receives JSON texts containing such values is unpredictable; for
   example, implementations might return different values for the length
   of a string value or even suffer fatal runtime exceptions.

I would recommend to add a section that servers should try to reject illegal surrogate pairs if possible in order to avoid unexpected behavior. 

### Previous discussion

Previous Issue about it: #214 
Previous PR which was not merged: #231

Please comment, leave feedback. 

The JS PR for this change is here: https://github.com/graphql/graphql-js/pull/2449

leebyron:
This spec text implements #687 (full context and details there) and also introduces a new escape sequence.

Three distinct changes:

1. Change SourceCharacter to allow points above 0xFFFF, now to 0x10FFFF.
2. Allow surrogate pairs within StringValue. This handles illegal pairs with a parse error.
3. Introduce new syntax for full range code point EscapedUnicode. This syntax (`\u{1F37A}`) has been adopted by many other languages and I propose GraphQL adopt it as well.

(As a bonus, this removes the last instance of a regex in the lexer grammar!)

Reference implementation https://github.com/graphql/graphql-js/pull/3117

Closes #687
Closes #214

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -50,7 +50,7 @@ SourceCharacter ::
 - "U+0009"
 - "U+000A"
 - "U+000D"
-- "U+0020–U+FFFF"
+- "U+0020–U+10FFFF"
 
 GraphQL documents are expressed as a sequence of
 [Unicode](https://unicode.org/standard/standard.html) code points (informally
@@ -812,7 +812,16 @@ StringCharacter ::
 - `\u` EscapedUnicode
 - `\` EscapedCharacter
 
-EscapedUnicode :: /[0-9A-Fa-f]{4}/
+EscapedUnicode ::
+
+- HexDigit HexDigit HexDigit HexDigit
+- `{` HexDigit+ `}` "but only if <= 0x10FFFF"
+
+HexDigit :: one of
+
+- `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`
+- `A` `B` `C` `D` `E` `F`
+- `a` `b` `c` `d` `e` `f`
 
 EscapedCharacter :: one of `"` `\` `/` `b` `f` `n` `r` `t`
 
@@ -897,7 +906,24 @@ StringValue :: `""`
 
 StringValue :: `"` StringCharacter+ `"`
 
-- Return the sequence of all {StringCharacter} code points.
+- Let {string} be the sequence of all {StringCharacter} code points.
+- For each {codePoint} at {index} in {string}:
+  - If {codePoint} is >= 0xD800 and <= 0xDBFF (a
+    [_High Surrogate_](https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs)):
+    - Let {lowPoint} be the code point at {index} + {1} in {string}.
+    - Assert {lowPoint} is >= 0xDC00 and <= 0xDFFF (a
+      [_Low Surrogate_](https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs)).
+    - Let {decodedPoint} = ({codePoint} - 0xD800) × 0x400 + ({lowPoint} -
+      0xDC00) + 0x10000.
+    - Within {string}, replace {codePoint} and {lowPoint} with {decodedPoint}.
+  - Otherwise, assert {codePoint} is not >= 0xDC00 and <= 0xDFFF (a
+    [_Low Surrogate_](https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs)).
+- Return {string}.
+
+Note: {StringValue} should avoid encoding code points as surrogate pairs. While
+services must interpret them accordingly, a braced escape (for example
+`"\u{1F4A9}"`) is a clearer way to encode code points outside of the
+[Basic Multilingual Plane](https://unicodebook.readthedocs.io/unicode.html#bmp).
 
 StringCharacter :: SourceCharacter but not `"` or `\` or LineTerminator
 
@@ -905,8 +931,9 @@ StringCharacter :: SourceCharacter but not `"` or `\` or LineTerminator
 
 StringCharacter :: `\u` EscapedUnicode
 
-- Let {value} be the 16-bit hexadecimal value represented by the sequence of
-  hexadecimal digits within {EscapedUnicode}.
+- Let {value} be the 21-bit hexadecimal value represented by the sequence of
+  {HexDigit} within {EscapedUnicode}.
+- Assert {value} <= 0x10FFFF.
 - Return the code point {value}.
 
 StringCharacter :: `\` EscapedCharacter
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -45,32 +45,22 @@ match, however some lookahead restrictions include additional constraints.
 
 ## Source Text
 
-SourceCharacter ::
+SourceCharacter :: "Any Unicode scalar value"
 
-- "U+0009"
-- "U+000A"
-- "U+000D"
-- "U+0020–U+10FFFF"
+GraphQL documents are interpreted from a source text, which is a sequence of
+{SourceCharacter}, each {SourceCharacter} being a _Unicode scalar value_ which
+may be any Unicode code point from U+0000 to U+D7FF or U+E000 to U+10FFFF
+(informally referred to as _"characters"_ through most of this specification).
 
-GraphQL documents are expressed as a sequence of
-[Unicode](https://unicode.org/standard/standard.html) code points (informally
-referred to as _"characters"_ through most of this specification). However, with
-few exceptions, most of GraphQL is expressed only in the original non-control
-ASCII range so as to be as widely compatible with as many existing tools,
-languages, and serialization formats as possible and avoid display issues in
-text editors and source control.
+A GraphQL document may be expressed only in the ASCII range to be as widely
+compatible with as many existing tools, languages, and serialization formats as
+possible and avoid display issues in text editors and source control. Non-ASCII
+Unicode scalar values may appear within {StringValue} and {Comment}.
 
-Note: Non-ASCII Unicode characters may appear freely within {StringValue} and
-{Comment} portions of GraphQL.
-
-### Unicode
-
-UnicodeBOM :: "Byte Order Mark (U+FEFF)"
-
-The "Byte Order Mark" is a special Unicode character which may appear at the
-beginning of a file containing Unicode which programs may use to determine the
-fact that the text stream is Unicode, what endianness the text stream is in, and
-which of several Unicode encodings to interpret.
+Note: An implementation which uses _UTF-16_ to represent GraphQL documents in
+memory (for example, JavaScript or Java) may encounter a _surrogate pair_. This
+encodes a _supplementary code point_ and is a single valid source character,
+however an unpaired _surrogate code point_ is not a valid source character.
 
 ### White Space
 
@@ -174,6 +164,17 @@ which appear in {Ignored} may also appear within a lexical {Token} in a
 significant way, for example a {StringValue} may contain white space characters.
 No {Ignored} may appear _within_ a {Token}, for example no white space
 characters are permitted between the characters defining a {FloatValue}.
+
+**Byte order mark**
+
+UnicodeBOM :: "Byte Order Mark (U+FEFF)"
+
+The _Byte Order Mark_ is a special Unicode code point which may appear at the
+beginning of a file which programs may use to determine the fact that the text
+stream is Unicode, and what specific encoding has been used.
+
+As files are often concatenated, a _Byte Order Mark_ may appear anywhere within
+a GraphQL document and is {Ignored}.
 
 ### Punctuators
 
@@ -814,8 +815,8 @@ StringCharacter ::
 
 EscapedUnicode ::
 
+- `{` HexDigit+ `}`
 - HexDigit HexDigit HexDigit HexDigit
-- `{` HexDigit+ `}` "but only if <= 0x10FFFF"
 
 HexDigit :: one of
 
@@ -830,19 +831,58 @@ BlockStringCharacter ::
 - SourceCharacter but not `"""` or `\"""`
 - `\"""`
 
-Strings are sequences of characters wrapped in quotation marks (U+0022). (ex.
-{`"Hello World"`}). White space and other otherwise-ignored characters are
-significant within a string value.
+{StringValue} is a sequence of characters wrapped in quotation marks (U+0022).
+(ex. {`"Hello World"`}). White space and other characters ignored in other parts
+of a GraphQL document are significant within a string value.
+
+A {StringValue} is evaluated to a Unicode text value, a sequence of Unicode
+scalar values, by interpreting all escape sequences using the static semantics
+defined below.
 
 The empty string {`""`} must not be followed by another {`"`} otherwise it would
 be interpreted as the beginning of a block string. As an example, the source
 {`""""""`} can only be interpreted as a single empty block string and not three
 empty strings.
 
-Non-ASCII Unicode characters are allowed within single-quoted strings. Since
-{SourceCharacter} must not contain some ASCII control characters, escape
-sequences must be used to represent these characters. The {`\`}, {`"`}
-characters also must be escaped. All other escape sequences are optional.
+**Escape Sequences**
+
+In a single-quoted {StringValue}, any Unicode scalar value may be expressed
+using an escape sequence. GraphQL strings allow both C-style escape sequences
+(for example `\n`) and two forms of Unicode escape sequences: one with a
+fixed-width of 4 hexadecimal digits (for example `\u000A`) and one with a
+variable-width most useful for representing a _supplementary character_ such as
+an Emoji (for example `\u{1F4A9}`).
+
+The hexadecimal number encoded by a Unicode escape sequence must describe a
+Unicode scalar value, otherwise parsing should stop with an early error. For
+example both sources `"\uDEAD"` and `"\u{110000}"` should not be considered
+valid {StringValue}.
+
+Escape sequences are only meaningful within a single-quoted string. Within a
+block string, they are simply that sequence of characters (for example
+`"""\n"""` represents the Unicode text [U+005C, U+006E]). Within a comment an
+escape sequence is not a significant sequence of characters. They may not appear
+elsewhere in a GraphQL document.
+
+Since {StringCharacter} must not contain some characters, escape sequences must
+be used to represent these characters. All other escape sequences are optional
+and unescaped non-ASCII Unicode characters are allowed within strings. If using
+GraphQL within a system which only supports ASCII, then escape sequences may be
+used to represent all Unicode characters outside of the ASCII range.
+
+For legacy reasons, a _supplementary character_ may be escaped by two
+fixed-width unicode escape sequences forming a _surrogate pair_. For example the
+input `"\uD83D\uDCA9"` is a valid {StringValue} which represents the same
+Unicode text as `"\u{1F4A9}"`. While this legacy form is allowed, it should be
+avoided as a variable-width unicode escape sequence is a clearer way to encode
+such code points.
+
+When producing a {StringValue}, implementations should use escape sequences to
+represent non-printable control characters (U+0000 to U+001F and U+007F to
+U+009F). Other escape sequences are not necessary, however an implementation may
+use escape sequences to represent any other range of code points. If an
+implementation chooses to escape a _supplementary character_, it should not use
+a fixed-width surrogate pair unicode escape sequence.
 
 **Block Strings**
 
@@ -898,7 +938,13 @@ Note: If non-printable ASCII characters are needed in a string value, a standard
 quoted string with appropriate escape sequences must be used instead of a block
 string.
 
-**Semantics**
+**Static Semantics**
+
+A {StringValue} describes a Unicode text value, a sequence of *Unicode scalar
+value*s. These semantics describe how to apply the {StringValue} grammar to a
+source text to evaluate a Unicode text. Errors encountered during this
+evaluation are considered a failure to apply the {StringValue} grammar to a
+source and result in a parsing error.
 
 StringValue :: `""`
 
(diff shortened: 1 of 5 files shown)


spec/Section 2 -- Language.md
```diff
@@ -892,32 +900,47 @@ StringValue :: `""`
...
+    * If {codePoint} is >= 0xD800 and <= 0xDBFF (a [*High Surrogate*](https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs)):
+      * Let {lowPoint} be the code point at {index} + {1} in {string}.
+      * Assert {lowPoint} is >= 0xDC00 and <= 0xDFFF (a [*Low Surrogate*](https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs)).
+      * Let {decodedPoint} = ({codePoint} - 0xD800) × 0x400 + ({lowPoint} - 0xDC00) + 0x10000.
+      * Within {string}, replace {codePoint} and {lowPoint} with {decodedPoint}.
+    * Else assert {codePoint} is not >= 0xDC00 and <= 0xDFFF (a [*Low Surrogate*](https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs)).
+  * Return {string}.
+
+Note: {StringValue} should avoid encoding code points as surrogate pairs.
+While services must interpret them accordingly, a bracked escape (for example
```
benjie, line 914:
```suggestion
While services must interpret them accordingly, a braced escape (for example
```

benjie: Or `bracketed`

leebyron: thank you for the careful read - lol "bracked"

andimarek:
One thing I am not sure yet: should a single surrogate pair be really an error. 

Wikipedia says: 

t is possible to unambiguously encode an unpaired surrogate (a high surrogate code point not followed by a low one, or a low one not preceded by a high one) in the format of UTF-16 by using a code unit equal to the code point. The result is not valid UTF-16, but the majority of UTF-16 encoder and decoder implementations do this then when translating between encodings.[citation needed]

(But without clear citation as you can see)

Looking at the Java and JS internal encoder it seems that they don't throw errors which is consistent with the wiki page.

So I am asking if we are trying to be to strict here. As a side effect: not handling single surrogate code points as errors would also simplify implementations.

leebyron:
I'll be honest that this is on the edge of my expertise with Unicode. I did a ton of reading while working on this and stupidly did not take any notes. I'll try to retrace some steps.

Here's an excerpt from the [Unicode FAQ](https://unicode.org/faq/utf_bom.html):

> Q: How do I convert a UTF-16 surrogate pair such as <D800 DC00> to UTF-8? As one 4-byte sequence or as two separate 3-byte sequences?
>
> A: The definition of UTF-8 requires that supplementary characters (those using surrogate pairs in UTF-16) be encoded with a single 4-byte sequence. However, there is a widespread practice of generating pairs of 3-byte sequences in older software, especially software which pre-dates the introduction of UTF-16 or that is interoperating with UTF-16 environments under particular constraints. Such an encoding is not conformant to UTF-8 as defined. See UTR #26: Compatability Encoding Scheme for UTF-16: 8-bit (CESU) for a formal description of such a non-UTF-8 data format. When using CESU-8, great care must be taken that data is not accidentally treated as if it was UTF-8, due to the similarity of the formats. [AF]
> 
> Q: How do I convert an unpaired UTF-16 surrogate to UTF-8?
>
> A different issue arises if an unpaired surrogate is encountered when converting ill-formed UTF-16 data. By representing such an unpaired surrogate on its own as a 3-byte sequence, the resulting UTF-8 data stream would become ill-formed. While it faithfully reflects the nature of the input, Unicode conformance requires that encoding form conversion always results in a valid data stream. Therefore a converter must treat this as an error. [AF]


That is about UTF16 and UTF8 conversions, which is not exactly the domain we are working in the spec. GraphQL does not have an official encoding, it stays at the abstraction of a "sequence of code points"

I suppose if your implementation relied on UTF8 under the hood then you would expect the behavior described here, only an error would be appropriate when interpreting the string value. If your implementation relied on "older software" UTF16, as both Java and JavaScript do, then you might just expect them to pass through without error, since that's what the host environment does.

---

I think either way we need to solve for:

Given a string which includes a surrogate pair (directly or escaped), we need to describe the string value of code points in a well defined way.

So we will have to have this bit of complexity one way or another. Our choice is then what to do with unpaired surrogates:

1. Treat them as an error (current)
2. Pass them through as code points (potentially invalid for some systems?)
3. Leave it to the implementation (ie. "A service may raise an error or pass them through")

I'm not sure which of the latter two you'd suggest, though I think we may need to call in for more Unicode brain power 🧠

andimarek:
I am very happy if we bring in more unicode expertise: same as you I am not absolutely comfortable with making this decision.

But I have a preference: I would leave it up to the implementation (option 3). 

I will try to explain why: Unicode itself is mixing up encoding and code points by having surrogate code points embedded into the BMP in order to support UTF-16 encoding. (The reason why is afaiu to preserve backwards compatibility when they expanded the amount of possible code points in Unicode 2.0)

This is why we can't cleanly separate encoding and a more abstract "sequence of code points": we have to talk about surrogates even when we try to ignore encoding. I think error handling is an implementation detail of the encoding, even if we can talk about surrogates on the abstract level.

Also: I am sorry to add even more nuances, but if we would go down option 1 we would also need to specify what error means: the current spec text implies an error is thrown (by asserting something) which doesn't have to be always the case:

Here is quote from [Unicode 13 chapter 3](https://www.unicode.org/versions/Unicode13.0.0/ch03.pdf)

> For example, in UTF-8 every code unit of the form 110xxxx2 must be followed
by a code unit of the form 10xxxxxx2. A sequence such as 110xxxxx2 0xxxxxxx2 is
ill-formed and must never be generated. When faced with this ill-formed code
unit sequence while transforming or interpreting text, a conformant process
must treat the first code unit 110xxxxx2 as an illegally terminated code unit
sequence—for example, by signaling an error or representing the code unit
with a marker such as U+FFFD replacement character

This explains how errors could be represented as replacement characters without aborting everything: 

Also another quote from Wikipedia (https://en.wikipedia.org/wiki/UTF-8) which also talks about replacement characters instead of throwing errors/signaling errors:

> Some implementations of decoders throw exceptions on errors.[27] This has the disadvantage that it can turn what would otherwise be harmless errors (such as a "no such file" error) into a denial of service. For instance early versions of Python 3.0 would exit immediately if the command line or environment variables contained invalid UTF-8.[28] An alternative practice is to replace errors with a replacement character. Since Unicode 6[29] (October 2010), the standard (chapter 3) has recommended a "best practice" where the error ends as soon as a disallowed byte is encountered. In these decoders E1,A0,C0 is two errors (2 bytes in the first one). This means an error is no more than three bytes long and never contains the start of a valid character, and there are 21,952 different possible errors.[30] The standard also recommends replacing each error with the replacement character "�" (U+FFFD).

leebyron:
On errors: We should be clear that what we're doing here is not UTF encoding/decoding, so I don't think U+FFFE replacements would be a reasonable thing to do. We are defining an interpreted string value which is part of the static semantics of grammar parsing, so I think the only reasonable way to represent an error is to fail to parse (parse error).

---

The more I think about it, the more I think explicit errors for invalid input is the most appropriate behavior. Letting it be defined by the implementation could be unsafe - different systems could exhibit different behavior and if done incorrectly could result in bad behavior or taking out a service as an attack vector if it hard fails on invalid Unicode text. 

If we explicitly define invalid input as a parse error then we make these attacks impossible (if a parser is spec compliant at least)

---

I think I'm still not being crisp about Unicode concepts in these spec changes and that's leading to some suboptimal results.

Right now we spec SourceCharacter as any Unicode code point and we spec a string value as a sequence of code points. However maybe we should be specifying one or both as Unicode scalar values (which Unicode defines as any code point except a surrogate). The logic I've proposed for string value is assuming that only scalar values are appropriate and need to deal with potential surrogates in the lexical input stream. 

An alternative would be to define both source and strings as sequences of Scalar Values (this is what modern languages like Swift and Rust do). 

That would still require the parser to validate a correct stream of values, but hopefully that would be simpler to implement since it isn't intermixed with escape sequences.

Then string value semantics only need to deal with surrogates specifically for Unicode code point escape sequences. I think your initial proposal here to narrowly allow escape sequences surrogate pairs is appropriate. We _could_ just outright disallow encoded surrogates, but considering how common this is in other encoders and likely already happening in existing GraphQL queries for emoji, I think it's smart to explicitly include it. 

However I might propose that the `\u{1234}` syntax, since its new, be explicitly limited to scalar values.

andimarek:
I see allowing only Unicode Scalar Values instead of code points as a valid strategy: we already restrict code points to dissallow certain control characters. I don't see any value in allowing surrogates unencoded.

As you mentioned the tricky part is if you allow surrogate pairs as encodes pairs: graphql.js and graphql Java both support surrogate pairs at the moment.

leebyron:
I think we would definitely allow that, but the interpretation isn't that we have two Unicode surrogate code points in the "source text" that we then have to specify how to handle, a surrogate pair in Java/JavaScript is actually _one_ Unicode scalar value code point in the "source text". In other words, the fact that Java/JavaScript encode strings as UTF-16 is below the abstraction of the specification.

leebyron:
I'm going to take another revision on this RFC in the next couple days to try out these ideas. I think I will need to add a decent amount of notes to help guide through the Unicode terms 😅

andimarek:
> but the interpretation isn't that we have two Unicode surrogate code points in the "source text" that we then have to specify how to handle, a surrogate pair in Java/JavaScript is actually one Unicode scalar value code point in the "source text". In other words, the fact that Java/JavaScript encode strings as UTF-16 is below the abstraction of the specification.

I am not sure I understand what you mean exactly: we would dissallow surrogate code points in the sources text. They are simply not allowed to appear.
BUT: we allow valid surrogate pair encodings like `\ud83c\udf7a` which is interpreted as a Unicode Scalar Value. 
How the implementations store these code points (UTF-8,16 etc) is not relevant.

That what you mean?

leebyron:
Not quite what I mean. We should allow code points above U+FFFF, and should not depend on any particular encoding but instead operate on "Unicode text". If an implementation is based on UTF-16 strings, this shouldn't mean that such an implementation cannot use astral plane code points, it just means that it needs to encode those as surrogate pairs in UTF-16. But GraphQL syntax does not care about this implementation specific encoding, as long as it represents a sequence of Unicode scalar values.

andimarek:
I don't think this will give us a coherent picture, because we can't ignore the fact that surrogates are valid code
Points in the BMP (but not valid scalar values). 

We have to adress that I think. This what I meant with us not being able to just say: we don't care about encoding. Surrogates are encoding specific code points embedded into BMP, which means even if we say we don't care about encoding, we have to adress them. 

(Disclaimer again:
Not a Unicode expert so we should try to get another opinion.)

andimarek:
Actually not sure we really talked about the same thing :-). So let me be very specific what I thought could be a solution (based on how I understood your idea):

we allow only a restricted set of code points as `SourceCharacter`:

```
SourceCharacter ::
  - "U+0009"
  - "U+000A"
  - "U+000D"
  - "U+0020–U+D7FF" 
  -  "U+E000-U+10FFFF"
  ```

This means every surrogate code point is illegal and needs to be rejected.

But: we allow surrogate pair encodings like `\ud83c\udf7a` which needs to be translated as `U+1F37A`, which is a valid `SourceCharacter` code point. 

For braced encodings we only allow the code points specified in `SourceCharacter`. So `\u{ud83c}` or `\u{d83c}\{udf7a}` would be illegal and need to be rejected.

**Comment by leebyron** [comment]
Almost exactly right. 

> braced encodings we only allow the code points specified in SourceCharacter

Braced encodings we only allow Unicode scalar values (eg `\u{0001}` is valid)

---

My only clarification is that a UTF-16 based implementation may use surrogate pair _code units_ to encode an astral plane _code point_. That would be interpreted as a single Unicode scalar value in a GraphQL source text, and thus a valid SourceCharacter and character in a string value.

This has no impact on spec language since the spec itself should not care how code points in a source text are encoded, but it may be worth a note since it would be an easy mistake to fail on otherwise valid UTF-16 input.

**Comment by andimarek** [comment]
> Braced encodings we only allow Unicode scalar values (eg \u{0001} is valid)

In general I think I am fine with that, but we should talk JSON encoding of Strings here too.

Specifically how Introspection values should be encoded via JSON (not sure we do this at the moment).

JSON does not allow braced unicode espaces so a braced escape would need to printed as normal unicode espace if you want to feed it back into GraphQL.

**Comment by leebyron** [comment]
I think that's fine, and I'm not sure introspection is any different from any other query that can return a string value, maybe we need a note that a serialization format needs to support representing all possible string values? (Which JSON does)

**Comment by bhamiltoncx** [comment]
I'm very supportive of this RFC. Thanks!

> we allow only a restricted set of code points as `SourceCharacter`:
> 
> ```
> SourceCharacter ::
>   - "U+0009"
>   - "U+000A"
>   - "U+000D"
>   - "U+0020–U+D7FF" 
>   - "U+E000-U+10FFFF"
> ```
> 
> This means every surrogate code point is illegal and needs to be rejected.
>
> For braced encodings we only allow the code points specified in `SourceCharacter`. So `\u{ud83c}` or `\u{d83c}\u{udf7a}` would be illegal and need to be rejected.

+1 to this; we should explicitly forbid surrogates (dangling or paired) in GraphQL.

We should probably be clear about what happens if a SourceCharacter with a special meaning is expressed with braced encoding — e.g., if someone does:

```
\u{007b}
field
\u{007d}
```

is it the same as:

```
{
field
}
```

e.g., should GraphQL parsers decode Unicode *before* or *after* lexical evaluation?
 
> But: we allow surrogate pair encodings like `\ud83c\udf7a` which needs to be translated as `U+1F37A`, which is a valid `SourceCharacter` code point.

IMHO, JSON encoding is its own thing and should be dealt with using the JSON standards.

I think it goes without saying that any encoding used to transport GraphQL must support the full range of Unicode, but it might be worth mentioning that legacy transports which don't support Unicode are not supported by GraphQL. (Hopefully there aren't any such transports in modern use.)

**Comment by andimarek** [comment]
Regarding lees comment: 
> My only clarification is that a UTF-16 based implementation may use surrogate pair code units to encode an astral plane code point. That would be interpreted as a single Unicode scalar value in a GraphQL source text, and thus a valid SourceCharacter and character in a string value.

I think in order to make this clear we could talk about String literal evaluation. Similar to the JS spec:
> String literals evaluate to ECMAScript String values.

I would recommend that we adapt a similar model where we say that String literals are evaluated to sequences of Unicode Value Scalars (horrible naming overload here whith our GraphQL Scalars), but the details of how this evaluated String is hold/stored/processed is up to the implemtentation (can be UTF-16 or UTF-8 or whatever).

I hope that reflects the intention of you comment above @leebyron

**Comment by leebyron** [comment]
Updated this RFC, reference implementation coming soon.

Updates:

* SourceCharacter is now "Any Unicode scalar value". Previously we forbid the ASCII-range control characters from sources, but I think the rationale for this was somewhat flimsy since we don't forbid any other control character. This should simplify.
* Added updated semantic for expressing escape sequences as Unicode scalar values only while including the legacy surrogate pair escape sequence.
* Moved the BOM section to a subsection and rewrote.
* Added a subsection on "Escape Sequences" to elaborate on usage and evaluation rules

**Comment by leebyron** [comment]
**On supporting surrogate pair escape sequences (eg `"\uD83D\uDCA9"`):**

I agree with @bhamiltoncx that ideally we should explicitly forbid surrogates anywhere in GraphQL, however part of the discussion that led us to this RFC is that many popular GraphQL libraries have been implicitly supporting this pattern for years (since many existing implementations actually have been parsing UTF16 code values and building strings the same way). I think we need to ensure this RFC is not going to cause any existing valid documents to become invalid, and I suspect there are plenty out there encoding Emoji in this form. In the spec change I tried to make it clear that this pattern is legacy and should be avoided.

**On escapes within triple quote Strings:**

Block quote strings were intended to intentionally exclude escape sequences so that the `\` character can appear freely. It's primary purpose is for documentation so being legible in source form is more important than its expressivity for arbitrary text.

File: spec/Section 2 -- Language.md [hunk 2]
```diff
@@ -47,31 +47,31 @@ match, however some lookahead restrictions include additional constraints.
...
+interpreted as if it was an sequence of {SourceCharacter} Unicode scalar values.
 
+As an example, implementations which use [UTF-16](https://www.unicode.org/glossary/#UTF_16)
+to store GraphQL documents (for example, JavaScript or Java) may natively encode
+[supplementary code points](https://www.unicode.org/glossary/#supplementary_code_point)
+(those above U+FFFF) as a [surrogate pair](https://www.unicode.org/glossary/#surrogate_pair).
+Surrogate code points are not Unicode scalar values, but it is compliant to
+interpret such surrogate pairs as single scalar values within the sequence of
+{SourceCharacter}. However individual surrogate code points must not be
+contained within this sequence.
```
leebyron, line 69: @andimarek this paragraph is based on our discussion around abstracting away an implementation specific encoding which may itself contain surrogate pairs

spec/Section 2 -- Language.md
```diff
@@ -815,27 +826,66 @@ StringCharacter ::
...
 
-Strings are sequences of characters wrapped in quotation marks (U+0022).
-(ex. {`"Hello World"`}). White space and other otherwise-ignored characters are
-significant within a string value.
+{StringValue} is a sequence of characters wrapped in quotation marks (U+0022).
+(ex. {`"Hello World"`}). White space and other characters ignored in other parts
+of a GraphQL document are significant within a string value.
+
+A {StringValue} is evaluated to a Unicode text value, a sequence of Unicode
+scalar values, using the semantics defined below.
```
leebyron, line 849: @andimarek this is based on your suggestion to be explicit that the grammar source and the encoded string value are not the same

spec/Section 2 -- Language.md
andimarek: I find this confusing to read and understand tbh. I will try to unpcack my thinking:

It is true that JS/Java use UTF-16 internally to store Strings. But this is purely a runtime, in memory representation.

If I read a Document via IO (http request, file) I need to assume an encoding of this document to interpret the bytes I get. Lets say it is UTF-8.

UTF-8 encoded Document -> Internal JS UTF-16 representation of the document as String -> GraphQL Parser/Lexer -> Internal JS UTF-16 representation of evaluated Strings

This means the GraphQL Lexer always operates on JS UTF-16 [code units](https://unicode.org/glossary/#code_unit) and code points outside of the BMP are always representated as surrogate pairs in memory (for JS/Java at least).

Are you trying to say that this internal UTF-16 representation of documents is acceptable and doesn't conflict with the definition of `SourceCharacter`?

spec/Section 2 -- Language.md
andimarek: @leebyron can we make clear what evaluate means: unescaping the literal by interpreting the escape sequences?

spec/Section 2 -- Language.md
leebyron: I'm definitely open to alternative wordings. What I wrote makes sense to me of course, but whatever makes this more clear I'm definitely open to. Let me know what you think.

> Are you trying to say that this internal UTF-16 representation of documents is acceptable and doesn't conflict with the definition of SourceCharacter?

Correct. For example, in saying that GraphQL accepts any Unicode Scalar Value it should be valid to provide U+1F600 as a Source Character. However if your implementation uses UTF-16, then 0x1F600 does not fit in 16 bits, so UTF-16 defines that this code point should be represented by the code unit pair 0xD800, 0xDC00, and thus these two code units will appear in the input for a GraphQL source text. These two code units do not define Unicode Scalar Values since they are surrogates, so then does this result in an input that is no longer valid GraphQL source text? That can't be the case since we claim that both a source text is a sequence of "any Unicode Scalar Value" *and* "GraphQL documents may be stored and transmitted in any encoding." If both of these are true, then we must accept such an input. The way we accept it is that a GraphQL source text "should be
interpreted as if it was an sequence of {SourceCharacter} Unicode scalar values" and the UTF-16 specific implementation detail of a surrogate pair (0xD800, 0xDC00) encodes one Unicode scalar value (U+1F600) in that sequence.

andimarek: My feeling is that we are making it a bit more complicated than it is: 

Maybe we could simply highlight that `SourceCharacter` is a "logical or semantic view" indepdended of the actual encoding used in the implementation, because they are code points and not code units with a link to the Unicode spec.

leebyron: I'm definitely okay with simplifying the language, and what you're suggesting here is definitely accurate. My concern was that we would need this guidance for implementers, but maybe the reference implementation is enough (would love your thoughts on whether the documentation within that new PR is clear enough - https://github.com/graphql/graphql-js/pull/3117/)

andimarek: I would recommend to simplify this here and depending on the feedback we get we can expand it again. There reference impl. is probably a better place for the details as you pointed out.

spec/Section 2 -- Language.md
```diff
@@ -47,31 +47,22 @@ match, however some lookahead restrictions include additional constraints.
...
-UnicodeBOM :: "Byte Order Mark (U+FEFF)"
-
-The "Byte Order Mark" is a special Unicode character which
-may appear at the beginning of a file containing Unicode which programs may use
-to determine the fact that the text stream is Unicode, what endianness the text
-stream is in, and which of several Unicode encodings to interpret.
+Note: An implementation which uses *UTF-16* to represent GraphQL documents in
+memory (for example, JavaScript or Java) may encounter a *surrogate pair*. This
+encodes a *supplementary code point* and is a single valid source character,
+however an unpaired *surrogate code point* is not a valid source character.
```
leebyron, line 65: @andimarek - how do you feel about this?

spec/Section 2 -- Language.md
```diff
@@ -178,6 +169,16 @@ significant way, for example a {StringValue} may contain white space characters.
...
 
+**Byte order mark**
+
+UnicodeBOM :: "Byte Order Mark (U+FEFF)"
+
+The *Byte Order Mark* is a special Unicode code point which may appear at the
+beginning of a file which programs may use to determine the fact that the text
+stream is Unicode, and what specific encoding has been used.
+
+As files are often concatenated, a *Byte Order Mark* may appear anywhere within
```
andimarek, line 180: I don't have enough background to judge that. How often and in which ares is that used these days that we need to consider that?

spec/Section 2 -- Language.md
```diff
@@ -47,31 +47,22 @@ match, however some lookahead restrictions include additional constraints.
 ## Source Text
 
 SourceCharacter ::
-  - "U+0009"
-  - "U+000A"
-  - "U+000D"
-  - "U+0020–U+FFFF"
+  - "Any Unicode scalar value"
```
andimarek, line 50: This is the one change I can see so far which I am not sure about.
I can see how you want to escape control character in a String, because you really want to be able to express ALL unicode scalar values. 
But why unescaped in a document itself? In 6 years GraphQL Java nobody ever wanted to that afaik.

andimarek (commented):
I have two open questions, see inline comments.

spec/Section 2 -- Language.md
leebyron: UTF-8 doesn't need a BOM, so since that's become the dominant encoding this comment is a little anachronistic (BOM is common in other encodings like UTF-16), but I think still important for completeness.

This same clause appears in the JavaScript spec for similar reasons.

spec/Section 2 -- Language.md
leebyron: I think this earlier change to exclude control characters was well intentioned but ultimately more costly than helpful.

In practice I agree with your point, but there are corner cases which have caused real confusion.

First, it only excluded the ascii-range control chars, and not the control chars > 0x7F, so while the intent was to exclude control chars, in practice there were others allowed.

But I think more importantly, the real cost is being different from other languages as a source of complexity and confusion. No other modern language has this kind of source level exclusion, and this ends up being a land-mine for dynamic queries with string concatenation which need to then remember to do additional input sanitation.

Ultimately this previous attempt to exclude control chars was just a source of implementation complexity of dubious value. My intent with the change is to make Unicode source chars no different from nearly all other modern languages.

spec/Section 2 -- Language.md
leebyron: I should point out that this is an editorial change. The current spec has this same clause, it was just in an awkward location before we had a space to discuss Unicode support directly.

spec/Section 2 -- Language.md
```diff
@@ -815,27 +816,73 @@ StringCharacter ::
...
 EscapedCharacter :: one of `"` `\` `/` `b` `f` `n` `r` `t`
 
 BlockStringCharacter ::
   - SourceCharacter but not `"""` or `\"""`
   - `\"""`
 
-Strings are sequences of characters wrapped in quotation marks (U+0022).
-(ex. {`"Hello World"`}). White space and other otherwise-ignored characters are
-significant within a string value.
+{StringValue} is a sequence of characters wrapped in quotation marks (U+0022).
```
andimarek, line 834: It is getting a bit meta but I don't think character is completely correct here. It is either a sequence of Unicode Scalar Values or a sequence of Code Points depending on how you look at it.

spec/Section 2 -- Language.md
andimarek: Yeah ... GraphQL Java actually already handles that :-)

spec/Section 2 -- Language.md
andimarek: Thanks for explaining. I am happy with it.

PR #858, opened 2021-04-23
Add spellcheck to test suite

leebyron:
Adds `cspell` and an extra word config, adding it to the test suite

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,11 +13,14 @@
     "url": "http://github.com/graphql/graphql-spec.git"
   },
   "scripts": {
-    "test": "spec-md spec/GraphQL.md > /dev/null",
+    "test": "npm run test:build && npm run test:spellcheck",
+    "test:build": "spec-md spec/GraphQL.md > /dev/null",
+    "test:spellcheck": "cspell 'spec/**/*.md'",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -13,11 +13,14 @@
     "url": "http://github.com/graphql/graphql-spec.git"
   },
   "scripts": {
-    "test": "spec-md spec/GraphQL.md > /dev/null",
+    "test": "npm run test:build && npm run test:spellcheck",
+    "test:build": "spec-md spec/GraphQL.md > /dev/null",
+    "test:spellcheck": "cspell 'spec/**/*.md'",
```
benjie, line 18: Might as well spellcheck the README and the RFCs too?

benjie (approved):

benjie:
(My review assumes that you also did a search for the replaced text to ensure nothing escaped; I can't easily do that through the GitHub interface.)

package.json
leebyron: README yes. RFCs have a zillion spurious errors (like flagging github usernames) so I'll exclude those

leebyron:
> (My review assumes that you also did a search for the replaced text to ensure nothing escaped; I can't easily do that through the GitHub interface.)

Yep, I applied the changes manually rather than using any automated replacements to ensure nothing broke

Issue #826, opened 2021-03-07
Question about scalars


sungam3r:
See discussion in https://github.com/graphql-dotnet/graphql-dotnet/pull/2357#discussion_r588738887

The spec says 
![изображение](https://user-images.githubusercontent.com/21261007/110234641-a2bd1c80-7f3c-11eb-8f67-6fd361bac466.png)
but technically, it is very easy to make that the scalar will be returned as an object (at least fro GraphQL.NET). The question is whether it is a permissible expansion of behavior or completely unacceptable.

benjie:
The value that is at the tip of the GraphQL tree is a leaf, even if it's an object returned from a custom JSON scalar rather than just a `String`. The word "primitive" here generally implies that it's a simple type, but really it's just a "building block" type - GraphQL sees it as an "atom", indivisible.

That said; the wording here is potentialyl ambiguous because it says the leaves on the trees are GraphQL scalars, but they could also be enums, which in some parts of the spec are treated as within the scope of the term "scalars" and in some parts are not.

The concept of a JSON scalar has been discussed a few times at the GraphQL Spec WG and it's seen as allowable (though it's an "escape hatch" and should be avoided where possible). It has caused us quite a few issues (it's restricted how we can grow certain parts of the spec) but it's something that is in widespread use so we have to continue to support.

andimarek:
To add a bit on that:

the value returned by custom scalar for the result coercion (in GraphQL.js this function is called `serialize`) can by anything that the Programming language/environment allows. For example it can be a `Money` object. 

GraphQL in general doesn't not limit at all this aspect: it is an implementation detail of the GraphQL implementation. 

To phrase it a bit different: The result of an GraphQL execution is tree, where Scalars or Enums are the leaves from a type perspective. But what the actual values of these types are are runtime can't be predicted or described by the spec. 

Normally In practical terms a complex value returned by a Scalar result coercion probably needs to be serialized (warning: horrible naming, this a transport serialize which is a separate step from the result coercion) into a JSON representation. 

Have a look at https://www.graphql-scalars.com/scalars-guide and https://www.graphql.de/blog/scalars-in-depth/ for more details.

sungam3r:
@benjie @andimarek To summarize:
1. It is absolutely normal that scalars can eventually be represented by objects after serialization in JSON.
2. The specification explicitly does not speak about it.

Right? If so, doesn't you think that the specification needs to be supplemented in such a way as to eliminate the possibility of different interpretations?

[...]

PR #870, opened 2021-05-21
Editorial: Define type definitions

leebyron:
This broad change uses definition phrases to define each type, as well as expands the set of sub-sections at the top of the "Types" section which define *kinds* of types. In doing so applies the definitions arrived at in #845 for "wrapped types".

Makes editorial improvements throughout, removing duplicate language or attempting to improve correctness or legibility.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -263,74 +263,108 @@ TypeDefinition :
   - InputObjectTypeDefinition
 
 The fundamental unit of any GraphQL Schema is the type. There are six kinds
-of named type definitions in GraphQL, and two wrapping types.
+of *named type* and two kinds of *wrapped type* definitions in GraphQL.
 
-The most basic type is a `Scalar`. A scalar represents a primitive value, like
-a string or an integer. Oftentimes, the possible responses for a scalar field
-are enumerable. GraphQL offers an `Enum` type in those cases, where the type
-specifies the space of valid responses.
-
-Scalars and Enums form the leaves in response trees; the intermediate levels are
-`Object` types, which define a set of fields, where each field is another
-type in the system, allowing the definition of arbitrary type hierarchies.
-
-GraphQL supports two abstract types: interfaces and unions.
-
-An `Interface` defines a list of fields; `Object` types and other Interface
-types which implement this Interface are guaranteed to implement those fields.
-Whenever a field claims it will return an Interface type, it will return a
-valid implementing Object type during execution.
-
-A `Union` defines a list of possible types; similar to interfaces, whenever the
-type system claims a union will be returned, one of the possible types will be
-returned.
-
-Finally, oftentimes it is useful to provide complex structs as inputs to
-GraphQL field arguments or variables; the `Input Object` type allows the schema
-to define exactly what data is expected.
+:: A *named type* is one of the six types defined with a name: *Scalar type*,
+*Enum type*, *Object type*, *Interface type*, *Union type*,
+and *Input Object type*.
 
 
-### Wrapping Types
+### Leaf and Composite Types
+
+:: A *leaf type* is found at the leaves in a response tree. There are two kinds
+of leaf types: *Scalar type* and *Enum type*.
+
+The most basic *leaf type* is the *Scalar type*. A scalar represents a primitive
+value, like a string or an integer. GraphQL defines a set of *built-in scalars*
+and also allows a service to additionally provide any *custom scalar*.
+
+Oftentimes, the possible values for a *leaf type* are enumerable. GraphQL offers
+the *Enum type* in those cases, where the each enum type definition specifies
+the set of possible values.
+
+:: A *composite type* compose together leaf types and other composite types as a
+set of named fields, allowing the composition of arbitrary type hierarchies and
+nested trees of data. There are three composite *output type*: Object,
+Interface, and Union, and one composite *input type*: Input Object.
+
+The *Object type* is the concrete output *composite type* which describes an
+actual concrete value which will a field will resolve during execution, composed
+of other *output type* fields, allowing responses of nested data.
+
+Oftentimes it is useful to provide a *composite type* as an *input type* to a
+field argument or variable; the *Input Object type* composes a set of
+*input type* fields, allowing inputs of nested data.
+
+
+### Abstract Types
+
+:: An *abstract type* defines a set of possible types, one of which will be
+resolved during field execution. There are two kinds of abstract types:
+*Interface type* and *Union type*.
+
+:: The type an abstract type resolves to during execution is referred to as the
+*concrete Object type*, which is always an *Object type*.
+
+An *Interface type* defines a list of fields. An type which implements this
+Interface must also implement those fields. Whenever a field claims it will
+return an *Interface type*, it will return an implementing
+*concrete Object type* during field execution.
+
+A *Union type* defines a list of possible types. Similar to Interfaces,
+whenever a field claims it will return a *Union type*, it will return one of the
+possible types as the *concrete Object type* during field execution.
+
+
+### Wrapped Types
 
 All of the types so far are assumed to be both nullable and singular: e.g. a
-scalar string returns either null or a singular string.
+scalar string returns either {null} or a singular string.
 
-A GraphQL schema may describe that a field represents a list of another type;
-the `List` type is provided for this reason, and wraps another type.
+:: A *wrapped type* wraps another type (its *inner type*) to create a collection
+or variation of that type. There are two kinds of wrapped types: *List type* and
+*Non-Null type*.
 
-Similarly, the `Non-Null` type wraps another type, and denotes that the
-resulting value will never be {null} (and that a field error cannot result in a
-{null} value).
+:: The *inner type* is the type immediately wrapped by a *wrapped type*, which
+itself may be a *wrapped type* allowing for combinations of *wrapped type*. If
+continuously unwrapped, ultimately a *named type* is found.
 
-These two types are referred to as "wrapping types"; non-wrapping types are
-referred to as "named types". A wrapping type has an underlying named type,
-found by continually unwrapping the type until a named type is found.
+A GraphQL schema may describe that a field or argument expects a list of another
+type; the *List type* is provided for this purpose, and wraps an *inner type*
+describing the items in the list.
+
+Similarly, a field or argument may describe that it must never be {null}. The
+*Non-Null type* is provided for this purpose, and wraps an *inner type*
+describing the expected type, if not {null}.
 
 
 ### Input and Output Types
 
 Types are used throughout GraphQL to describe both the values accepted as input
 to arguments and variables as well as the values output by fields. These two
-uses categorize types as *input types* and *output types*. Some kinds of types,
-like Scalar and Enum types, can be used as both input types and output types;
-other kinds of types can only be used in one or the other. Input Object types can
-only be used as input types. Object, Interface, and Union types can only be used
-as output types. Lists and Non-Null types may be used as input types or output
-types depending on how the wrapped type may be used.
+uses categorize types as an *input type* or an *output type*. Some kinds of
+types, like Scalar and Enum types, can be used as both input types and output
+types; other kinds of types can only be used in one or the other.
+
+:: An *input type* is a *Scalar type*, *Enum type*, *Input Object type*, or a
+*wrapped type* of one of these types.
 
 IsInputType(type) :
-  * If {type} is a List type or Non-Null type:
-    * Let {unwrappedType} be the unwrapped type of {type}.
-    * Return IsInputType({unwrappedType})
+  * If {type} is a *wrapped type* (a *List type* or *Non-Null type*):
+    * Let {innerType} be the *inner type* of {type}.
+    * Return IsInputType({innerType})
   * If {type} is a Scalar, Enum, or Input Object type:
     * Return {true}
   * Return {false}
 
+:: An *output type* is a *Scalar type*, *Enum type*, *Object type*,
+*Interface Type*, *Union Type* or a *wrapped type* of one of these types.
+
 IsOutputType(type) :
-  * If {type} is a List type or Non-Null type:
-    * Let {unwrappedType} be the unwrapped type of {type}.
-    * Return IsOutputType({unwrappedType})
-  * If {type} is a Scalar, Object, Interface, Union, or Enum type:
+  * If {type} is a *wrapped type* (a *List type* or *Non-Null type*):
+    * Let {innerType} be the *inner type* of {type}.
+    * Return IsOutputType({innerType})
+  * If {type} is a Scalar, Enum, Object, Interface, or Union type:
     * Return {true}
   * Return {false}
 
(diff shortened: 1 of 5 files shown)


spec/Section 3 -- Type System.md
```diff
@@ -263,74 +263,108 @@ TypeDefinition :
...
+Interface must also implement those fields. Whenever a field claims it will
+return an *Interface type*, it will return an implementing
+*concrete Object type* during field execution.
+
+A *Union type* defines a list of possible types. Similar to Interfaces,
+whenever a field claims it will return a *Union type*, it will return one of the
+possible types as the *concrete Object type* during field execution.
+
+
+### Wrapped Types
```
leebyron, line 319: @benjie - would love your thoughts on this newly written subsection.

This is based on [your comment](https://github.com/graphql/graphql-spec/pull/845#issuecomment-826341088) with very minor changes:

* **Wrapped type** - the kind of type including Lists and Non-Nulls
  * **List type** - changed from your suggestion of **List wrapped type**
  * **Non-Null type** - changed from your suggestion of **Non-Null wrapped type**
* **Inner type** - the type immediately wrapped (eg the `ofType`)
  * **Item type** - I added this - the inner type of a List
  * **Nullable type** - I added this - the inner type of a Non-Null

spec/Section 3 -- Type System.md
```diff
@@ -263,74 +263,108 @@ TypeDefinition :
...
+:: A *named type* is one of the six types defined with a name: *Scalar type*,
+*Enum type*, *Object type*, *Interface type*, *Union type*,
+and *Input Object type*.
 
-Scalars and Enums form the leaves in response trees; the intermediate levels are
-`Object` types, which define a set of fields, where each field is another
-type in the system, allowing the definition of arbitrary type hierarchies.
 
-GraphQL supports two abstract types: interfaces and unions.
+### Leaf and Composite Types
```
sungam3r, line 273: Good time to clarify #826 .

leebyron: Thanks for making this connection! That earlier phrasing is definitely confusing. I'll weigh in on that issue and make sure this PR covers the solution. 

In writing this change I found a number of cases where we confuse "value" and "type" and interchange them even in the same paragraph. We've also made the mistake of using "scalar" when we mean "leaf".

benjie:
@leebyron I think some unrelated changes have made their way into this PR, e.g. `signed-agreements/README.md`

spec/Section 3 -- Type System.md
```diff
@@ -263,74 +263,108 @@ TypeDefinition :
...
 to arguments and variables as well as the values output by fields. These two
-uses categorize types as *input types* and *output types*. Some kinds of types,
-like Scalar and Enum types, can be used as both input types and output types;
-other kinds of types can only be used in one or the other. Input Object types can
-only be used as input types. Object, Interface, and Union types can only be used
-as output types. Lists and Non-Null types may be used as input types or output
-types depending on how the wrapped type may be used.
+uses categorize types as an *input type* or an *output type*. Some kinds of
+types, like Scalar and Enum types, can be used as both input types and output
+types; other kinds of types can only be used in one or the other.
```
benjie, line 347: *A little off-topic* Whilst we're on this subject... generally `scalar` types are suitable for input and output, but in some cases (for example the `Upload` scalar used by the GraphQL Upload Spec) a custom scalar may only be suitable for one but not the other. Is this specifically disallowed, or do we need to be more cautious with our wording?

spec/Section 3 -- Type System.md
```diff
@@ -354,35 +388,35 @@ represent additional fields a GraphQL client only accesses locally.
 
 ScalarTypeDefinition : Description? scalar Name Directives[Const]?
 
-Scalar types represent primitive leaf values in a GraphQL type system. GraphQL
-responses take the form of a hierarchical tree; the leaves of this tree are
-typically GraphQL Scalar types (but may also be Enum types or {null} values).
+:: A *Scalar type* represents primitive leaf values in a GraphQL type system.
+GraphQL responses take the form of a hierarchical tree; the leaves of this tree
+are typically GraphQL Scalar types (but may also be Enum types or {null} values).
```
benjie, line 393: When we're talking about `types` I'm not sure it makes sense to include `{null}`? This might imply you could have `type A {foo: null}`. We should also factor in that the leaves may be lists/non-nulls:

```suggestion
:: A *Scalar type* represents primitive leaf values in a GraphQL type system.
GraphQL responses take the form of a hierarchical tree; the *named type* of the
leaves of this tree are typically GraphQL Scalar types (but may also be Enum
types).
```

spec/Section 3 -- Type System.md
```diff
@@ -906,24 +938,23 @@ IsValidImplementation(type, implementedType):
...
+  4. If {fieldType} is an *Object type* and {implementedFieldType} is
+     a *Union type* and {fieldType} is a possible type of {implementedFieldType}
      then return {true}.
-  5. If {fieldType} is an Object or Interface type and {implementedFieldType}
-     is an Interface type and {fieldType} declares it implements
-     {implementedFieldType} then return {true}.
+  5. If {fieldType} is an *Object type* or *Interface type* and
+     {implementedFieldType} is an *Interface type* and {fieldType} declares it
+     implements {implementedFieldType} then return {true}.
   6. Otherwise return {false}.
```
benjie, line 958: Much clearer :+1:

spec/Section 3 -- Type System.md
```diff
@@ -1418,7 +1447,7 @@ EnumValuesDefinition : { EnumValueDefinition+ }
 
 EnumValueDefinition : Description? EnumValue Directives[Const]?
 
-GraphQL Enum types, like Scalar types, also represent leaf values in a GraphQL
+:: An *Enum type*, like a *Scalar type*, also represent leaf values in a GraphQL
```
benjie, line 1450:
```suggestion
:: An *Enum type*, like a *Scalar type*, also represents leaf values in a GraphQL
```

spec/Section 3 -- Type System.md
```diff
@@ -1492,9 +1521,9 @@ InputFieldsDefinition : { InputValueDefinition+ }
 Fields may accept arguments to configure their behavior. These inputs are often
 scalars or enums, but they sometimes need to represent more complex values.
 
-A GraphQL Input Object defines a set of input fields; the input fields are either
-scalars, enums, or other input objects. This allows arguments to accept
-arbitrarily complex structs.
+:: An *Input object type* defines a set of input fields; the input fields are
```
benjie, line 1524:
```suggestion
:: An *Input Object type* defines a set of input fields; the input fields are
```

benjie (changes requested):
I'll have to return to this, but the changes really improve the clarity of the text IMO - excellent work. I've only reviewed sections 2 and 3 so far.

benjie:
I have now completed the review but have no further feedback :+1:

spec/Section 3 -- Type System.md
```diff
@@ -263,74 +263,108 @@ TypeDefinition :
...
+and also allows a service to additionally provide any *custom scalar*.
 
-Finally, oftentimes it is useful to provide complex structs as inputs to
-GraphQL field arguments or variables; the `Input Object` type allows the schema
-to define exactly what data is expected.
+Oftentimes, the possible values for a *leaf type* are enumerable. GraphQL offers
+the *Enum type* in those cases, where the each enum type definition specifies
+the set of possible values.
 
+:: A *composite type* compose together leaf types and other composite types as a
```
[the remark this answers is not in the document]
leebyron: A leaf type is scalar/enum, not including wrapped. Same with composite types being the set described here, not including wrapped.

This is good feedback on this paragraph being falsely precise - since we can compose leaf or composite types (leaf + composite = named) and any wrapped of either of those (named + wrapped(named) = all types). Thats just all types.

Going to replace with something simple:

> A *composite type* composes other types together as a set of named fields

spec/Section 3 -- Type System.md
leebyron: We actually don't have the tools to assert this restriction right now. We don't specifically disallow it, but we also can't enforce it (with validation - a custom scalar could just fail coercion, but thats a bit of a hack IMO)

spec/Section 3 -- Type System.md
leebyron: Agreed. This was a case where we were mixing "types" and "values". I'll rewrite this

spec/Section 3 -- Type System.md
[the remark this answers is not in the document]
benjie: Sounds good to me 👍

spec/Section 3 -- Type System.md
benjie: Yep; I’m happy keeping the status quo; just felt it was worth mentioning.

mjmahone: > We actually don't have the tools to assert this restriction right now

I'd be interested in digging into this distinction for a later release: I have some usages around input vs output enums that I'd like to add clarity to. As the spec stands, [this statement](https://spec.graphql.org/draft/#sel-GAHfRDABABvEomB):
> GraphQL services must return one of the defined set of possible values. If a reasonable coercion is not possible they must raise a field error.

can mean it's actually a breaking change to add a new enum value to an existing enum  type (we make no statement about avoiding exhaustive switches on output enums, for example). We don't get into details in the spec itself on what a breaking type system change is, with just a [short blurb about type system evolution](https://spec.graphql.org/draft/#sec-Validation.Type-system-evolution).

yaacovCR: https://github.com/graphql/graphql-js/issues/968#issuecomment-317834310

yaacovCR: Relevant

spec/Section 3 -- Type System.md
```diff
@@ -1800,19 +1849,20 @@ a non-null input type as invalid.
...
 
-1. A Non-Null type must not wrap another Non-Null type.
+1. A *Non-Null type* must not wrap another *Non-Null type*.
 
 
 ### Combining List and Non-Null
 
-The List and Non-Null wrapping types can compose, representing more complex
-types. The rules for result coercion and input coercion of Lists and Non-Null
+The *List type* and *Non-Null types* can compose, representing more complex
```
benjie, line 1857: You've been careful to always use the singular elsewhere; I'm not sure if this relates to how the glossary feature in spec-md works. If this need not be singular on these technical grounds then plural is fine :+1: 

```suggestion
The *List type* and *Non-Null type* can compose, representing more complex
```

spec/Section 5 -- Validation.md
```diff
@@ -429,19 +429,19 @@ SameResponseShape(fieldA, fieldB):
...
-    * Let {typeA} be the item type of {typeA}
-    * Let {typeB} be the item type of {typeB}
+  * If {typeA} or {typeB} is a *Non-Null type*.
+    * If {typeA} or {typeB} is not a *Non-Null type*, return false.
+    * Let {typeA} be the *nullable type* of {typeA}
+    * Let {typeB} be the *nullable type* of {typeB}
+  * If {typeA} or {typeB} is a *List type*.
+    * If {typeA} or {typeB} is not a *List type*, return false.
+    * Let {typeA} be the *item type* of {typeA}
+    * Let {typeB} be the *item type* of {typeB}
```
benjie, line 439: Added periods.

```suggestion
    * Let {typeA} be the *nullable type* of {typeA}.
    * Let {typeB} be the *nullable type* of {typeB}.
  * If {typeA} or {typeB} is a *List type*.
    * If {typeA} or {typeB} is not a *List type*, return false.
    * Let {typeA} be the *item type* of {typeA}.
    * Let {typeB} be the *item type* of {typeB}.
```

spec/Section 5 -- Validation.md
```diff
@@ -1882,33 +1882,33 @@ IsVariableUsageAllowed(variableDefinition, variableUsage):
...
   * Return {AreTypesCompatible(variableType, locationType)}.
 
 AreTypesCompatible(variableType, locationType):
 
-  * If {locationType} is a non-null type:
-    * If {variableType} is NOT a non-null type, return {false}.
-    * Let {nullableLocationType} be the unwrapped nullable type of {locationType}.
-    * Let {nullableVariableType} be the unwrapped nullable type of {variableType}.
+  * If {locationType} is a *non-null type*:
+    * If {variableType} is NOT a *non-null type*, return {false}.
```
benjie, line 1899:
```suggestion
  * If {locationType} is a *Non-Null type*:
    * If {variableType} is NOT a *Non-Null type*, return {false}.
```

spec/Section 5 -- Validation.md
```diff
@@ -1882,33 +1882,33 @@ IsVariableUsageAllowed(variableDefinition, variableUsage):
...
+    * Let {nullableLocationType} be the *nullable type* of {locationType}.
+    * Let {nullableVariableType} be the *nullable type* of {variableType}.
     * Return {AreTypesCompatible(nullableVariableType, nullableLocationType)}.
-  * Otherwise, if {variableType} is a non-null type:
+  * Otherwise, if {variableType} is a *non-null type*:
```
benjie, line 1903:
```suggestion
  * Otherwise, if {variableType} is a *Non-Null type*:
```

spec/Section 5 -- Validation.md
```diff
@@ -1882,33 +1882,33 @@ IsVariableUsageAllowed(variableDefinition, variableUsage):
...
     * Let {nullableVariableType} be the nullable type of {variableType}.
     * Return {AreTypesCompatible(nullableVariableType, locationType)}.
-  * Otherwise, if {locationType} is a list type:
-    * If {variableType} is NOT a list type, return {false}.
-    * Let {itemLocationType} be the unwrapped item type of {locationType}.
-    * Let {itemVariableType} be the unwrapped item type of {variableType}.
+  * Otherwise, if {locationType} is a *list type*:
+    * If {variableType} is NOT a *list type*, return {false}.
```
benjie, line 1907:
```suggestion
  * Otherwise, if {locationType} is a *List type*:
    * If {variableType} is NOT a *List type*, return {false}.
```

spec/Section 5 -- Validation.md
```diff
@@ -1882,33 +1882,33 @@ IsVariableUsageAllowed(variableDefinition, variableUsage):
...
+    * Let {itemLocationType} be the *item type* of {locationType}.
+    * Let {itemVariableType} be the *item type* of {variableType}.
     * Return {AreTypesCompatible(itemVariableType, itemLocationType)}.
-  * Otherwise, if {variableType} is a list type, return {false}.
+  * Otherwise, if {variableType} is a *list type*, return {false}.
```
benjie, line 1911:
```suggestion
  * Otherwise, if {variableType} is a *List type*, return {false}.
```

benjie (changes requested):
Quite a lot of these are suggestions rather than strict change requests. I think this is significantly clearer than the previous spec text - and given it's taken me over an hour review it I can only imagine how long it took to write it - excellent work Lee!

spec/Section 3 -- Type System.md
```diff
@@ -263,74 +263,115 @@ TypeDefinition :
...
-returned.
+:: A *leaf type* is found at the leaves in a response tree. There are two kinds
+of leaf types: *Scalar type* and *Enum type*.
 
-Finally, oftentimes it is useful to provide complex structs as inputs to
-GraphQL field arguments or variables; the `Input Object` type allows the schema
-to define exactly what data is expected.
+The most common *leaf type* is the *Scalar type*. GraphQL defines a set of
+*built-in scalars* for representing common primitive values and also allows a
+service to additionally provide a *custom scalar type* to represent any value.
```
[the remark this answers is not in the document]
benjie: I'm closing this, we're consistently "custom scalar type" everywhere.

PR #873, opened 2021-05-27
Clarify query shorthand relationship with directives

benjie:
The spec text currently reads:

> If a document contains only one operation and that operation is a query which
> defines no variables and contains no directives then that operation may be
> represented in a short-hand form which omits the {`query`} keyword and operation
> name.

The word "contains" here is ambiguous, and it makes it sound like the following query is invalid since it "contains" a directive:

```graphql
{
  __typename @skip(if: true)
}
```

Really we mean that there is no direcive applied to the operation itself (directly); I've attempted to clarify this without making the text more verbose.


spec/Section 2 -- Language.md
```diff
@@ -308,9 +308,8 @@ mutation {
 **Query shorthand**
 
 If a document contains only one operation and that operation is a query which
-defines no variables and contains no directives then that operation may be
-represented in a short-hand form which omits the {`query`} keyword and operation
-name.
+has neither variables nor directives then that operation may be represented
+in a short-hand form which omits the {`query`} keyword and operation name.
```
IvanGoncharov, line 312: Can we clarify it even more, like `doesn't define variables or has directives applied to it`?

benjie: That's what I went with first but it became a bit wordy, hence my comment on "verbose". But yes, I'll give it a go :+1:

leebyron (approved):

leebyron:
I'll wait to merge until after the spec release cut

sungam3r (approved):

Issue #533, opened 2018-11-14
Proposal: Implicitly include properties of implemented interfaces


kaitlynbrown:
If an interface contains more than a few properties and is used in multiple places, you end up with a very significant amount of repetition that clutters up your schema. This means it requires a lot more scrolling to move around the schema, and the important details of a type (eg which properties are unique to it) are obscured by the noise added by all of the extra repetition.

As a contrived example, take the following currently legal schema, similar in structure to what I'm currently working on:
```
interface Vehicle {
  make: VehicleMake
  wheelCount: Int
  acceleration: Float
  topSpeed: Float
  mass: Float
  capacity: Int
}

type Bicycle implements Vehicle {
  make: BicycleMake
  wheelCount: Int
  acceleration: Float
  topSpeed: Float
  mass: Float
  capacity: Int
  gearCount: Int
}

interface Automobile implements Vehicle {
  make: AutoMake
  wheelCount: Int
  acceleration: Float
  topSpeed: Float
  mass: Float
  capacity: Int
  engineCylinders: Int
  engineSize: Float
}

type Car implements Automobile {
  make: CarMake
  wheelCount: Int
  acceleration: Float
  topSpeed: Float
  mass: Float
  capacity: Int
  engineCylinders: Int
  engineSize: Float
  trunkSize: Float
}

type Motorcycle implements Automobile {
  make: MotorcycleMake
  class: MotorcycleClass
  wheelCount: Int
  acceleration: Float
  topSpeed: Float
  mass: Float
  capacity: Int
  engineCylinders: Int
  engineSize: Float
}

type Truck implements Automobile {
  make: TruckMake
  wheelCount: Int
  acceleration: Float
  topSpeed: Float
  mass: Float
  capacity: Int
  engineCylinders: Int
  engineSize: Float
  maxLoad: Float
}

union AutoMake = CarMake | TruckMake | MotorcycleMake
union VehicleMake = AutoMake | BicycleMake
```
I propose that the following would be valid, and equivalent to the above:
```
interface Vehicle {
  make: VehicleMake
  wheelCount: Int
  acceleration: Float
  topSpeed: Float
  mass: Float
  capacity: Int
}

type Bicycle implements Vehicle {
  make: BicycleMake
  gearCount: Int
}

interface Automobile implements Vehicle {
  make: AutoMake
  engineCylinders: Int
  engineSize: Float
}

type Car implements Automobile {
  make: CarMake
  trunkSize: Float
}

[...]

Issue #500, opened 2018-08-15
[Proposal] Interface implementation violate "Don't Repeat Yourself"


yaquawa:
## Problem
I found it's very inefficient to implement interface in GraphQL.

```graphql
interface Character {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
}

type Human implements Character {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
  starships: [Starship]
  totalCredits: Int
}
```
In GraphQL, the word `implements` means "copy all the fields from interface to the target type".
The more types you get, you have to repeat your self more and more... If you add or delete a field in the interface, then you have to delete all the types that implement the interface, which is repeat yourself again. 

In GraphQL it's all about "copy and paste", no real implementation there. So why should we repeat ourself like this way?

## Proposal

Introducing a simple `extends` operator just like most of the program languages would be an excellent solution.

```graphql
type Character {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
}

type Human extends Character {
  starships: [Starship]
  totalCredits: Int
}
```

For consistency,`Interface` also can be extended using the `extends` operator.
```graphql
interface Character {
  id: ID!
  name: String!
  friends: [Character]
  appearsIn: [Episode]!
}

interface Human extends Character {
  starships: [Starship]
  totalCredits: Int
}
```

This approach won't introduce any breaking changes, and it's natural to have such an operator because it's not something new concept and fit with GraphQL very well too.

IvanGoncharov:
@yaquawa It's a duplication of #335.
If you think it differs in any significant way please describe this difference and I will reopen this issue.

yaquawa:
@IvanGoncharov 
Hi, thanks for the link.

#335 still need to COPY the `implementations` into the concrete `type`. 
What I want to emphasize is that there is no `Implementations` in GraphQL, both `interface` and `type` is just about typing something, so we don't need to implement a type, what we actually do for the Implementations of a type is just COPY&PASTE works.  (Eventually both `type` and `interface` is Interface)

IvanGoncharov:
@yaquawa There are many reasons why we need interfaces in GraphQL type system. 
But most importantly it's already part of the GraphQL spec (including SDL syntax) and we can't do any breaking changes.

Can you please rewrite your proposal so it doesn't introduce breaking changes to GraphQL?

yaquawa:
@IvanGoncharov
[...]

PR #881, opened 2021-08-09
[RFC] Inherit fields from interface automatically

tobiasdiez:
This is currently a Stage 0: Strawman proposal looking to advance to Stage 1.
The RFC is mainly based on the discussion in the related issues https://github.com/graphql/graphql-spec/issues/533 https://github.com/graphql/graphql-spec/issues/500 and https://github.com/graphql/graphql-js/issues/703.

If someone has additions or clarifications, please feel free to edit the added document directly (if you have the right) or open a PR against my fork. Many thanks in advance!

I hope it is okay that took some of the examples and formulations almost verbatim from the above issues, and expanded on them in form of a RFC. Also the contribution process is not completely clear to me, so please forgive if I overlooked important steps along the way.

### DIFF
--- a/rfcs/ImplicitInheritance.md
+++ b/rfcs/ImplicitInheritance.md
@@ -1,0 +1,236 @@
+# RFC: Implicit Inheritance
+
+
+Related issues:
+- [graphql-spec: Proposal: Implicitly include properties of implemented interfaces](https://github.com/graphql/graphql-spec/issues/533)
+- [graphql-spec: [Proposal] Interface implementation violate "Don't Repeat Yourself"](https://github.com/graphql/graphql-spec/issues/500)
+- [graphql-js: Thought: avoiding self-repitition when a schema is defined as a string](https://github.com/graphql/graphql-js/issues/703)
+
+## Problem Statement
+
+Currently, when defining an interface that implements another interface, the implementing interface must define each field that is specified by the implemented interface.  
+If an interface contains more than a few properties and is used in multiple places, one ends up with a very significant amount of repetition of the same field definitions.
+This becomes particularly apparent if you have deeper hierarchies of interfaces since fields are repeated on each level.
+The following schema exhibits this exemplarily.
+
+```graphql
+interface Vehicle {
+  id: ID!
+  make: VehicleMake
+  wheelCount: Int
+  acceleration: Float
+  topSpeed: Float
+  mass: Float
+}
+
+interface Automobile implements Vehicle {
+  id: ID!
+  make: AutoMake
+  wheelCount: Int
+  acceleration: Float
+  topSpeed: Float
+  mass: Float
+  engineCylinders: Int
+  engineSize: Float
+}
+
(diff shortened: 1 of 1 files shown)


leebyron:
This has been discussed before and rejected. See one comment explaining rationale here https://github.com/graphql/graphql-spec/issues/500#issuecomment-426146328

However I appreciate the work in collecting all details about this RFC in one file. I'm about to move all RFCs to another repository so merging this in advance of doing that

tobiasdiez:
Thanks @leebyron for your reaction. I understand that there are certain objections against automatically inheritance (as for example outlined in your linked comment). But as you see in the reactions to your comment and in #533 (which is under the top 15 of issues with +1), not everyone agrees with the point of favoring readability over not-repeat-yourself. In my opinion, this comes also from the fact that IDE support for graphql caught-up and jumping around the schema is nicely supported right now; in my opinion there is also a shift in using the graphql schema more for writing and using other tools for exploring and reading (like graphql playground, automatically generated docs etc). Due to these natural difference between philosophies, I phrased this RFC as an optional extension - so if a team decides it's more suitable for them to keep the current behavior than that's still possible.

For these reasons I hope the decision to reject this proposal can be revisited. If not, I would suggest to close #533 as well.

leebyron:
You're absolutely welcome to join the next working group meeting. Proposals typically only move forward when their champions bring them to WG meetings to work through the details.

rfcs/ImplicitInheritance.md
```diff
@@ -0,0 +1,236 @@
...
+## Disadvantages
+- It is no longer possible to immediately see all fields that are provided by a type (without traversing the inheritance hierarchy), which makes reading the schema more challenging in some aspects.
+- If a type inherits multiple interfaces that declare a field with different but compatible types, the resulting type is not immediately clear.  
+
+It should be noted that these disadvantages can mostly be mitigated using tools or IDE integrations that can easily determine all fields on the fly.
+
+## Evaluation against [guiding principles](https://github.com/graphql/graphql-spec/blob/main/CONTRIBUTING.md#guiding-principles)
+- Backwards compatibility: ✅
+
+  Existing queries and schema are still valid.
```
sungam3r, line 192: Yes, but some GraphQL frameworks/parsers most likely adhere to current "hard" rules about inheriting fields from interfaces so there may be problems when trying to use them with the "new lightweight" schema.

meostyles:
@leebyron This is **absolutely** the worst part about GraphQL, as someone who works on a large type system. Copying fields to many classes is incredibly frustrating:
- you lose any comments on those fields in the interface, so you'd better not forget to copy the comments to each concrete class too!
- developers hack around the limitation with composition, where a class that would otherwise implement an interface and decorate with additional fields instead holds a reference to a concrete type with fields that are shared. This is unfortunate, as it adds an extra layer in the schema since you're introducing a new field, but not entirely unreasonable given the pain of interfaces

This problem was solved decades ago in programming languages. Why is GraphQL stuck in the past?

Issue #101, opened 2015-10-03
Map type


miracle2k:
Should there be a map type?

I have a structure that looks like this:

      Item {
         title {
            de_DE,
            fr,
            pa,
            .....
         }
      }

The point here is that the number of language strings might be arbitrary. I find this hard to model using GraphQLObject, because:

1. The number of locales is undefined. While the resolver could handle this dynamically, I think it's a problem in Relay that the schema cannot reflect these dynamic fields.
2. I would like to query all the existing locales, which isn't allowed, it seems ("must have a sub-selection").

If I make my GraphQL server return a JSON object for "title", the Relay client doesn't complain (although  maybe shouldComponentUpdate breaks), but I think I'm skirting by here. At the very least, I think I couldn't generate a schema that confirms to the spec.

OlegIlyenko:
I think this is a valid concern as well. I currently thinking of ways to add GraphQL endpoint for our API. We are building project-based multi-tenant service. For example every project has a list of products which have name and description. Just like in your case these are localized strings as well. 

One way one can approach this problem is to parametrize the field:

```js
query ProductNames {
  products {
    nameEn: name(locale: "en")
    nameDe: name(locale: "de_DE")
  }
}
```

Another approach, as you mentioned, would be to generate a schema of-the-fly. In our case it would be possible, since every project has a limited set of locales which are defined in the config of this project. 

Where it becomes more tricky is an addition user-defined data types. Users of our API can define new attributes for products (visually through the merchant-center application). These attributes are also typed, so it's possible to generate GraphQL schema for this project, but it has some implications:

* Generating schema on every request comes with big performance penalty, since we need to load config from a DB in order to do so
* to compensate for this performance penalty we need to implement some schema caching and cache invalidation logic
* Schema now becomes tenant-specific. This means that it becomes much  harder to integrate with generic tools like GraphiQL and do introspection in general
* You can't even do an introspection queries without the auth anymore.

I guess one can just put all custom attribute JSON in a string scalar, but I don't think that other developers will appreciate JSON inside of string inside of another JSON :) I feel that generic JSON/Map-like type can provide a very useful middle-ground for these use-cases. It can also help a lot with a migration to the GraphQL. One can quickly start with Map-like structure and after some time develop schema generation and caching mechanisms.

Sandreu:
Hello,

I agree about this, and as @OlegIlyenko said, JSON string inside JSON string seems awkward.
[...]

Issue #867, opened 2021-05-08
[RFC] Client controlled nullability operator


twof:
[You can read a fuller version of the RFC here.](https://github.com/twof/graphql-spec/pull/1)

# TL;DR:
This RFC proposes new syntax for GraphQL clients to express that fields in an operation are **non-null**.

## Problem Statement
The [official best practice](https://graphql.org/learn/best-practices/#nullability) for schema design is to make all fields nullable by default. If “null” is not an appropriate value for a failed field, the guidance recommends making it nonNull in the SDL.

The problem with the SDL nonNull (`!`) is that it eliminates the possibility of partial failure on a given type. This forces the schema author to decide for which fields partial failure is acceptable. A GraphQL schema author may not be in the best position to decide whether partial failure is acceptable for every possible use case.

Additionally, the answer to whether a missing value is acceptable may vary between UI canvases. Specifying nonNull (aka. required) in the query gives the UI developer the power to decide where partial failure is acceptable, and the flexibility for different canvases to have different requirements in this regard.

## 🧑‍💻 Proposed Solution

Three new operators are introduced to the query language:
1. A "required designator" represented by `!` which marks a field Non-Nullable
2. An "optional designator" represented by `?` which marks a field Nullable
3. A "list operator" represented by `[]` which allows the above two operators to be applied to the elements of a list, or the list as a whole.

## 🎬 Behavior

The proposed query-side Non-Null designator and SDL-defined [Non-Null](https://spec.graphql.org/June2018/#sec-Errors-and-Non-Nullability) would have slightly different behavior. When null is resolved for a SDL Non-Nullable field, the executor propagates until it finds the nearest Nullable parent, and sets that to null. 

This proposal also introduces a second "optional designator", `?`,  which designates that the field is Nullable, even if it is Non-Nullable in the schema. When null is resolved for a field marked with a required designator (`!`) the executor instead propagates null to the nearest field marked with an optional designator (`?`). If there is no optional designator to propagate to, then the response's `data` field will become null ie the entire response will be lost. Because of that, in most cases we expect that required designators will be paired with optional designators to preserve partial responses.

## ✏️ Proposed syntax - `!`
The client can express that a schema field is required by using the `!` syntax in the query definition:
```graphql
query GetAllBusinesses {
  businesses [!]? {
    name!
  }
}
```
[...]

PR #825, opened 2021-02-19
RFC: Oneof Input Objects, Oneof Fields

benjie:
First came ~~[the @oneField directive](https://github.com/graphql/graphql-spec/pull/586)~~.

Then there was ~~[the Tagged type](https://github.com/graphql/graphql-spec/pull/733)~~.

Introducing: Oneof Input Objects and Oneof Fields.

Oneof Input Objects are a special variant of Input Objects where the type system asserts that exactly one of the fields must be set and non-null, all others being omitted. This is represented in introspection with the `__Type.oneField: Boolean` field, and in SDL via the `@oneOf` directive on the input object.

Oneof Fields are a special variant of Object Type fields where the type system asserts that exactly one of the field's arguments must be set and non-null, all others being omitted. This is represented in introspection with the `__Field.oneArgument: Boolean!` field, and in SDL via the `@oneOf` directive on the field.

_(Why a directive? See the FAQ below.)_

These two variants introduce a form of input polymorphism to GraphQL. For example, the following `PetInput` input object lets you choose between a number of potential input types:

```graphql
input PetInput @oneOf {
  cat: CatInput
  dog: DogInput
  fish: FishInput
}

input CatInput { name: String!, numberOfLives: Int }
input DogInput { name: String!, wagsTail: Boolean }
input FishInput { name: String!, bodyLengthInMm: Int }

type Mutation {
  addPet(pet: PetInput!): Pet
}
```

Previously you may have had a situation where you had multiple ways to locate a user:

```graphql
type Query {
  user(id: ID!): User
  userByEmail(email: String!): User
  userByUsername(username: String!): User
  userByRegistrationNumber(registrationNumber: Int!): User
}
```

with Oneof Fields you can now express this via a single field without loss of type safety:

```graphql
type Query {
  user(
    id: ID
    email: String
    username: String
    registrationNumber: Int
  ): User @oneOf
}
```

### Why is this a directive?

It's not. Well, not really - its an internal property of the type/field that's exposed through introspection - much in the same way that deprecation is. It just happens to be that after [I analysed a number of potential syntaxes](https://gist.github.com/benjie/5e7324c64f42dd818b9c3ac2a91b6b12) (including keywords and alternative syntax) I've found that the directive approach is the least invasive (all currently GraphQL parsers can already parse it!) and none of the alternative syntaxes sufficiently justified the increased complexity they would introduce.

### Why is this a good approach?

[...]

PR #888, opened 2021-09-29
RFC: Map type

nojvek:
This is an RFC for a new "Map" type to be added to GraphQL.

I acknowledge issue #101, that has 79 comments and 150+ 👍 votes. @leebyron locked the issue with the comment 
> If someone feels strongly that this concept deserves first-class support in GraphQL, I suggest following the RFC procedure to take this from a general suggestion to an actual proposal.` 

This is that proposal.

## Problem statement

This proposal aims to keep in mind "The Guiding Principles" laid out in the CONTRIBUTING.md.

Currently, GraphQL doesn't offer a way to return a Map/Dictionary response. 

A workaround is to return a key/value pair response as suggested in https://stackoverflow.com/questions/56705157/best-way-to-define-a-map-object-in-graphql-schema

```gql
type ArticleMapTuple {
     key: String!
     value: Article!
}

type Article {
  name: String!
}
```

response
```
[
  {
    "key": "foo1",
    "value": {name: "Foo1"}
  },
  {
    "key": "foo2",
    "value": {name: "Foo2"}
  },
  {
    "key": "foo3",
    "value": {name: "Foo3"}
  },
]
```

The problem is searching for the key "foo3" in the list requires traversing through the list. The alternative is to process the response into a local object via `Object.fromEntries` and then use it for fast lookups. 

Maps/Dictionaries are core data types in most languages. The json spec  supports objects with key: value pairs. By having support for Maps, GraphQL clients can make effient key:value lookups on responses.

This proposal introduces `field: { Type }` syntax to specify Maps. Similar to existing `field: [ List ]` syntax. 

The primary motivation in this proposal is the idea that **Maps are Lists with ID! (non-null string) keys, and should behave similar to Lists.**

Most relational databases have tables with schemas in the format:

```
type SomeEntity {
  id: ID!
  field1: String!
  field2: Int!
}
```

Having the response with IDs as keys gives GraphQL consumers/clients the ability for O(1) map lookups instead of O(n) list lookups. 

```
{
 "idAbc": {field1: "foo", field2: 123}
}
```

The other argument is that in many instances, GraphQL sits on top of an existing REST-ful api which returns responses with map responses. A real-world example is algolia.

Algolia indexes map fields for very fast facet lookups. e.g.

```js
{
 id: "123"
 name: "K95 Face Shield 24 PK",
 stockByLocation: {
   "seattle": 30,
   "portland": 40,
   "miami": 30,
   "st_louis": 10,
   ...
 }
}
```

To implement a GraphQL api over algolia, it would require changing the shape of `stockByLocation` response. By having GraphQL as schema enforcer, Map type would open a lot more possibilities of GraphQL adoption. 

The schema for above response would be:

```gql

type InventoryItem {
  id: ID!
  name: String!
  stockByLocations: { Int! }!
}
```

### List type

Currently, the List type is the only unbounded type in GraphQL.

SDL
```gql
type Query {
  users: [User!]!
}

type User {
  id: ID!
  firstName: String!
  lastName: String!
}
```

query:
```gql
{
  users {
    id
    firstName
    lastName
  }
}
```

response:
```json
{
  "users": [
    {"id": "foo", "firstName": "Foo", "lastName": "Bar"},
    {"id": "hello", "firstName": "Hello", "lastName": "World"}
  ]
}
```

Notice how the query didn't specify `[]` to specify a list response. Based on the type declaration `users: [User!]!`, only the fields of the List's value type are specified.

```
{
  users [{
    id
    firstName
    lastName
  }]
}
```

^ NOTE: this is an invalid gql query. 

The response can return any number of items in the list. GraphQL doesn't control what will be returned at the 0th index of the list, or the 1st index. This is upto the GraphQL service to determine.

A list can be seen a map with incremental numeric keys. It supports fast lookups at an index.

```
[
    0: {"id": "foo", "firstName": "Foo", "lastName": "Bar"},
    1: {"id": "hello", "firstName": "Hello", "lastName": "World"},
    2: {"id": "jsmith", "firstName": "John", "lastName": "Smith"}
]
```

### Map type

Following the principle of "Maps are Lists with string keys, and should behave simiar to Lists."

Note: The value type will still need to be explicitly specified. This is not an escape hatch for `Any` type. 

SDL
```gql
type Query {
  users: {User!}!
}

type User {
  id: ID!
  firstName: String!
  lastName: String!
}
```

query:
```gql
{
  users {
    firstName
    lastName
  }
}
```

response:
```json
{
  "users": {
    "foo": {"firstName": "Foo", "lastName": "Bar"},
    "hello": {"firstName": "Hello", "lastName": "World"}
  }
}
```

### Q: Why non-null string keys only?

A: Because grapqhl responses are json, and json only supports string keys. 

Alternative syntax is `field: {ID!: Type}`, however that would indicate that GraphQL may support other key types like Ints. I'd love for users to fall into the pit of success, so feel the semantics should be simple. Only string key types. Less is more. 

`field: { Type }` for Maps, `field: [ Type ]`  for Lists. The non-null versions being. `field: { Type! }!` and  `field: [ Type! ]!`.

--

### Q: What about nested Maps?

A: Nested lists work e.g. `field: [[ Type ]]`, therefore, nested maps should also work in a similar fashion i.e. `field: {{ Type }}`. The difference is that there would be no automatic coercion. If the shape of response doesn't match then there is a type error.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -1039,7 +1039,6 @@ ListValue : [ Value+ ]
     * Append {value} to {inputList}.
   * Return {inputList}
 
-
 ### Input Object Values
 
 ObjectValue[Const] :
@@ -1144,15 +1143,19 @@ that variable, that operation is invalid (see [All Variable Uses Defined](#sec-A
 Type :
   - NamedType
   - ListType
+  - MapType
   - NonNullType
 
 NamedType : Name
 
 ListType : [ Type ]
 
+MapType : { Type }
+
 NonNullType :
   - NamedType !
   - ListType !
+  - MapType !
 
 GraphQL describes the types of data expected by arguments and variables.
 Input types may be lists of another input type, or a non-null variant of any
@@ -1171,6 +1174,12 @@ Type : [ Type ]
 
   * Let {itemType} be the result of evaluating {Type}
   * Let {type} be a List type where {itemType} is the contained type.
+  * Return {type}
+
+Type : { Type }
+
+  * Let {itemType} be the result of evaluating {Type}
+  * Let {type} be a Map type where {itemType} is the value corresponding to a non-null string key.
   * Return {type}
 
 Type : Type !


nojvek:
Just filled in CLA.

dotansimha:
This is interesting! 
@nojvek in this example, the `key` is always a `string`, right? how will it effect resolvers and execution?

nodkz:
I like this approach. 
Pitty that it breaks backward compatibility with old servers and tools. 

BUT maybe this RFC opens a way to bring backward incompatible features and helps somehow to speed up & simplify InputUnions implementation.

nojvek:
@dotansimha - The proposal recommends only supporting string keys. Reason being json objects are always string keys. Even in JavaScript doing a `x = {}; x[1] = 1;` will result in “1” as string keys. If other types are desired, then user should fall back to a list with key, value fields on a custom type. The json response from a graphql api shouldn’t contain any special semantics to handle maps, they should return as plain json objects with key value pairs. 

@nodkz why would this be backwards incompatible? IIUC this is an additive change, hence all existing schemas should work as is. 

As for resolving, I plan to make a PR on graphql.js to get a working implementation with examples. The core idea is that map field resolvers would work the same way as list field resolvers. The only exception being that no automatic coercion would take place. 

Also there would be no per-key resolvers. The field ought to return a map with key value pairs. Like lists, the sub fields of the values can be recursively resolved.

nodkz:
@nojvek `category: { Category }` will break the parser which affects on already running eslint-plugin-graphql, graphql-eslint,  graphql-tools, graphiql, playground, graphql-inspector, graphql-compose and bunch of other tools
 
Screen Shot 2021-10-01 at 00 15 31

**So yep, existing schemas would work, but existing tooling with new syntax - not.**

It's bad. BUT it's not a big problem. If will be introduced protocol/spec version to GraphQL. For example via addition `graphql: 2.0` as the first line of SDL or JSON introspection like it did in [OpenAPI/swager](https://petstore.swagger.io/v2/swagger.json), [HTTP 1,2,3](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Evolution_of_HTTP), [grpc syntax=proto3](https://developers.google.com/protocol-buffers/docs/proto3) and other protocols.

nojvek:
In terms of semantic versioning that isn't considered a breaking change right? It's in the same realm as introducing directives syntax, or any new syntax for that matter. It's an additive change. 

Breaking change is when an upgrade would break existing graphql schemas and documents because a feature got removed. 

I do understand your point. I  do feel pretty motivated by map types, that I'm willing to put the work to make PRs on graphql-js, the spec, language server, Vscode-graphql. 

I understand it's not a small amount of work, therefore I would love to keep the scope limited. 

--

I take it that this is a strawman level RFC. How does one find a champion to take this to the next level?

rivantsov:
I would strongly object this approach - map is a list with string keys. 
So for a map all values are of the same type?! this is so limiting that it makes the whole map thing practically useless. Considering that it involves serious syntax extension, I do not think it's worth the effort. 

Yes, we need a map (dictionary) type. Not everything is strongly typed. There are cases when a column is a container for free-form data that does not fit any specific type. I work with logs; Imagine a log record - there's Time, Source, Message, Severity - strongly typed. But there's also need to save extra info, depends on the system/situation So the solution is Data column that contains key-value pairs. This situation is very common out there, and GraphQL needs to support a 'type' for this 'whatever' data - which would be Map
The keys are strings (like prop names in an object), but the values are of any type, and not necessarily the same for all keys. That MAP would be really useful.

As far as I know, some GraphQL implementations added Map scalar. And I just did it too in NGraphQL. 
The problem I see is - it would be logical to allow selection subset for the Map field, to select specific keys if needed; especially if there are many keys. But since Map is implemented as custom Scalar,  this I guess is not allowed. I believe we need this part to figure out - yes or no on selection subset. 

To sum it up my suggestion: add Map type as dict with values of any type; no new syntax, clarify if subset is allowed or not.

benjie:
I think one of the big potential issues with this is it breaks GraphQL's query/response symmetry:

For the query:

```graphql
{ users { firstName } }
```

With lists:

```json
{ "users": [ { "firstName": "Alice" }, { "firstName": "Bob" } ] }
```

With the proposed map:

```json
{ "users": { "KLJSDF239sdfkjh9q": { "firstName": "Alice" }, "llw0gis80q3jisd": { "firstName": "Bob" } } }
```

You may think of lists as having "index keys" but the critical thing here is that those keys are implicit - they are unstated - so they do not add additional data to the response body.

---

> So for a map all values are of the same type?! this is so limiting that it makes the whole map thing practically useless.

I strongly disagree with this sentiment; it's very common to have `{ [key: string]: User }` or similar map types in applications. And given that the value could be a union your options are even more flexible `{ [key: string]: User | Org | Post | Forum | ... }`. If we were to add [oneof](#825) to output types then you'd even be able to add scalars and lists to the mix - all without losing type safety.

---

> How does one find a champion to take this to the next level?

_Be_ that champion! Add yourself to the next GraphQL WG: https://github.com/graphql/graphql-wg/blob/main/agendas/2021-10-07.md

---

> `category: { Category }` will break the parser which affects on already running eslint-plugin-graphql, graphql-eslint, graphql-tools, graphiql, playground, graphql-inspector, graphql-compose and bunch of other tools

Tools that operate based on introspection (and correctly ignore types they do not understand) should not be broken by this additive change. GraphQL only has non-breaking guarantees with respect to existing operation documents remaining valid (including the introspection query), if you're using SDL then it's anticipated that your tool will not support future SDL versions that introduce new syntaxes and will need to be updated.

I'm not sure that GraphiQL/Playground will be broken by this since they use introspection. If they are, they should be modified to handle/ignore the newer types they do not understand in the introspection response.

nojvek:
> Be that champion! Add yourself to the next GraphQL WG:

Awesome. I'll make a PR and add to agenda. 

As for maps being very flexible and allowing anything, that would go against how request/response format works as @benjie pointed out. IMHO Graphql having strict value types is a feature. I'm not sure how resolution would even work for `any` type. Union types are great. They would work with existing proposal.

rivantsov:
Yes, GraphQL is about strong types, so far. The problem is that the world is not all strongly typed. There are numerous cases when the type is not known in advance, and Dict<string, object> is the best solution. NoSql world is full of stuff like that, and they are quite loose on types. 
Don't get me wrong, when it comes to strong typing - I'm the biggest advocate. My dev language is c#, for g sake! But there are a lot of legit cases when you need an 'object' field among other strongly typed stuff. 
Benji - 'it is very common to have...' - yes, it a world of 30m devs and millions of projects you can have thousands of cases of almost anything. But my problem with suggested map is that it will occupy a conceptual space (map) that belongs to a broader concept (Dict<string, object), that sooner or later we will have to bring in I think, and then it will be hell of a confusion with 2 of these things in the standard. 
I think Dict<string, object> covers the cases described here, without any syntax extensions, and it actually is way more useful.

IvanGoncharov:
I think this proposal addresses the real issue, I just disagree that we need to add Map type for that.
I see a few problems with the `Map` type:
1. We can't guarantee the order of fields in maps, so we drop support for ordered fields in #191
   So we can't guarantee the order of fields in JSON for Map type. That means you can add sorting criteria only to fields returning `List`.
2. Any `Map` can be represented as `List` with key as one of the fields that mean we two ways to represent collections and switching between them results in a breaking change.

Bellow is my strawman proposal, for query-side solution:
I think `List` is perfect for exposing data (since it maintains order) and we just need to allow the client to control the shape inside the query:
```
{
  users {
    [id]: {
      firstName
      lastName
    }
  }
}
```
With following limitations:
1. Key can only be of type `String!` or `ID!`. But you can do `[id!]` per #867 if `id` is nullable.
2. This transformation is valid only on `List` fields and applied only to the most inner list.
   Meaning if you have a field returning `[[User]!]!` you will get `[Map<User>!]!` after transformation.
3. If two items have the same keys it results in a runtime error

That way schema designer always exposes collections as `List` and it's up to a client to decide if cares about ordering or not.

nojvek:
@IvanGoncharov thats an interesting idea. Although not a fan that it only works at the first level but not at the second level. 

For users to fall in pit of success I believe, they shouldn't have to learn about the special edge cases. I.e map of maps of maps work just like list of lists of lists are supported by GraphQL. `Maps are string indices, as Lists are incremental numeric indices`

I'm not sure I understand the key ordering argument. The key ordering shouldn't matter right? List like when you query for a list, it's upto the server to decide what to return at 1st index and thereon. I could be missing something.

IvanGoncharov:
> Although not a fan that it only works at the first level but not at the second level.

@nojvek If you are worried about that we can complicate the syntax a bit.
For example, if we have:
```
type Query {
  chessBoard: [[CheessSquare!]!]!
}

type CheessSquare {
  coordinates: String! # E.g. E2
  piece: ChessPieceEnum
}
```

If you do:
```
{
  chessBoard: {
    [coordinate]: {
      piece
    }
  }
}
```
You will get `Map<string, CheessSquare>`.
But if you add square bracess like so:
```
{
  chessBoard: [{
    [coordinate]: {
      piece
    }
  }]
}
```
It means you leave the first level of the array as-is and converting its items to maps so you get `Array<Map<string, CheessSquare>>`.
If you add one more set of brackets you will get `Array<Array<Map<string, ChessSquare>>>`.

rivantsov:
@nojvek , this is not Relay spec, there's no ID here. 
Implicit ordering - although ordering matters for clients in most apps, returning in ANY order if order is not specified is OK. For example, SQL (in any db server) is free to return recs in any order if OrderBy is not specified. Dictionary entries or properties in objects in languages like c# are NOT ordered, rules explicitly state that there is no order, and it works OK all the time. 
If you try to force some order all the time on returned set, you put unnecessary burden on server and client, forcing to sort entries all the time, even when it is a small set and it does not matter for client or server. Much better to follow SQL pattern - server is free to return in any order when order is not specified.
@IvanGoncharov, I like your proposal, at least looks better to me than the original. But then what we get is an option to slightly reformat the returned list (group it by some ID). Is this such a big deal for clients, making life so much easier so that it's worth this syntax extension?

nojvek:
After sleeping over this, I think there's still a case for the original proposal with `field: {Value}` semantic.

In the list to dict transformation proposal by @IvanGoncharov, there is no way to specify a return type of dict with scalar keys. We can't specify `stockByLocation` as `Map<ID!, Int!>`.

```
{
 id: "123"
 name: "K95 Face Shield 24 PK",
 stockByLocation: {
   "seattle": 30,
   "portland": 40,
   "miami": 30,
   "st_louis": 10,
   ...
 }
}
```

They are not mutually exclusive proposals. One is syntax sugar to transform list into dict, while the other is allowing resolution of map fields which return map. The map of map of scalars is a valid usecase supported by `field: {{Bool!}}`

which can return 

```
{
  field: {
    foo: {
      bar: true,
    }
    baz: {
      qaz: false
    }
  }
}
```

The usecase for returning map types is when using document databases like mongo or KV stores like redis, there are many instances where the document is stored as json containing maps. There's also graphql being a layer on top of existing REST-ful apis that return values in similar shape. E.g an api returning feature flags.

```
{
  showPage1: true,
  showNewNav: true,
  showNewDesign: false
}
```

IvanGoncharov:
Another interesting by-product of this proposal is that we don't need to "single entity" field + alias combination anymore.
For example, with GitHub API you may need to get a few selected repositories.
With the current API you can only do:
```graphql
{
  organization(login: "graphql") {
    GraphQLSpec: repository(name: "graphql-spec") { ...SomeInfo }
    GraphQLJS: repository(name: "graphql-js") { ...SomeInfo }
  }
}
```

Aliases are non-dynamic (need to be explicitly written in query) and can't have arbitrary names.
With the proposed map transformation you can add `repositories` filed instead and do:
```
{
  organization(login: "graphql") {
    repositories(names: ["graphql-spec" "graphql-js"]) {
      [name]: {
        ...SomeInfo
      }
    }
  }
}
```
And you will get a nice Map with names as keys.


Another use case it allows query dynamic data more naturally, for example we can have `JSON` type as follows:
```
type JSON { # making it oneof per #825 will make it even better
  number: Float!
  boolean: Boolean!
  string: String!
  array: [JSON]!
  object(keys: [ID!]): [JSONProperty!]!
}

type JSONProperty {
  key: ID!
  value: JSON
}
```

That allows you to query JSON and get result with very minimal wrapping:
```
{
  jsonField {
    object(keys: ["foo", "bar"]) {
      [key]: value {
        number
        string
      }
    }
  }
}
```
With oneof sematic from #825 you will get:
```
{
  jsonField: {
    foo: { number: 3.14 }
    bar: { string: "baz" }
  }
}
```


@nojvek But I don't hijack this PR so I'm ok creating separate one if you think both proposals are worth exploring. What do you think?

mjmahone:
IMO there is a real need for user-created generic types, a la 
```
abstract type KeyPair<K, V> {
  key: K
  value: V 
} 
```
which you could use in a schema type via:
```
type User {
  named_friends: [KeyPair<ID, User>]
}
```
which you could use at execution time like:
```
query Q {
  me {
    named_friends {
      key
      value {
        name
      }
    }
  }
}
```
Note we already have generic types (lists and nullable types), we just don't allow user-creation of generic types.

This would also allow different schemas to expose different *kinds* of maps depending on what's needed:
```
abstract type MapViaTwoLists<K, V> {
  keys: [K]
  values: [V]
}

abstract type MapViaPairs<K, V> {
  values: [Pair<K, V>]
}

abstract type TreeMapNode<K, V> {
  key: K
  value: V
  left: TreeMapNode<K, V>
  right: TreeMapNode<K, V>
}
abstract type FullTreeMap<K, V> {
  nodes: [TreeMapNode<K, V>]
}
```

I believe this is an alternative to adding a single map type: it's unclear what is actually wanted for a map type. Do you want to create some tree with O(logn) lookup? A bijection between two groups of data? Something else?

leebyron:
Feedback from Oct WG for posterity:

---

We should better define what problem we want to solve, which will guide us to the best path to invest proposals into. For example:

If the problem is that we have "bags of data" on the server, such that the keys and values are generally unknown - then the past discussions around the JSON scalar might be most compelling.

If the problem is true Map types on the server that need to be translated to true Map types on the client, then a full featured Map type (or pattern) will be the most compelling.

If the problem is wanting to utilize JSON objects as Maps in the query response, then a transformation syntax will be the most compelling.

It could be that multiple of these are worth exploring to a degree, but better understanding what problem we're solving will help us focus on making the right tradeoffs here.

---

In terms of the transformation syntax. I personally think it is super interesting. It feels inline with other related ideas of aliased fragments and nested objects as well as field chaining. I almost want to say we should hold back on something like this until we could properly explore the use cases and surface area of all kinds of transformation to better understand the total design space.

That said, I want to stay cautious about my own interest because if I apply our [guiding principles](https://github.com/graphql/graphql-spec/blob/main/CONTRIBUTING.md#guiding-principles) then I think this path starts to seem misaligned.

Most importantly, **"Enable new capabilities motivated by real use cases"** seems hard to argue since this is purely a transformation of something we could already do. We have to ask ourselves if having the query and service do this transform rather than a client-side post-process is worth the complexity. Which comes to **"Simplicity and consistency over expressiveness and terseness"**. The syntactic complexity, the restrictions on keys, the concerns about lost-data footguns... these all seem like they're big hits to simplicity and consistency all for some fairly minor expressiveness wins. And then **"Favor no change"** means we need to set a high bar of added value before making significant changes. I think a broader transformation proposal could potentially leap this hurdle, but right now I'm not so sure.

---

In terms of the Map type, looking to prior art I think our clearest source of inspiration could be [protobuf map types](https://developers.google.com/protocol-buffers/docs/proto3#maps)

An important distinction is that this protobuf map type still uses a list of (k, v) tuples in the serialized transport, rather than something unique. Doing the same ourselves would maybe be disappointing to those looking to leverage JSON objects as Maps, but would allow support for far more key types.

This would require some client-side transformation to convert the entry pairs back into a map type. Presumably a code generator or something like that could do that for you.

I can see two paths forward for this space.

1. An explicit `Map` type, whatever the syntax. A rough idea could be `{ key: value }` as lists are `[value]`, so as an example `{ Int: MyType }`
2. A community standard, similar to `*Connection`, roughly `*Map` so examples of each:

    ```graphql
    # Example of existing Connection pattern:
    type UserConnection {
      edges: [UserEdge]
    }
    type UserEdge {
      node: User
    }

    # Potential example of a Map pattern:
    type UserMap {
      entries: [UserEntry]
    }
    type UserEntry {
      key: ID
      value: User
    }
    ```

I think what @mjmahone is getting at is that maybe there exists a higher level type we could introduce that would allow both Connection and Map to be defined in user space, but I'm slightly worried that side-steps the part where client tooling needs to generate code to handle these types.

Certainly the community standard would be the easiest path since there's no change to spec necessary and we can immediately build tools that use it. But it would be more awkward to query than something built into the spec.

nojvek:
Thanks @leebyron

I just wanted to say this was my first time at GraphQL WG meeting. It was very well organized and agenda being followed on time. This is pretty rare. 

Even though folks had strong opinions, it directed towards the technical details rather than personal. I felt quite welcome. Thank you.

I did learn that GraphQL can be a superset of JSON responses. Previously my assumption was that GraphQL responses are always JSON. That changes the design constraints.

I am okay if we close this PR / RFC. Happy for someone else to open another more concrete proposal.

nojvek:
I also very much like the proposal of built in / user-extendable generics-ish types by @mjmahone

e.g `Pair<Val1T, Val2T>` or `Map<KeyT, ValT>`. 

But having worked in Typescript's codebase, generics can get pretty wild fast. So there needs to be a more trimmed down proposal.

---

As for the transformation proposal by @IvanGoncharov, I feel it has legs. We shouldn't discount developer user experience.

GraphQL's alias syntax is powerful. One could say that is possible if frontend made multiple GraphQL queries and stitched it with a rename. Alias makes this much easy.

Currently consuming Dictionary structures from GraphQL is painful. Doesn't have to be JSON objects. JS has `Map` type, Python has `Dict` type, Go has `map` type. Most languages have built in map types that can handle any Key, Value pair of types.

I do agree that having keys restricted to Strings is bad design.

GraphQL supporting a List -> Map transform that can work with a variety of key/ value types is more powerful.

I feel there is a world where both @mjmahone and @IvanGoncharov's ideas can live together. Will have to do some deeper thinking.

leebyron:
> But having worked in Typescript's codebase, generics can get pretty wild fast. So there needs to be a more trimmed down proposal.

Totally agree. Our existing very simple type system already has a fair amount of complex ramifications, and I'd be very worried without some firm constraints.

We also need to consider the ecosystem around the query & schema language itself. Introspection, code generators, and more need to know how to make the best use of these things. I think this is all solvable, but we should be confident that it is (and reasonably so, such that the ecosystem doesn't constrain too much) before we seriously consider it

nojvek:
I'm closing this proposal PR as a no go. We can always open another proposal keeping some of the discussion points in this thread in consideration.

acao:
For the record, please don’t avoid new language features on our behalf! Graphiql, the lsp server and playground use the same underlying parser and language service, we are set up to handle new language features. We added interfaces implements interfaces quite easily. Also remember that there is a ton of graphql tooling  outside the js/ts space, often using a graphql-js mirror reference  implementation in their language, so a reference PR to graphql-js has ripple effects across the many languages that use graphql

PR #894, opened 2021-10-18
Define 'operation execution', clarifying 'before execution begins'

benjie:
**UPDATE: 2025-07-03**

There are various parts of the spec that relate to a concept of "execution":

- "before execution begins" defines the boundary between request error/field error
- "the request must fail without execution" indicates that no `data` should be produced
- "validation is performed in the context of a request immediately before execution"

All of these things actually happen during `ExecuteRequest()`, which makes it confusing to the reader that `ExecuteRequest()` is not itself considered execution despite the name.

It seems that **the boundary that we refer to as "execution" is actually the [Executing Operations](https://spec.graphql.org/draft/#sec-Executing-Operations) section** of the spec. So I have replaced the term "execution" in the relevant places with "operation execution", defined this new term, and slightly tweaked the algorithms so that they adhere to this division. I've also added a couple of clarifying notes.


IvanGoncharov:
My second attempt, was to produce "non-confusing criteria" 😄 
What if we just say that `data` should be null only due to "error propagation to the root":
https://spec.graphql.org/draft/#sec-Handling-Field-Errors
De facto it's the only algorithm in a spec that explicitly requires setting `data` to null.
It is very easy to check programmatically, and I actually think it matches the original intention.

benjie:
I think that's a fair assessment: `data` should only be `null` in the event of a field error (including a non-null assertion) being triggered on a root field. In all other cases it should be either an object (map), or not be present. (This comment not intended to be taken as a strictly worded specification.)

JoviDeCroock (approved):

spec/Section 7 -- Response.md
```diff
@@ -58,6 +58,10 @@ present in the response.
 If an error was raised during the execution that prevented a valid response, the
 `data` entry in the response should be `null`.
 
+Note: Request errors (including those raised during {ExecuteRequest()}) occur
+before execution begins; when a request error is raised the `data` entry should
```
martinbonnin, line 62: This is very hard for my brain to process. I may already be in `ExecuteRequest()` but still before execution begins?This feels weird.

Do we have a formal definition of what "execution" is and when it starts that we could link to maybe?

benjie: That's what the purpose of this PR is :100:

Execution begins the moment that the `data` property should be added to the response, which is when the root selection set is executed (after validating variables, establishing a stream in the case of subscriptions, etc).

benjie: I've taken a different tact by defining execution, see what you think.

martinbonnin:
```
 We define _execution_ as the process of executing the operation's root
selection set through {ExecuteSelectionSet()}, thus _execution_ begins when
{ExecuteSelectionSet()} is called for the first time in a request. The
{ExecuteRequest()} algorithm is a preamble for _execution_.
```

This is clarifying but also also adding more language to processs. I think a rename of `ExecuteRequest()` would be even more clarifying.

benjie: Why not both?

martinbonnin: Agreed.

mjmahone (approved):

benjie:
I have rebased this on top if #1039 and merged in @martinbonnin's changes in #1154 and the applied further editorial on top based on the discussions at last months WG, restoring "execution" where appropriate, eliding "Execute" (or "Process" or "Perform") from algorithm names, and making it clear that you:

- **process** a request
- **perform** an operation
- **execute** a selection set

I've further clarified that you validate the _document_ not the _request_, and you can execute another request without validation if the _document_ is exactly the same - the _request_ doesn't have to be exactly the same! (Longstanding issue.)

martinbonnin (approved):

BoD (approved):

leebyron:
Hey @benjie it looks like rebasing this may be more pain than just replicating the intent of this change in a new PR given the other changes that we've merged over the last few months.

I'll try to replicate it

leebyron:
Reopening to track a rebase

leebyron:
Did my best to re-introduce this change in https://github.com/graphql/graphql-spec/pull/1174 as written

benjie:
As mentioned in [this comment](https://github.com/graphql/graphql-spec/pull/1174#issuecomment-3028463041); I have:

Merged the change from #1174 back into this PR [^1] and have then[^5]:

1. wound back the changes to algorithm names[^2]
2. renamed "execution" to "operation execution" throughout
3. used the existing "Executing Operations" section as the point at which "operation execution begins" [^3]
4. moved establishing the source event stream into `ExecuteRequest()`[^4]

All in all I think this makes the spec text a lot clearer without having to rename any algorithms or introduce different verbs.

[^1]: commit [`7bafeaf` (#894)](https://github.com/graphql/graphql-spec/pull/894/commits/7bafeaf25423a68a687bbba1c461d76bcbd33e26) has no diff against this PR
[^2]: also removing "perform" and "process" verbs
[^3]: which feels really natural, right?
[^4]: since it happens "before execution begins"
[^5]: you can see my changes since this PR here: https://github.com/graphql/graphql-spec/pull/894/files/7bafeaf25423a68a687bbba1c461d76bcbd33e26..94241a78d020fb0bdeea483a0f50d54beff890ba - but personally I'd just read the <a href="https://github.com/graphql/graphql-spec/pull/894/files">full diff</a> as it's simpler

leebyron:
@benjie I like this new draft! Its a much tighter editorial change if it isn't renaming algos. I think this new "operation execution" term and definition seems right to me. "ExecuteRequest" remains slightly weird in the context of having both "execution error" and "request error" but I think your clarification note mitigates that pretty reasonably.

I'll go through this again, but I think this has a clear path to merge

mjmahone (approved):
This feels, even with the subscription piece discussed in the WG, to be a net improvement over the status quo, so I'd be happy with this then tinkering around the subscription phase question later.

benjie:
I've submitted this to be discussed (and hopefully merged) on Thursday.

martinbonnin (approved):
LGTM

leebyron (approved):

PR #908, opened 2021-12-10
add note regarding the normative force of directives

yaacovCR:
This is to better anticipate the nature of defer/stream directives as client side recommendations rather than mandates.


yaacovCR:
Exact language requires some changes, this is just a temporary placeholder so I don’t forget about it, other thoughts also welcome of course!

spec/Section 2 -- Language.md
```diff
@@ -1203,6 +1203,9 @@ As future versions of GraphQL adopt new configurable execution capabilities,
 they may be exposed via directives. GraphQL services and tools may also provide
 any additional *custom directive* beyond those described here.
 
+Note: Directives simply describe additional information. The normative force of
+the directive depends on the meaning of each directive.
```
yaacovCR, line 1207:
```suggestion
the directive, if any, is as described where the directive is encountered.
```

spec/Section 2 -- Language.md
```diff
@@ -1203,6 +1203,9 @@ As future versions of GraphQL adopt new configurable execution capabilities,
 they may be exposed via directives. GraphQL services and tools may also provide
 any additional *custom directive* beyond those described here.
 
+Note: Directives simply describe additional information. The normative force of
+the directive, if any, is as described where the directive is encountered.
```
yaacovCR, line 1207:
```suggestion
the directive may depend on its arguments, the location it is applied, and/or the context in which it is encountered.
```

yaacovCR:
Directives "direct" behavior only indirectly, by providing information.
Directive "direct" behavior, but do not mandate it.

The existing specified directives that change execution behavior have "must" language attached (skip, include), but custom directives, and potentially future specified directives (defer/stream), may have only "should" language.

yaacovCR:
https://github.com/robrichard/defer-stream-wg/discussions/7

spec/Section 2 -- Language.md
```diff
@@ -1203,6 +1203,10 @@ As future versions of GraphQL adopt new configurable execution capabilities,
 they may be exposed via directives. GraphQL services and tools may also provide
 any additional *custom directive* beyond those described here.
 
+Note: Directives simply provide additional information. The normative force
+of directives may vary and may also depend on a directive's arguments or
+location, as well as the context in which the directive is encountered.
+
```
leebyron, line 1209: Except for those spec'd by this document? Those should be normative right?

mjmahone:
Is there an existing paragraph we should be rewording such that new readers of the spec come to the understanding that we're trying to get to?

Basically: what existing text is likely to cause confusion, and is the reason we feel like we need this additional paragraph?

yaacovCR:
> Is there an existing paragraph we should be rewording such that new readers of the spec come to the understanding that we're trying to get to?
> 
> Basically: what existing text is likely to cause confusion, and is the reason we feel like we need this additional paragraph?

I don't think there is existing test that causes confusion -- I think it's more like the basic confusion raised in the working group. Tools (such as validators, executors) MUST respect all directives, but only at the normative level at which they are defined, which can vary. Spec directives (skip, include) describe MUST behavior, future planned directives are at SHOULD (defer/stream, to allow for advanced performance enhancing behavior by custom servers) and future directives could (but probably won't?) be at some other normative level.

Spec and custom directives themselves describe metadata, and so as "specified" within the spec or by a potential "specifiedByURL", they have variable normative force.

spec/Section 2 -- Language.md
yaacovCR: tried to clarify that all directives MUST be respected, but that they may prescribe behavior changes that are OPTIONAL

PR #727, opened 2020-06-02
Format the spec with prettier

benjie:
_This PR has been work in progress for a while; and a lot has changed. Please see the edit history for previous versions of this comment._

Thanks to advances to `spec-md`, we can now format the GraphQL spec with `prettier`! This means consistent formatting throughout the document, and no need for PR comments about there being too many or two few linebreaks :wink: 

Prettier has been desired in the GraphQL spec for [over 3 years](https://github.com/graphql/graphql-spec/pull/403#issue-165727240) - I'm excited to have worked with @leebyron over the past year to have finally made this a reality! :tada: 

This PR is formed of two commits:

1. The first commit adds the tooling around prettier, including the `yarn format` (or `npm run format`) command which formats the repository.
2. The final commit simply runs `yarn format` and commits the results.

To view the HTML diff:

1. checkout `main`
2. run `yarn build`
3. run `mv public/draft/index.html before.html` (using `mv` will also ensure that the old output cannot be reused)
4. checkout the final commit
5. run `yarn build`
6. run `mv public/draft/index.html after.html`
7. At this point, before.html and after.html differ due to different `data-source` URLs; let's remove these:
8. run `cat before.html | sed 's/ data-source="[^"]*"//g' > before-nosource.html`
9. run `cat after.html | sed 's/ data-source="[^"]*"//g' > after-nosource.html`

You can now see the versions with data-source removed are identical:

```
$ shasum before.html after.html before-nosource.html after-nosource.html
2380c111a29891fa4e8547ef5cee32d429299982  before.html
280252b0fe1aac6065cb280470efb1a096447fa4  after.html
09a7ae84464cc49f6affab9bd544ad4d687837a1  before-nosource.html
09a7ae84464cc49f6affab9bd544ad4d687837a1  after-nosource.html
```

PR #925, opened 2022-01-24
consistent indentation and punctuation

thomasheyenbrock:
Hello dear spec-maintainers 👋 I noticed some inconsistencies in terms of indentation and punctuation in the validation section, tried to clean it up and make it all consistent.

I also did a tiny change to the spec for `Single root field`, where I lifted up one declaration that is not depending on the context of the loop.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -95,9 +95,9 @@ union HumanOrAlien = Human | Alien
 
 **Formal Specification**
 
-- For each definition {definition} in the document.
-- {definition} must be {ExecutableDefinition} (it must not be
-  {TypeSystemDefinitionOrExtension}).
+- For each definition {definition} in the document:
+  - {definition} must be {ExecutableDefinition} (it must not be
+    {TypeSystemDefinitionOrExtension}).
 
 **Explanatory Text**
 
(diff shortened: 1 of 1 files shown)


spec/Section 5 -- Validation.md
```diff
@@ -249,14 +249,14 @@ query getName {
...
-- Let {selectionSet} be the top level selection set on {subscription}.
-- Let {variableValues} be the empty set.
-- Let {groupedFieldSet} be the result of {CollectFields(subscriptionType,
-  selectionSet, variableValues)}.
-- {groupedFieldSet} must have exactly one entry, which must not be an
-  introspection field.
+- For each subscription operation definition {subscription} in the document:
+  - Let {selectionSet} be the top level selection set on {subscription}.
+  - Let {variableValues} be the empty set.
+  - Let {groupedFieldSet} be the result of {CollectFields(subscriptionType,
```
benjie, line 256: You've dropped the definition of `subscriptionType` but it's used here

benjie (changes requested):
Thanks for doing this, I had meant to do it after the prettier PR (#727) was merged since I discovered a lot of these issues whilst working on that, but alas it took a while and completely slipped off my radar!

spec/Section 5 -- Validation.md
benjie: Sorry my bad, I misread the diff :+1:

benjie (approved):

PR #926, opened 2022-01-24
add note about selection sets in serial execution examples

thomasheyenbrock:
The examples for serial execution in the section `Normal and Serial Execution` actually use a query as operation type. This PR proposes to change these examples to mutations (which are actually required to be executed serially).

I also added in the variables in the first of these examples for the operation to be valid.

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -452,6 +452,9 @@ A correct executor must generate the following result for that selection set:
 }
 ```
 
+Note: The previous two examples depict only a selection set and _not_ a full
+document using the query shorthand.
+
 ### Field Collection
 
 Before execution, the selection set is converted to a grouped field set by


benjie (changes requested):
Thanks for submitting a suggestion :raised_hands: 

The spec text here states (emphasis mine):

> For example, given the following **selection set to be executed serially**:

The examples are **not** GraphQL documents, they are selection sets, your edits turn them into documents but that change is not desired.

thomasheyenbrock:
Aaah, got it 👍 Though it's easy to read over that and there's a bit of ambiguity here in the sense that one can't tell the difference from a selection set to a document using the query shorthand. What do you think about adding a note here to clear this up?

benjie:
> What do you think about adding a note here to clear this up?

Let's see what you have in mind :+1:

thomasheyenbrock:
Updated the PR description and replaced the commit with a new one that just adds a note 🙂

benjie:
I'm not convinced this is necessary given the text above the example explicitly states `selection set`, but I'll let others comment :+1:

PR #927, opened 2022-01-24
use `findDog` query from example schema only after defining it

thomasheyenbrock:
In the validation section there's an example that uses the `findDog` query before actually defining it in the type system. This PR adds the query right at the beginning of the section instead of extending the schema later. It also adds the required selection sets in the all examples that use this query.

I also propose renaming the `ComplexInput` type and the `complex` argument to something more verbose that fits into the rest of the schema and makes it a "real world" example.


leebyron:
Thank you for this improvement!

PR #936, opened 2022-03-31
Make UnionMemberTypes non-optional

fotoetienne:
In the Validation section, it is stated:

```A Union type must include one or more unique member types.```
(https://spec.graphql.org/draft/#sec-Unions.Type-Validation)

Thus it seems that UnionMemberTypes are not optional.


benjie:
As with the other types, a union can be declared (without members) and later extended:

http://spec.graphql.org/draft/#sec-Union-Extensions

PR #941, opened 2022-04-30
introduce new intersection type

yaacovCR:
see https://github.com/graphql/graphql-js/pull/3550


rivantsov:
sorry, did I miss something, a lot maybe? - where's the discussion? Straight PR into spec?

yaacovCR:
Open questions:

1. If intersection types include interfaces that implement interfaces, should the ancestor interfaces be required to be listed explicitly within the intersection? (My answer: yes, just like with interfaces.)
2. Should intersection types be allow to include other intersection types? (My answer: no, the "child" intersection types should just be listed explicitly within the "parent.")
3. Should one be allowed to query fields directly on the intersection? (My answer: no, intersections that define no union members will have no fields guarantees, so for consistency fragments on intersections should never be allowed.)

## Example and further discussion for question #1

```
type ParentInterface {
  someField: String
}

type ChildInterface implements ParentInterface {
  someField String
}

type TypeA implements ParentInterface & ChildInterface {
  someField: String
}

type TypeB implements ParentInterface & ChildInterface {
  someField: String
}

union MyUnion = TypeA | TypeB
```

Should we allow:
```
intersection MyIntersection = MyUnion & ChildInterface
```

vs. requiring
```
intersection MyIntersection = MyUnion & ParentInterface & ChildInterface
```

As we can see for above, ancestor interfaces are already required for readability reasons on object and interface types. The same argument could be made here -- and I would tend to agree with it! Note that the spec/implementation PRs do not yet require this.

## Example and further discussion for question #2

type SomeInterface {
  someField: String
}

type AnotherInterface {
  anotherField: String
}

type TypeA implements SomeInterface & AnotherInterface {
  someField: String
  anotherField: String
}

type TypeB implements AnotherInterface {
  anotherField: String
}

union MyUnion = TypeA | TypeB

intersection SomeIntersection = MyUnion & SomeInterface
intersection AnotherIntersection = MyUnion & AnotherInterface
```

Should we allow:
```
intersection IntersectingIntersection = SomeIntersection & AnotherIntersection
```

Or require:
```
intersection SomeIntersection = MyUnion & SomeInterface & AnotherInterface
```

One could argue that similarly to the required verbosity for interfaces, we should require intersections to explicitly list all of their members and not include any other intersections. That is my opinion, although, as above, the spec/PR do not yet reflect this stricter requirement.

## Example and further discussion for question #3

```
type Pet {
  name: String
}

type Dog implements Pet {
  name: String
}

type Cat implements Pet {
  name: String
}

union DogOrCat = Dog | Cat

intersection DogOrCatPet = DogOrCat & Pet

type Query {
  dogOrCatPets: [DogOrCatPet]  
}
```

Should we allow:

```
{
  dogOrCatPets {
    name
  }
}
```

or require:
```
{
  dogOrCatPets {
    ... on Pet {
      name
    }
  }
}
```

An argument in favor of allowing the former is that for this particular intersection, name is always defined, so this should be fine. An argument against allowing could be just from theory => the intersection itself does not define any fields, it just does so transitively through the interface it includes, so it is "wrong" to allow querying on the type. Because some intersections may contain only unions and not have any fields, it might be simpler to "teach" intersections as being more similar to unions, and never allow querying for fields directly on the intersection. I lean against allowing it, but not strongly.

**Comment by yaacovCR** [comment]
alternative to #939

**Comment by yaacovCR** [comment]
Open questions:

1. If intersection types include interfaces that implement interfaces, should the ancestor interfaces be required to be listed explicitly within the intersection? (My answer: yes, just like with interfaces.)
2. Should intersection types be allow to include other intersection types? (My answer: no, the "child" intersection types should just be listed explicitly within the "parent.")
3. Should one be allowed to query fields directly on the intersection? (My answer: no, intersections that define no union members will have no fields guarantees, so for consistency fragments on intersections should never be allowed.)

## Example and further discussion for question 1

```
type ParentInterface {
  someField: String
}

type ChildInterface implements ParentInterface {
  someField String
}

type TypeA implements ParentInterface & ChildInterface {
  someField: String
}

type TypeB implements ParentInterface & ChildInterface {
  someField: String
}

union MyUnion = TypeA | TypeB
```

Should we allow:
```
intersection MyIntersection = MyUnion & ChildInterface
```

vs. requiring
```
intersection MyIntersection = MyUnion & ParentInterface & ChildInterface
```

As we can see for above, ancestor interfaces are already required for readability reasons on object and interface types. The same argument could be made here -- and I would tend to agree with it! Note that the spec/implementation PRs do not yet require this.

## Example and further discussion for question 2

```
type SomeInterface {
  someField: String
}

type AnotherInterface {
  anotherField: String
}

type TypeA implements SomeInterface & AnotherInterface {
  someField: String
  anotherField: String
}

type TypeB implements AnotherInterface {
  anotherField: String
}

union MyUnion = TypeA | TypeB

intersection SomeIntersection = MyUnion & SomeInterface
intersection AnotherIntersection = MyUnion & AnotherInterface
```

Should we allow:
```
intersection IntersectingIntersection = SomeIntersection & AnotherIntersection
```

Or require:
```
intersection SomeIntersection = MyUnion & SomeInterface & AnotherInterface
```

One could argue that similarly to the required verbosity for interfaces, we should require intersections to explicitly list all of their members and not include any other intersections. That is my opinion, although, as above, the spec/PR do not yet reflect this stricter requirement.

## Example and further discussion for question 3

```
type Pet {
  name: String
}

type Dog implements Pet {
  name: String
}

type Cat implements Pet {
  name: String
}

union DogOrCat = Dog | Cat

intersection DogOrCatPet = DogOrCat & Pet

type Query {
  dogOrCatPets: [DogOrCatPet]  
}
```

Should we allow:

```
{
  dogOrCatPets {
    name
  }
}
```

or require:
```
{
  dogOrCatPets {
    ... on Pet {
      name
    }
  }
}
```

An argument in favor of allowing the former is that for this particular intersection, name is always defined, so this should be fine. An argument against allowing could be just from theory => the intersection itself does not define any fields, it just does so transitively through the interface it includes, so it is "wrong" to allow querying on the type. Because some intersections may contain only unions and not have any fields, it might be simpler to "teach" intersections as being more similar to unions, and never allow querying for fields directly on the intersection. I lean against allowing it, but not strongly.

**Comment by yaacovCR** [comment]
> sorry, did I miss something, a lot maybe? - where's the discussion? Straight PR into spec?

This is a continuation of/alternative to the "unions implementing interfaces" work I presented at the last WG.
I started to post some discussion thoughts here but have moved the, to: https://github.com/graphql/graphql-wg/discussions/944

File: spec/Appendix B -- Grammar Summary.md [hunk 1]
```diff
@@ -343,6 +343,19 @@ UnionTypeExtension :
 - extend union Name Directives[Const]? UnionMemberTypes
 - extend union Name Directives[Const]
 
+IntersectionTypeDefinition : Description? intersection Name Directives[Const]?
+IntersectionMemberTypes?
```
rivantsov, line 347: so IntersectionMemberTypes element is optional? we can have intersection with empty list of types?

yaacovCR: It is same as Union it could be extended later. So valid to be empty initially in sdl. When validated, it has to be non/empty if I recall correctly, just like unions. But we do allow an intersection with non empty list of member types and currently empty list of possible types due to conflict.

Just like interfaces without implementing types are allowed. I believe :)

spec/Section 3 -- Type System.md
```diff
@@ -287,6 +288,11 @@ A `Union` defines a list of possible types; similar to interfaces, whenever the
 type system claims a union will be returned, one of the possible types will be
 returned.
 
+An `Intersection` defines a list of constraining abstract types. If a field
+claims it returns an Intersection type, it will return only types that are
+contained within all of the Intersections's Unions and types that implement all
+of the Intersection's Interfaces.
+
```
rivantsov, line 295: 1. what is Intersection's Unions? in the grammar rule you mention IntersectionMemberTypes
2. so... if at runtime field resolver returns a value of type that IS one of union types but does not implement one of interfaces - then what? it is replaced with null? what if field is not nullable? Hell confusing

yaacovCR: At runtime, only concrete object types can be returned as the actual type just like with any other abstract type.

yaacovCR: And yes, an intersection's unions are its member types that are unions. I copied the grammar from unions where even though object types are the only currently allowed member types, that is not specified at the grammar level, it's specified only as part of validation.

yaacovCR: If a non possible type is returned at runtime you get the same error as if you return a non implementing type when using an interface or an object type not in the Union.

In the implementation changes, there was no need to directly touch Execution because the only thing that changes is what goes on in the isSubType method

PR #944, opened 2022-05-05
add @experimental

martinbonnin:
Pull request for https://github.com/graphql/graphql-spec/issues/943

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -2094,25 +2094,25 @@ type ExampleType {
 ### @experimental
 
 ```graphql
-directive @experimental(reason: String = "Experimental") on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
+directive @experimental(
+  reason: String = "Experimental"
+) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
 ```
 
 The `@experimental` _built-in directive_ is used within the type system
 definition language to indicate experimental portions of a GraphQL service's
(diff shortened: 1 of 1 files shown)


spec/Section 3 -- Type System.md
```diff
@@ -2067,6 +2087,26 @@ type ExampleType {
 }
 ```
 
+### @experimental
+
+```graphql
+directive @experimental on FIELD_DEFINITION | ENUM_VALUE
```
jturkel, line 2093: It would be really useful to support `@experimental` on argument definitions and input field definitions too.

martinbonnin: Do you have a use case in mind? I mainly copy/pasted the `@deprecated` definition cause I liked the symmetry there. Is there any reason to allow this for `@external` but not `@deprecated`?

jturkel: We (Salsify) support lots of equality filtering on fields in our GraphQL API e.g. retrieve a list of products in a certain category. We're experimenting with also supporting more complex filtering via more advanced operators like `and`, `or`, `not`, `<`, etc. These additional inputs are very much experimental and we don't yet want to commit to maintaining backwards compatibility but we do want clients trying them out and giving us feedback on the API ergonomics. It would be great to mark the advanced filtering field arguments and input fields as `@experimental`.

It looks like #805 hasn't actually merged yet but the RFC for adding `@deprecated` to input fields/field arguments seems to have been approved so hopefully this would be symmetrical soon enough.

martinbonnin: Thanks for the context! Symmetry is always good 🤗 . I just pushed changes to add arguments and input fields.

martinbonnin:
Closing in favor of https://github.com/graphql/graphql-wg/pull/1006

PR #947, opened 2022-05-20
Add spec text for KnownOperationType

yaacovCR:
see: https://github.com/graphql/graphql-js/issues/3592

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -129,7 +129,7 @@ extend type Dog {
 
 ## Operations
 
-### Operation Types
+### Operation Type Existence
 
 **Formal Specification**
 
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -129,6 +129,39 @@ extend type Dog {
 
 ## Operations
 
+### Operation Types
+
+**Formal Specification**
+
+- For each {operation} in the document.
+- Let {operationType} be the type of {operation}
+- Let {operationRootType} be the root type in {schema} for {operationType}.
+- {operationRootType} must exist.
+
+**Explanatory Text**
+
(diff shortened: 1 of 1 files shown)


yaacovCR:
See `graphql-js` PR: https://github.com/graphql/graphql-js/pull/3601
Spec enhancement suggested by @benjaminjkraft

benjaminjkraft:
Neat.  Not sure what the style in the spec is -- would it make sense to refer to this from 3.3.1 which describes this more informally ("if [the mutation root type] is not provided, the service does not support mutations")?

benjie:
The text as currently proposed will conflict with https://github.com/graphql/graphql-spec/pull/825/files#diff-607ee7e6b71821eecadde7d92451b978e8a75e23d596150950799dc5f8afa43e which is currently RFC2. Either #825 needs to be refactored to not use mutation in the examples, or this should use a separate example schema for this validation rule. May I suggest we follow the latter approach as it allows our example schema to grow over time.

yaacovCR:
> Neat. Not sure what the style in the spec is -- would it make sense to refer to this from 3.3.1 which describes this more informally ("if [the mutation root type] is not provided, the service does not support mutations")?

There seem to be a few references to Validation Rules, namely:

https://github.com/graphql/graphql-spec/blame/main/spec/Section%202%20--%20Language.md#L1134-L1135

We could try that here for sure, in my opinion.

May I suggest something => perhaps it makes sense for you to "take over" this PR, considering you will be leading the discussion at the WG and this rule was your idea to begin with. I sometimes wish GitHub had a UI for transferring ownership of PRs => I think it would be relatively easy for you to reopen a new one, and -- if you do -- I will close this one in favor of yours.

benjaminjkraft:
Sure, I can take over the PR and add a separate schema.

benjaminjkraft:
Replaced by #955.

PR #948, opened 2022-05-23
RFC: OneOf Objects

benjie:
This is a follow-up to #825 that proposes adding the `@oneOf` directive to object types (output objects).

This is a draft right now as I've not yet had time to think through it well enough, but it raises a few questions:

1. The proposed resolution algorithm is a slight divergence from the status quo, omitting fields from the response rather than setting them null. This makes it align with #825, but is it desired? Should we change #825 to only require one non-null field rather than just one field?
2. Should we allow aliases? The current proposed resolution algorithm ensures that every identical field requested on a `@oneOf` selection set will get the same value, but it does so in a somewhat awkward way. Forbidding aliases could remove the need for this.
3. What happens if a `@oneOf` is queried as part of an interface or union? The proposed resolution algorithm allows for this by leaving it up to the implementation, I think this is probably best unless anyone has a better idea?

TODO:

- [ ] When validating the schema, forbid `@oneOf` on operation types.
- [ ] Tidy up the language around maps/accessing keys.


yaacovCR:
> Should we change [RFC: OneOf Input Objects #825](https://github.com/graphql/graphql-spec/pull/825) to only require one non-null field rather than just one field?

Could you expand on this sentence? Thanks so much!

benjie:
Indeed! Sorry this is mostly “notes for myself” right now 😬

benjie:
> Can that happen? I don't think this should be valid in the schema at all:

`@oneOf` types are indeed forbidden from implementing interfaces in the current proposal, I've edited that out of the questions - thanks!

> [Regarding union] I don't see the open question here. A OneOf Object Type still is a named type, so it'll have a __typename, and on a schema like

I've designed it this way currently, yes. That was the result of subtleties around how the oneof value is actually resolved; for example if we were to dictate that a oneof value has to be "an unordered map with one entry" then there'd be no way to generally disambiguate what type it is:

```graphql
type A @oneOf {
  a: Int
  b: String
}
type B @oneOf {
  a: Float
  c: String
}
union U = A | B


query Q {
  u { __typename }
}
```

When the `u` field resolves, if the type returned is just a map with one key such as `{"a": 1}` then we cannot tell whether it was an A or a B.

My current proposal handles this by having the system dictate the key, value and type; however, this new "oneof resolver" feels a bit hacky, and I wonder if there's a better way that doesn't open up the problems detailed above.

ab-pm:
We do have some use cases for this, and would really liked to have used the `@oneOf` annotation. Our current schema contains
```graphql
type User { id: Int! }
type Usergroup { id: Int!, members: [Member!] }
union Member = User | UserGroup

"""An input describing a single `UserGroup` member. Must contain exactly one of the fields."""
input MemberInput { # @oneOf
  """Passed when referring to a `User`."""
  userId: Int
  """Passed when referring to a `Usergroup`."""
  usergroupId: Int
}

extend type Mutation {
  addMemberToGroup(usergroupId: Int!, member: MemberInput!): Usergroup
  addMembersToGroup(usergroupId: Int!, member: [MemberInput!]!): Usergroup
}
```
Sure there's lots of other approaches to solve this (e.g. adding `memberId` surrogate keys to users and usergroups that are assigned unambiguously, or using two separate `addUserToGroup` and `addUsergroupToGroup` mutations) but I really liked the "union input" approach since it works also in nested input objects (like in `addMembersToGroup`) and nicely mirrors the output structure.

benjie:
From [this comment](https://github.com/graphql/graphql-spec/pull/825#issuecomment-931080230):

> At their talk "Automated GraphQL Schema Generation from ER-based Service Modeling" at GraphQL Conf yesterday, LinkedIn's Min Chen revealed how GraphQL union type was not sufficient to map their existing types, so they have converted it to regular object types where all fields are nullable and one is expected to be set - which is exactly how we would map Oneof to output types. They were mixing object types and scalars into their union - one of the big features that Oneof would unlock for GraphQL output polymorphism. It was interesting to hear this GraphQL limitation and see this solution serindipitously from a large vendor - I wonder how we can discover others in the ecosystem handling this same problem?

Issue #711, opened 2020-04-20
union does not allow sub-union or covariance


VladimirAlexiev:
In RDF it's quite common to use subclasses, and somewhat common to use union types (eg `schema:rangeIncludes` states a field can take values from several different classes; `owl:unionOf` does a similar thing). 

RDF is based on formal logic, so it considers classes (types) as **sets**. In this interpretation, it's clear how to determine subtypes (especially useful for field **covariance**). Eg

```graphql
interface A
interface B
interface C
type A1 implements A
type B1 implements B
type C1 implements C
union U1=A|B|C
union U2=A1|B1|C1   # subtype of U1 because it uses subclasses
union U3=A|B        # subtype of U1 because it uses fewer union members
union U4=A|A1|B|C   # equivalent to U1 because A subsumes A1
union U5=A1|B1      # subtype of U1, U2 and U3
```

But the spec says:
- [fieldType (child) is an Object type and implementedFieldType (parent) is a Union type and fieldType is a possible type of implementedFieldType](https://spec.graphql.org/draft/#sel-HAHZhCJCHCAACKB5lK)
- [Member types of a Union type must all be Object base types; Scalar, Interface and Union types must not be member types of a Union](https://spec.graphql.org/draft/#sel-HAHdfFDABEBlEn_b)

This means that `U1, U3, U4` are not allowed, and neither of them can be used in an object type field.

This creates severe limitations on what polymorphic constructs (in particular RDF constructs) can be mapped to GraphQL.

(I searched here for "union but not 'input union'" and came up only with #518 as a possibly related issue. cc @derek-miller @mavilein)

evgen4uk:
Unfortunately, there are not many discussions here but as for me, support of covariance is very crucial.
Let me show a simple example, I will start with schema without union and then I will add unions
```
interface Animal {
    parent: Animal
}

type Cow implements Animal {
    parent: Cow
}

type Wolf implements Animal{
    parent: Wolf
}

```
In the above example, we see that parent of a cow can be only a cow, not any animal.
Now let's add a union, for this purpose I will add logic that animals can be cloned, so technically he doesn't have a parent, just information about cloning

```
type CloningInfo
union AnimalParentOrCloninginfo = Cow | Wolf | CloningInfo
```
It may be not really a good example, but I just want to show that field `parent` can return different types

_For better reading of this issue, I will use "anonymous" unions and will define them on the field definition_
```
interface Animal {
    parent: Cow | Wolf | Error
}

[...]

PR #950, opened 2022-05-27
allow unions to include interfaces and unions

yaacovCR:
Complements #939
Addresses #711

Similar to #939, this PR expands the robustness of the type system by allowing types that actually fulfill interfaces to be recognized as such by the GraphQL type system.

```graphql
interface CloningInfo {
  ...
}

union CowOrWolf implements Animal = Cow | Wolf  # allowed by #939
union CowOrCloningInfo = Cow | CloningInfo  # allowed by this PR, note that CloningInfo is an interface
union WolfOrCloningInfo = Wolf | CloningInfo # allowed by this PR, note that CloningInfo is an interface

# note that here we are marking unions explicitly as included within a union.
# Adding a type to the CowOrWolf union will automatically add it to the ParentUnion
# We could also consider adding a constraint on the union definition, see below discussion
union Parent = CowOrCloningInfo | WolfOrCloningInfo | CowOrWolf | Cow | Wolf | CloningInfo

interface Animal {
  parent: Parent
}

type Cow implements Animal {
    parent: CowOrCloningInfo  # unlocked by this PR
}

type Wolf implements Animal{
    parent: WolfOrCloningInfo # unlocked by this PR
} 
```

### Unions

With regard to unions, the goal is to explicitly mark some unions as members of other unions. We have two alternatives:

(A). Let unions include unions as members, as shown above. We could (or could not) require that all members of the unions also be listed (similar to how interfaces implementing child interfaces are required to explicitly list the parent.

Pro:
1. Simple to reason about, fits with how unions currently work.
Con:
1. Could lead to some automatic behavior, when adding a type to a union that is part of a union, the type gets added to multiple unions. This is ameliorated by requiring all child union members to be explicitly listed.
2. Union definitions could start to get pretty long if all the combinations must be listed therein.

(B) Add an additional optional constraint on the union requiring all of the members to be members of some other union, similar to how we have resolved #939. 

Potential Syntax: 

```graphql
union CowOrWolf implements Animal, subtypes Parent = Cow | Wolf 
union CowOrCloningInfo subtypes Parent = Cow | CloningInfo 
union WolfOrCloningInfo subtypes Parent = Wolf | CloningInfo

union Parent = Cow | Wolf | CloningInfo
# cf
# union Parent = CowOrCloningInfo | WolfOrCloningInfo | CowOrWolf | Cow | Wolf | CloningInfo
```

### Interfaces

For interfaces that are members of unions, it would _not_ seem to make sense to require all the implementations of the interfaces to be listed independently. The whole point is that it is often just as useful to indicate that several interfaces might be returned as it is that several individual member types might be returned. For unions, we also have potentially multiple layers of nesting (unions of unions of unions) for which it would be extremely helpful to require the individual member types to be listed (or to use the second syntax above) while we don't have the same issue with interfaces.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1462,12 +1462,14 @@ Union types have the potential to be invalid if incorrectly defined.
 2. The member types of a Union type must all be Object, Interface or Union
    types; Scalar and Enum types must not be member types of a Union. Similarly,
    wrapping types must not be member types of a Union.
-3. A parent Union must explicitly include as member types of all child Union
-   members.
+3. A Union type must explicitly include all possible types of any included
+   abstract types.
    1. Let this union type be {unionType}.
-   2. For each {memberType} declared as a member of {unionType}, if {memberType}
-      is a Union type, all of the members of {memberType} must also be members
-      of {unionType}.
+   2. For each {memberType} declared as a member of {unionType}:
+      1. If {memberType} is a Union type, all of the members of {memberType}
+         must also be members of {unionType}.
+      2. If {memberType} is an Interface type, all implementations of
+         {memberType} must also be members of {unionType}.
 
 ### Union Extensions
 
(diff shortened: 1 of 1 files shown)


yaacovCR:
Addresses #711

yaacovCR:
Now implemented in https://github.com/graphql/graphql-js/pull/3682

I added requirement that all subtypes of included abstract types must be explicitly includes within the union. This can always be relaxed later.

tobiasBora:
Any news on this? It has been around for 2 years now, and forbidding unions of unions create massive code duplication, possible errors, maintainability headaches, and poor readability. Consider for instance:
```
type QuestionMultipleChoice {
  question: String
  nbChoices: Int!
}

type QuestionText {
  question: String
}

type QuestionNumber {
  question: String
}

type WrongToken {
  _: Boolean
}

union Question = QuestionMultipleChoice | QuestionText | QuestionNumber

# What I'd like:
# union NewQuestionResult = Question | WrongToken
# What I need to write, **for each function that returns different errors**
union NewQuestionResult =  QuestionMultipleChoice | QuestionText | QuestionNumber| WrongToken

type Mutation {
  newQuestion: NewQuestionResult
}

type Query {
  hello: String
}
```

PR #955, opened 2022-06-01
Add validation rule that operation types exist

benjaminjkraft:
Right now the spec says that, for example, if the schema does not define
a mutation root type, then the schema does not support mutations.  But
there's no validation rule for it, which means that many parsers
(including graphql-js) treat a mutation as valid against such a schema.
(Indeed, many end up considering *any* mutation as valid, since they
don't know what type to validate the root selection set against.)  This
commit adds a validation rule to make the schema text explicit.

Slated for discussion at the June 2 working group meeting.
Replaces #947.  See also graphql/graphql-js#3592.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -129,6 +129,42 @@ extend type Dog {
 
 ## Operations
 
+### All Operation Definitions
+
+#### Operation Type Existence
+
+**Formal Specification**
+
+- For each operation definition {operation} in the document.
+- Let {operationRootType} be the root type in {schema} corresponding to the
+  type of {operation}.
+- {operationRootType} must exist.
+
+**Explanatory Text**
+
+Each operation must reference an operation type which has a valid root type
+in the schema.
+
+For example given the following schema:
+
+```graphql example
+type Query {
+  hello: String
+}
+```
+
+The following operation is valid:
+
+```graphql example
+query helloQuery { hello }
+```
+
+While the following operation is invalid:
+
+```graphql example
+mutation goodbyeMutation { goodbye }
+```
+
 ### Named Operation Definitions
 
 #### Operation Name Uniqueness
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -148,12 +148,14 @@ type system where those operations begin.
 The {`query`} root operation type must be provided and must be an Object type.
 
 The {`mutation`} root operation type is optional; if it is not provided, the
-service does not support mutations. If it is provided, it must be an Object
-type.
+service does not support mutations (see
+[Operation Type Existence](#sec-Operation-Type-Existence)). If it is provided,
+it must be an Object type.
 
 Similarly, the {`subscription`} root operation type is also optional; if it is
-not provided, the service does not support subscriptions. If it is provided, it
-must be an Object type.
+not provided, the service does not support subscriptions (see
+[Operation Type Existence](#sec-Operation-Type-Existence))s. If it is provided,
+it must be an Object type.
 
 The {`query`}, {`mutation`}, and {`subscription`} root types must all be
 different types if provided.


benjaminjkraft:
Ok!  I don't feel strongly about mentioning it in section 3.

BTW, the other thing to discuss is whether the rule that a schema must have a query and that the root types must be object types should have explicit validation language as well.  It seems like the schema side is a lot less consistent about that, so I wasn't sure.  We can discuss tomorrow.

benjie:
The validation section (section 5) is about validation of requests; the schema has its own validation rules in the relevant places, for example the rules about root operation types can be found here:

https://spec.graphql.org/draft/#sec-Root-Operation-Types

Shane32:
> BTW, the other thing to discuss is whether the rule that a schema must have a query and that the root types must be object types should have explicit validation language as well.

It seems to me that validation of a document is different than validation of a schema.  Section 5 validates "if a request is syntactically correct" and " is unambiguous and mistake‐free in the context of a given GraphQL schema".  So I don't believe section 5 is intended to cover validation of a schema, and I would not add any schema validation language there.

Rather, section 3 "describes the capabilities of a GraphQL server", and already has language stating that a query root type must exist, etc.

Perhaps it is worth considering rewriting all of section 3 to include schema validation rules more like the "formal specification" syntax used in section 5.  However, this may be more challenging in that a schema is not required to be represented by a SDL.

benjaminjkraft:
Yeah, to be clear, we would not add schema validation to section 5, if we wanted to add that it would go in section 3 and there's some precedent for validation rules there. As you say making that validation more formal is probably a bigger project, so I left it out for now; as currently written this is all about document validation.

benjie (changes requested):
Looking good, couple minor fixes :+1:

benjie:
I have:

- reverted the changes to section 3
- merged in the latest `main`
- merged in the changes proposed by @Shane32 in https://github.com/graphql/graphql-spec/pull/1098
- applied (heavy) editorial over the top of this

I think it's good to go now! :crossed_fingers:

spec/Section 5 -- Validation.md
```diff
@@ -135,6 +135,51 @@ extend type Dog {
...
+### All Operation Definitions
+
+#### Operation Type Existence
+
+**Formal Specification**
+
+- For each operation definition {operation} in the document:
+  - Let {rootOperationType} be the _root operation type_ in {schema}
+    corresponding to the kind of {operation}.
+  - {operationRootType} must exist.
```
Shane32, line 147:
```suggestion
  - {rootOperationType} must exist.
```

benjie: Oops; thanks!

Shane32 (approved):
Short and to the point.  I like it.  👍

Shane32 (approved):

yaacovCR (approved):

benjie:
@yaacovCR has an implementation of this here: https://github.com/graphql/graphql-js/pull/3601

benjie:
[Stated](https://youtu.be/6bW3Xcks5I4?t=1310) by @leebyron in Dec 2024 WG meeting:

> I will do an offline review [...] and I might just accept and merge it.

Conclusion: essentially this just needs a final editorial review, then it's good to merge.

spec/Section 5 -- Validation.md
```diff
@@ -135,6 +135,51 @@ extend type Dog {
...
+### All Operation Definitions
+
+#### Operation Type Existence
+
+**Formal Specification**
+
+- For each operation definition {operation} in the document:
+  - Let {rootOperationType} be the _root operation type_ in {schema}
+    corresponding to the kind of {operation}.
+  - {rootOperationType} must exist.
```
JoviDeCroock, line 147:
```suggestion
  - {rootOperationType} must be defined in the schema.
```

benjie: We have existing algorithms that follow this pattern, e.g.

https://github.com/graphql/graphql-spec/blob/b41339a0fd22fb3ca9a1a6110334a7094f1f7643/spec/Section%205%20--%20Validation.md#L664-L666

BoD (approved):

PR #957, opened 2022-06-03
Editorial: fixing a number of small issues

rivantsov:
Small issues in the spec: 
https://github.com/graphql/graphql-wg/discussions/969

this PR fixes a number of issues (20), not all of them, but those that are editorial and do not require discussions

PR #958, opened 2022-06-05
Leaf field selections clarification

benjie:
An alternative proposal to one of the changes in #957, this clarifies that we're talking about the "unwrapped type" or "underlying type" of the fields/selections.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -583,8 +583,8 @@ fragment conflictingDifferingResponses on Pet {
 
 **Explanatory Text**
 
-Field subselections are never allowed on leaf fields. A leaf field is any
-field with a scalar or enum result type.
+Field subselections are never allowed on leaf fields. A leaf field is any field
+with a scalar or enum result type.
 
 The following is valid.
 
@@ -637,7 +637,6 @@ query directQueryOnUnionWithoutSubFields {
 ```
 
 However the following example is valid since it includes a field subselection.
-
 
 ```graphql example
 query directQueryOnObjectWithSubFields {
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -605,8 +605,8 @@ fragment scalarSelectionsNotAllowedOnInt on Dog {
 ```
 
 Conversely the underlying type of leaf field selections of GraphQL operations
-must be scalar or enum. Leaf selections on fields whose underlying type is
-object, interface, or union are disallowed.
+must be scalar or enum. Leaf selections without subfields on fields whose
+underlying type is object, interface, or union are disallowed.
 
 Let's assume the following additions to the query root operation type of the
 schema:
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -583,8 +583,8 @@ fragment conflictingDifferingResponses on Pet {
 
 **Explanatory Text**
 
-Field selections on scalars or enums are never allowed, because they are the
-leaf nodes of any GraphQL operation.
+Scalars and enums are the underlying type of leaf nodes of any GraphQL
+operation. Field selections are never allowed on leaf nodes.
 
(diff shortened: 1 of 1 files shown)


spec/Section 5 -- Validation.md
```diff
@@ -604,9 +604,9 @@ fragment scalarSelectionsNotAllowedOnInt on Dog {
 }
 ```
 
-Conversely the leaf field selections of GraphQL operations must be of type
-scalar or enum. Leaf selections on objects, interfaces, and unions without
-subfields are disallowed.
+Conversely the underlying type of leaf field selections of GraphQL operations
+must be scalar or enum. Leaf selections on fields whose underlying type is
+object, interface, or union are disallowed.
```
benjie, line 609: "Leaf selections ... without subfields" seemed redundant to me - a leaf selection is one without subfields, right?

martinbonnin: Aren't leaf selections always fields? Making this whole section about "leaf fields" instead of "leaf selections" might clear up the language a bit?

benjie: Ah 🤦‍♂️  I shouldn't raise PRs before I've had breakfast. I've re-added "without subfields" for now, we could do a more significant rewrite via a later PR.

leebyron:
I think "result type" instead of 'underlying type" is a more correct term to use. I reframed the explanatory text to more directly mirror the algorithm.

@benjie let me know what you think

benjie:
Thanks for looking this over @leebyron! The term "result type" is only used once in the spec right now AFAICT and it's here. Imagine we have a selection set `{ foo }` which contains a selection `foo` of type `[[Int!]!]!`:

```graphql
type Query {
  foo: [[Int!]!]!
}

query Q { foo }
```

If you asked me what the "result type" of the selection `foo` was, I'd assume it was the type the `foo` resolver would need to return, that being `[[Int!]!]!`. Though the _unwrapped_ result type is a scalar, I'm uncomfortable saying that `[[Int!]!]!` is a scalar (not least because mathematically, `[[1, 2], [3, 4]]` is not a scalar).

We have this kind of "the type includes wrapped versions of the type" wording throughout the spec text:

> the leaves of this tree are typically GraphQL Scalar types (but may also be Enum types or null values).
> -- https://spec.graphql.org/draft/#sel-EAHXFCA_DCCJkwQ

> A GraphQL Input Object defines a set of input fields; the input fields are either scalars, enums, or other input objects.
> -- https://spec.graphql.org/draft/#sel-DAHhBDJAACLAy3B

We also have places where we are more explicit, for example:

> A field of an Object type may be a Scalar, Enum, another Object type, an Interface, or a Union. Additionally, it may be any wrapping type whose underlying base type is one of those five.
> -- https://spec.graphql.org/draft/#sel-FAHZpBABAB6Ft4T

In the algorithm text we tend to be much more precise, though thinking about it, maybe that's simply by virtue of always checking if it is a non-null or list type first, so by the time it hits "is an Object type" it's already unwrapped... 🤔 

I used the term "underlying type" influenced by:

> A wrapping type has an underlying named type, found by continually unwrapping the type until a named type is found.
> https://spec.graphql.org/draft/#sel-GAHVTJABqDB9GqE

Maybe we can solve this by defining "result type" to be this "underlying named type" (e.g. `Int`)? If we do, we must have a name for the actual type the selection uses though (e.g. `[[Int!]!]!`), and I fear calling that the "type" may lead to ambiguity, and "return type" would be confusing 😉

rivantsov:
agree with @benjie that 'return type' would be confusing for this particular use, 'unwrapped type' seem better

leebyron:
Cool, I agree that "unwrapped" is more accurate and more consistent with how we refer to this kind of type elsewhere.

My thought with "result" type was clarifying the type of value returned, as opposed to the type's involved with inputs or any other element of a field. Given the context that's probably unnecessary.

Thanks both of you for championing this clarification, it's much better

PR #960, opened 2022-06-10
RFC Allow recursion within ResolveAbstractType

yaacovCR:
This formalizes the proprosed feature within `graphql-js` whereby the internal method provided by JavaScript for runtime type resolution is allowed to return an intermediate interface.

See:
Issue: https://github.com/graphql/graphql-js/issues/3253
PR: https://github.com/graphql/graphql-js/pull/3599

@IvanGoncharov [suggested](https://github.com/graphql/graphql-js/pull/3599#pullrequestreview-1000172471) that this would require a spec change. Alternatively, perhaps the recursion [should be considered to be a feature of  the internal system](https://github.com/graphql/graphql-js/pull/3599#issuecomment-1150257261) itself, possibly limited to JavaScript-like implementations.

This PR provides some potential spec text, were a spec change to be considered necessary.

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -722,8 +722,13 @@ and output of {CoerceResult()} must not be {null}.
 
 When completing a field with an abstract return type, that is an Interface or
 Union return type, first the abstract type must be resolved to a relevant Object
-type. This determination is made by the internal system using whatever means
-appropriate.
+type. The internal system must provided a method for doing so using whatever
+means appropriate.
+
+Note: In cases where interfaces may implement interfaces, creating an abstract
+type hierarchy, the provided internal method may find it convenient on some
+systems to return an intermediate Interface type rather than the runtime Object
+type, requiring recursion, as detailed below.
 
 Note: A common method of determining the Object type for an {objectValue} in
 object-oriented environments, such as Java or C#, is to use the class name of
@@ -731,9 +736,13 @@ the {objectValue}.
 
 ResolveAbstractType(abstractType, objectValue):
 
-- Return the result of calling the internal method provided by the type system
-  for determining the Object type of {abstractType} given the value
+- Let {possibleRuntimeType} be the result of calling the internal method
+  provided by the type system, given Abstract type {abstractType} and value
   {objectValue}.
+- If {possibleRuntimeType} is an Object type:
+  - Return {possibleRuntimeType}.
+- If {possibleRuntimeType} is an Interface type:
+  - Return ResolveAbstractTypeImpl(possibleRuntimeType, objectValue).
 
 **Merging Selection Sets**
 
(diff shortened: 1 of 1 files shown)


spec/Section 6 -- Execution.md
```diff
@@ -722,18 +722,27 @@ and output of {CoerceResult()} must not be {null}.
...
 
-- Return the result of calling the internal method provided by the type system
-  for determining the Object type of {abstractType} given the value
+- Let {possibleRuntimeType} be the result of calling the internal method
+  provided by the type system, given Abstract type {abstractType} and value
   {objectValue}.
+- If {possibleRuntimeType} is an Object type:
+  - Return {possibleRuntimeType}.
+- If {possibleRuntimeType} is an Interface type:
+  - Return ResolveAbstractTypeImpl(possibleRuntimeType, objectValue).
```
benjie, line 745: Suggest simplification:

```suggestion
- If {possibleRuntimeType} is an Interface type:
  - Return ResolveAbstractTypeImpl(possibleRuntimeType, objectValue).
- Return {possibleRuntimeType}.
```

benjie (changes requested):

spec/Section 6 -- Execution.md
```diff
@@ -722,18 +722,26 @@ and output of {CoerceResult()} must not be {null}.
...
 the {objectValue}.
 
 ResolveAbstractType(abstractType, objectValue):
 
-- Return the result of calling the internal method provided by the type system
-  for determining the Object type of {abstractType} given the value
+- Let {possibleRuntimeType} be the result of calling the internal method
+  provided by the type system, given Abstract type {abstractType} and value
   {objectValue}.
+- If {possibleRuntimeType} is an Interface type:
```
leebyron, line 742: Following up on discussion - this probably needs an assertion that `{possibleRuntimeType}` (maybe `{resolvedType}`?) implements `{abstractType}`?

yaacovCR: Agreed!

yaacovCR: Hmmm. This PR allows the internal type system method to resolve to either an object or to a transitive interface. Everything should be fine as long as the internal type system method returns a type that implements the original abstract type, so an assertion is required.

However, the existing behavior also requires a similar assertion, as well as requiring an assertion that the internal method returned an object type. Perhaps to better clarify the proposed change, we should have a separate PR that adds the current assertions/behavior to the spec, and this PR can then modify/loosen those assertions/behavior as desired.

See: https://github.com/graphql/graphql-spec/pull/973

yaacovCR: @leebyron 

Hmmm. This PR allows the internal type system method to resolve to either an object or to a transitive interface. Everything should be fine as long as the internal type system method returns a type that implements the original abstract type, so an assertion is required.

However, the existing behavior also requires a similar assertion, as well as requiring an assertion that the internal method returned an object type. Perhaps to better clarify the proposed change, we should have a separate PR that adds the current assertions/behavior to the spec, and this PR can then modify/loosen those assertions/behavior as desired.

See: https://github.com/graphql/graphql-spec/pull/973

yaacovCR:
This proposal is currently in need of a new champion. I am no longer convinced that we need to provide it within the specification, as it is implementable in user land.

PR #961, opened 2022-06-11
Fix scripts to enable running on Windows

benjie:
Windows doesn't seem to recognise `''` for quoting CLI arguments, so we must use `""` instead (which unfortunately means we need to escape them with backslashes due to JSON 😞).


michaelstaib (approved):

leebyron:
ytho windows?

PR #962, opened 2022-06-14
Move punctuation outside quotation marks

spawnia:
As discussed in https://github.com/graphql/graphql-over-http/pull/175#discussion_r896551920 - @benjie 

Every variant of the MIT license does it this way, see https://fedoraproject.org/wiki/Licensing:MIT for reference.


spawnia:
Let this be the change we want to see in the grammar rules, they can be what we make it ✨

leebyron:
🤷

PR #966, opened 2022-06-21
Editorial: Error Terminology

leebyron:
Uses the definition syntax to set definitions for "field error" and "request error", and uses italic references in every formal normative location as well as the first reference in each prose.

This also clarifies when we previously used a plural "field errors" that we actually mean "a list of field error"

Include links to other sections as set up in #957

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -48,10 +48,10 @@ GetOperation(document, operationName):
 - If {operationName} is {null}:
   - If {document} contains exactly one operation.
     - Return the Operation contained in the {document}.
-  - Otherwise raise a request error requiring {operationName}.
+  - Otherwise raise a _request error_ requiring {operationName}.
 - Otherwise:
   - Let {operation} be the Operation named {operationName} in {document}.
-  - If {operation} was not found, raise a request error.
+  - If {operation} was not found, raise a _request error_.
   - Return {operation}.
 
 ### Validating Requests
@@ -75,8 +75,8 @@ result to avoid validating the same request again in the future.
 
 If the operation has defined any variables, then the values for those variables
 need to be coerced using the input coercion rules of variable's declared type.
-If a request error is encountered during input coercion of variable values, then
-the operation fails without execution.
+If a _request error_ is encountered during input coercion of variable values,
+then the operation fails without execution.
 
 CoerceVariableValues(schema, operation, variableValues):
 
@@ -95,14 +95,14 @@ CoerceVariableValues(schema, operation, variableValues):
     - Add an entry to {coercedValues} named {variableName} with the value
       {defaultValue}.
   - Otherwise if {variableType} is a Non-Nullable type, and either {hasValue} is
-    not {true} or {value} is {null}, raise a request error.
+    not {true} or {value} is {null}, raise a _request error_.
   - Otherwise if {hasValue} is true:
     - If {value} is {null}:
       - Add an entry to {coercedValues} named {variableName} with the value
         {null}.
     - Otherwise:
       - If {value} cannot be coerced according to the input coercion rules of
-        {variableType}, raise a request error.
+        {variableType}, raise a _request error_.
       - Let {coercedValue} be the result of coercing {value} according to the
         input coercion rules of {variableType}.
       - Add an entry to {coercedValues} named {variableName} with the value
@@ -134,7 +134,8 @@ ExecuteQuery(query, schema, variableValues, initialValue):
 - Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
   queryType, initialValue, variableValues)} _normally_ (allowing
   parallelization).
-- Let {errors} be any _field errors_ produced while executing the selection set.
+- Let {errors} be the list of all _field error_ raised while executing the
+  selection set.
 - Return an unordered map containing {data} and {errors}.
 
 ### Mutation
@@ -154,7 +155,8 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
 - Let {selectionSet} be the top level Selection Set in {mutation}.
 - Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
   mutationType, initialValue, variableValues)} _serially_.
-- Let {errors} be any _field errors_ produced while executing the selection set.
+- Let {errors} be the list of all _field error_ raised while executing the
+  selection set.
 - Return an unordered map containing {data} and {errors}.
 
 ### Subscription
@@ -256,7 +258,7 @@ CreateSourceEventStream(subscription, schema, variableValues, initialValue):
 - Let {selectionSet} be the top level Selection Set in {subscription}.
 - Let {groupedFieldSet} be the result of {CollectFields(subscriptionType,
   selectionSet, variableValues)}.
-- If {groupedFieldSet} does not have exactly one entry, raise a request error.
+- If {groupedFieldSet} does not have exactly one entry, raise a _request error_.
 - Let {fields} be the value of the first entry in {groupedFieldSet}.
 - Let {fieldName} be the name of the first entry in {fields}. Note: This value
   is unaffected if an alias is used.
@@ -302,7 +304,8 @@ ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
 - Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
   subscriptionType, initialValue, variableValues)} _normally_ (allowing
   parallelization).
-- Let {errors} be any _field errors_ produced while executing the selection set.
+- Let {errors} be the list of all _field error_ raised while executing the
+  selection set.
 - Return an unordered map containing {data} and {errors}.
 
 Note: The {ExecuteSubscriptionEvent()} algorithm is intentionally similar to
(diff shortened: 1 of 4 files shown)


spec/Section 6 -- Execution.md
```diff
@@ -769,13 +772,13 @@ MergeSelectionSets(fields):
...
+A _field error_ is an error raised from a particular field during value
+resolution or coercion. While these errors should be reported in the response,
+they are "handled" by producing a partial response.
 
-Note: This is distinct from ["request errors"](#sec-Errors.Request-errors) which
-are raised before execution begins. If a request error is encountered, execution
-does not begin and no data is returned in the response.
+Note: This is distinct from a _request error_ which is raised before execution
+begins. If a request error is encountered, execution does not begin and no data
+is returned in the response.
```
benjie, line 781:
```suggestion
Note: This is distinct from a _request error_ which is raised before execution
begins. If a _request error_ is encountered, execution does not begin and no data
is returned in the response.
```

I'm indifferent to this suggestion since it's already linked in the paragraph, feel free to dismiss

spec/Section 7 -- Response.md
```diff
@@ -4,8 +4,9 @@ When a GraphQL service receives a _request_, it must return a well-formed
 response. The service's response describes the result of executing the requested
 operation if successful, and describes any errors raised during the request.
 
-A response may contain both a partial response as well as any field errors in
-the case that a field error was raised on a field and was replaced with {null}.
+A response may contain both a partial response as well as a list of all _field
+error_ in the case that any field error was raised on a field and was replaced
```
benjie, line 8:
```suggestion
error_ in the case that any _field error_ was raised on a field and was replaced
```

I'm indifferent to this suggestion since it's already linked in the paragraph, feel free to dismiss

spec/Section 7 -- Response.md
```diff
@@ -60,33 +61,34 @@ response must not be empty. It must contain at least one error. The errors it
 contains should indicate why no data was able to be returned.
 
 If the `data` entry in the response is present (including if it is the value
-{null}), the `errors` entry in the response may contain any field errors that
-were raised during execution. If field errors were raised during execution, it
-should contain those errors.
+{null}), the `errors` entry in the response may contain a list of all _field
+error_ that were raised during execution. If any _field error_ was raised during
+execution, it should contain the `errors` entry.
```
benjie, line 66: Because we've gone from plural to singular, the "it" has become ambiguous (it could be referring to the "field error"); so let's clarify:

```suggestion
execution, the request should contain the `errors` entry.
```

spec/Section 7 -- Response.md
```diff
@@ -60,33 +61,34 @@ response must not be empty. It must contain at least one error. The errors it
...
 
 **Request errors**
 
-Request errors are raised before execution begins. This may occur due to a parse
-grammar or validation error in the requested document, an inability to determine
-which operation to execute, or invalid input values for variables.
+:: A _request error_ is an error raised during a _request_ but before execution
+begins. This may occur due to a parse grammar or validation error in the
+_Document_, an inability to determine which operation to execute, or invalid
+input values for variables.
```
benjie, line 73: Note `CreateSourceEventStream()` can also raise a request error:

> 5. If groupedFieldSet does not have exactly one entry, raise a request error.
>
> -- https://spec.graphql.org/draft/#CreateSourceEventStream()

There's also various places where there are `Assert:` that could raise an error (though if you've done validation they should not occur, except [when we don't handle that well](https://github.com/graphql/graphql-spec/pull/860)).

The phrase "before execution begins" is quite a fuzzy definition it turns out; here's a comment where I try and pick it apart a bit: https://github.com/graphql/graphql-js/pull/3640#issuecomment-1153711229

My current opinion is that "execution begins"

- for queries: when you enter [ExecuteQuery()](https://spec.graphql.org/draft/#ExecuteQuery())
- for mutations: when you enter [ExecuteMutation()](https://spec.graphql.org/draft/#ExecuteMutation())
- for subscriptions: when you enter [MapSourceToResponseEvent()](https://spec.graphql.org/draft/#MapSourceToResponseEvent())

spec/Section 7 -- Response.md
```diff
@@ -60,33 +61,34 @@ response must not be empty. It must contain at least one error. The errors it
...
 
 If a request error is raised, execution does not begin and the `data` entry in
 the response must not be present. The `errors` entry must include the error.
 
 **Field errors**
 
-Field errors are raised during execution from a particular field. This may occur
-due to an internal error during value resolution or failure to coerce the
-resulting value.
+:: A _field error_ is raised during the execution of a particular field. This
```
benjie, line 82: For consistency, either both "request error" and "field error" should have "is **an error** raised" or neither of them should. I think both is preferable:

```suggestion
:: A _field error_ is an error raised during the execution of a particular field. This
```

benjie (changes requested):
Generally great improvement, section 7 needs perhaps a little extra attention. Also, see related: https://github.com/graphql/graphql-spec/pull/967

spec/Section 7 -- Response.md
rivantsov: I believe the second sentence (If any field error was raised... it SHOULD (?) contain ...) is a repetition of the first sentence (the errors entry in the response MAY(?) contain a list...)  - basically saying 'errors' contains errors raised. 
At the same time, there is slight contradiction - MAY vs SHOULD

rivantsov: I suggest to remove the second sentence and change MAY -> SHOULD in the first.

leebyron:
Thanks for the detailed read! I'll incorporate

spec/Section 6 -- Execution.md
leebyron: I'll dismiss. My rule of thumb is to link every normative reference (i.e. in an algo) and to link the first or most prominent reference in a particular prose

spec/Section 7 -- Response.md
leebyron: Lots of confusing interchanging of must/should - let me try to reword this section further to clarify

spec/Section 7 -- Response.md
leebyron: Time boundary isn't the best way to describe this. Maybe prose about this being "typical" could be helpful, but really the difference is that field errors go alongside partial results where request errors do not

benjie: That's kind of circular reasoning though; for example if an error is raised from the `subscribe()` resolver of a field (i.e. in step 10 of [CreateSourceEventStream()](https://spec.graphql.org/draft/#CreateSourceEventStream()) via the [ResolveFieldEventStream()](https://spec.graphql.org/draft/#ResolveFieldEventStream()) algorithm) then it's not clear from the spec whether that should be a request error (since it prevents setting up the subscription, so the request cannot continue) or a field error (since it originated from a field). And thus, it's not clear if it should resolve to `{data: null, errors: [...]}` or just `{errors: [...]}` - and thus not clear if it's a partial result or not.

In my opinion, the boundary is effectively that it's a request error if it happens before the first `ExecuteSelectionSet()`; the boundaries I defined above give this rule of thumb explicit positions in the algorithm that are passed once only for each request.

spec/Section 7 -- Response.md
```diff
@@ -49,44 +50,47 @@ If an error was raised during the execution that prevented a valid response, the
...
+If the `data` entry in the response is not present, the `errors` entry must be
+present. It must contain at least one _request error_ indicating why no data was
+able to be returned.
 
 If the `data` entry in the response is present (including if it is the value
-{null}), the `errors` entry in the response may contain any field errors that
-were raised during execution. If field errors were raised during execution, it
-should contain those errors.
+{null}), the `errors` entry must also be present if a _field error_ was raised
+during execution.
```
benjie, line 67: context:
```
If the `data` entry in the response is present (including if it is the value
```


If we use "if and only if" then it would also enforce that errors must not be present if there were no field errors; I think this adds clarity.


```suggestion
{null}), the `errors` entry must be present if and only if a _field error_ was
raised during execution.
```

We could also change "a" to "one or more":


```suggestion
{null}), the `errors` entry must be present if and only if one or more _field
error_ were raised during execution.
```

leebyron:
Merged your suggestions @benjie

benjie (approved):
Really good improvements, thanks @leebyron :raised_hands:

spec/Section 6 -- Execution.md
```diff
@@ -154,7 +155,8 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
 - Let {selectionSet} be the top level Selection Set in {mutation}.
 - Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
   mutationType, initialValue, variableValues)} _serially_.
-- Let {errors} be any _field errors_ produced while executing the selection set.
+- Let {errors} be the list of all _field error_ raised while executing the
```
yaacovCR, line 158: field errors plural?

spec/Section 6 -- Execution.md
```diff
@@ -302,7 +304,8 @@ ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
 - Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
   subscriptionType, initialValue, variableValues)} _normally_ (allowing
   parallelization).
-- Let {errors} be any _field errors_ produced while executing the selection set.
+- Let {errors} be the list of all _field error_ raised while executing the
```
yaacovCR, line 307: Plural?

spec/Section 7 -- Response.md
```diff
@@ -4,8 +4,9 @@ When a GraphQL service receives a _request_, it must return a well-formed
 response. The service's response describes the result of executing the requested
 operation if successful, and describes any errors raised during the request.
 
-A response may contain both a partial response as well as any field errors in
-the case that a field error was raised on a field and was replaced with {null}.
+A response may contain both a partial response as well as a list of all _field
```
yaacovCR, line 7: Plural?

spec/Section 6 -- Execution.md
leebyron: For this to be a glossary referenced term it need to remain in the singular, hence "list of field error" vs "field errors"

spec/Section 6 -- Execution.md
benjie: That won’t work with the definitions feature of spec-md

PR #967, opened 2022-06-22
Fix normative requirement around `errors` property

benjie:
Whilst reviewing #966 I spotted this; I don't believe this is a breaking change since it can be derived from:

> The `errors` entry in the response is a **non-empty** list of errors, where each
error is a map.

So if no errors occur, then the "errors" key **must not** (rather than "should not") be present, since it's forbidden to be an empty list, it contains errors, and no errors occurred.


benjie:
Changes were incorporated into #966

PR #970, opened 2022-06-27
Editorial: Type kinds

leebyron:
This provides definitions at the introduction of "Type System" for _leaf type_, _composite type_, and _abstract type_. Then, these definitions are used throughout the spec.

---

Inspired by #957 which had included clarifications around "leaf" types and fields. This uses definition syntax to formally define the terms so they can be used.

This ended up being a pretty big change, so I'd love a review

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -264,32 +264,43 @@ TypeDefinition :
 - EnumTypeDefinition
 - InputObjectTypeDefinition
 
-The fundamental unit of any GraphQL Schema is the type. There are six kinds of
+The fundamental unit of any GraphQL schema is the type. There are six kinds of
 named type definitions in GraphQL, and two wrapping types.
 
-The most basic type is a `Scalar`. A scalar represents a primitive value, like a
-string or an integer. Oftentimes, the possible responses for a scalar field are
-enumerable. GraphQL offers an `Enum` type in those cases, where the type
-specifies the space of valid responses.
+:: A _leaf type_ is a kind of type representing a primitive value which cannot
+be further selected and thus form the leaves in a response tree. GraphQL
+provides two kinds of leaf types: `Scalar` and `Enum`.
 
-Scalars and Enums form the leaves in response trees; the intermediate levels are
-`Object` types, which define a set of fields, where each field is another type
-in the system, allowing the definition of arbitrary type hierarchies.
+A `Scalar` represents a primitive scalar value, such as a string or number.
+Oftentimes, the possible responses for a scalar field are enumerable. GraphQL
+offers an `Enum` type in those cases, where the type specifies the set of valid
+responses.
 
-GraphQL supports two abstract types: interfaces and unions.
+:: A _composite type_ is a type composed of other types via a set of named
+fields. Each field may provide a _leaf type_ or another composite type (or
+wrapped types of either), allowing for the definition of arbitrary type
+hierarchies.
 
-An `Interface` defines a list of fields; `Object` types and other Interface
-types which implement this Interface are guaranteed to implement those fields.
-Whenever a field claims it will return an Interface type, it will return a valid
-implementing Object type during execution.
+An `Object` type is a _composite type_ representing composite values selectable
+within a GraphQL operation. They provide the intermediate levels of a schema,
+allowing GraphQL to describe an interconnected graph of information.
 
-A `Union` defines a list of possible types; similar to interfaces, whenever the
-type system claims a union will be returned, one of the possible types will be
-returned.
+:: An _abstract type_ allows a GraphQL schema to introduce polymorphism; where a
+field may provide one of many possible types at runtime. GraphQL provides two
+kinds of abstract types: `Interface` and `Union`.
+
+An `Interface` defines a list of fields; `Object` types and other `Interface`
+types which implement this `Interface` are guaranteed to implement those fields.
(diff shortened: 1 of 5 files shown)


spec/Section 3 -- Type System.md
```diff
@@ -264,32 +264,43 @@ TypeDefinition :
...
-The fundamental unit of any GraphQL Schema is the type. There are six kinds of
+The fundamental unit of any GraphQL schema is the type. There are six kinds of
 named type definitions in GraphQL, and two wrapping types.
 
-The most basic type is a `Scalar`. A scalar represents a primitive value, like a
-string or an integer. Oftentimes, the possible responses for a scalar field are
-enumerable. GraphQL offers an `Enum` type in those cases, where the type
-specifies the space of valid responses.
+:: A _leaf type_ is a kind of type representing a primitive value which cannot
+be further selected and thus form the leaves in a response tree. GraphQL
```
rivantsov, line 271: something does not match; if subject is 'kind' (singular) then the verb should be 'forms' (singular)

spec/Section 3 -- Type System.md
```diff
@@ -264,32 +264,43 @@ TypeDefinition :
...
-Scalars and Enums form the leaves in response trees; the intermediate levels are
-`Object` types, which define a set of fields, where each field is another type
-in the system, allowing the definition of arbitrary type hierarchies.
+A `Scalar` represents a primitive scalar value, such as a string or number.
+Oftentimes, the possible responses for a scalar field are enumerable. GraphQL
+offers an `Enum` type in those cases, where the type specifies the set of valid
+responses.
 
-GraphQL supports two abstract types: interfaces and unions.
+:: A _composite type_ is a type composed of other types via a set of named
```
rivantsov, line 279: need consistency in capitalizing - if 'Leaf type' then 'Composite type'

spec/Section 3 -- Type System.md
```diff
@@ -264,32 +264,43 @@ TypeDefinition :
...
-Whenever a field claims it will return an Interface type, it will return a valid
-implementing Object type during execution.
+An `Object` type is a _composite type_ representing composite values selectable
+within a GraphQL operation. They provide the intermediate levels of a schema,
+allowing GraphQL to describe an interconnected graph of information.
 
-A `Union` defines a list of possible types; similar to interfaces, whenever the
-type system claims a union will be returned, one of the possible types will be
-returned.
+:: An _abstract type_ allows a GraphQL schema to introduce polymorphism; where a
```
rivantsov, line 288: I think it should be comma after polymorphism, not semicolon.

spec/Section 3 -- Type System.md
```diff
@@ -635,14 +646,14 @@ FieldDefinition : Description? Name ArgumentsDefinition? : Type
...
-GraphQL Objects represent a list of named fields, each of which yield a value of
-a specific type. Object values should be serialized as ordered maps, where the
-selected field names (or aliases) are the keys and the result of evaluating the
-field is the value, ordered by the order in which they appear in the selection
-set.
+GraphQL Objects are a _composite type_ representing a list of named fields, each
+of which yield a value of a specific type. Object values should be serialized as
+ordered maps, where the selected field names (or aliases) are the keys and the
+result of evaluating the field is the value, ordered by the order in which they
+appear in the selection set.
```
rivantsov, line 656: we might mention here that field names might repeat in this output map, as a result of field merging

spec/Section 3 -- Type System.md
```diff
@@ -1304,17 +1320,18 @@ UnionMemberTypes :
 - UnionMemberTypes | NamedType
 - = `|`? NamedType
 
-GraphQL Unions represent an object that could be one of a list of GraphQL Object
-types, but provides for no guaranteed fields between those types. They also
-differ from interfaces in that Object types declare what interfaces they
-implement, but are not aware of what unions contain them.
+GraphQL Unions are an _abstract type_ representing one of a list of GraphQL
+Object possible types, but provides for no guaranteed fields between those
```
rivantsov, line 1324: provides -> provide ? since Unions (plural)

spec/Section 5 -- Validation.md
```diff
@@ -446,7 +446,7 @@ SameResponseShape(fieldA, fieldB):
 - If {typeA} or {typeB} is Scalar or Enum:
   - If {typeA} and {typeB} are the same type return true, otherwise return
     false.
-- Assert: {typeA} and {typeB} are both composite types.
+- Assert: {typeA} and {typeB} are both Object, Interface or Union.
```
benjie, line 449: This could be considered as `(a is object and b is object) or (a is interface and b is interface) or (a is union and b is union)`. I think it's clearer without this change; alternatively use "each":

```suggestion
- Assert: each of {typeA} and {typeB} are an Object, Interface or Union.
```

benjie (changes requested):
Looking good; great readability/cross-referencing improvements I think 👍

spec/Section 3 -- Type System.md
rivantsov: General suggestion 
I think 'kind of type' is confusing, since we have already TYPE_KIND, different kind. What we introduce here are groups of types, sort of, purely for easier references in the spec; these groups or categories do not appear anywhere at runtime or in graphql artifacts - purely for shorter and simpler spec references.

So instead of 
  A Leaf type is a kind of type representing ...  

we can say simply 
  Leaf types are types representing

  without even explicitly calling it a group or category

PR #973, opened 2022-07-28
Clarify current ResolveAbstractType algorithm

yaacovCR:
add assertions from current behavior of reference implementation


yaacovCR:
This PR attempts to clarify the current behavior.

It can be seen as a prelude to #960 (which loosens the return type of the internal type system method to allow returning a transitive interface) -- but this PR might also be a helpful addition on its own.

spec/Section 6 -- Execution.md
```diff
@@ -734,9 +734,22 @@ the {objectValue}.
...
-- Return the result of calling the internal method provided by the type system
-  for determining the Object type of {abstractType} given the value
-  {objectValue}.
+- Let {resolvedType} be the result of calling the internal method provided by
+  the type system for determining the Object type of {abstractType} given the
+  value {objectValue}.
+- Assert {resolvedType} is an Object type.
+- If {IsSubType(abstractType, resolvedType)}:
+  - Return {resolvedType}.
+- Otherwise, raise a field error.
```
benjie, line 743: This should also be an assert, I think. "determining the Object type of {abstractType} given the value {objectValue}" already specifies it's an object type, and that it's a valid object type within {abstractType}; so the asserts should just underline this.

```suggestion
- Assert {IsSubType(abstractType, resolvedType)}.
- Return {resolvedType}.
```

benjie: I'm not sure "sub type" is the right terminology here; an object type isn't really a "sub type" of a union, it's a member. I'm not sure what the right terminology is though, I've had a quick scan through the spec and couldn't find a suitable term. I guess it's okay to say an object type that implements an interface is effectively a member of the interface, so let's use member for now until a better term appears:

```suggestion
- Assert {IsAbstractMemberType(abstractType, resolvedType)}.
- Return {resolvedType}.
```

spec/Section 6 -- Execution.md
```diff
@@ -734,9 +734,22 @@ the {objectValue}.
...
+
+IsSubType(superType, maybeSubType):
+
+- If {abstractType} is an Interface type.
+  - If {resolvedType} implements {abstractType}, return {true}.
+  - Otherwise, return {false}.
+- If {abstractType} is a Union type.
+  - If {resolvedType} is a member type of {abstractType}, return {true}.
+  - Otherwise, return {false}.
```
benjie, line 752: `superType` isn't used, and I think `abstractType` is clearer and more consistent with the calling method. `maybeSubType` is problematic for me, the term "maybe" often implies nullability, and "sub" doesn't feel right as mentioned above. How about `otherType`?

```suggestion
IsAbstractMemberType(abstractType, otherType):

- Assert: {abstractType} is an Interface or Union type.
- If {abstractType} is an Interface type.
  - If {otherType} implements {abstractType}, return {true}.
  - Otherwise, return {false}.
- If {abstractType} is a Union type.
  - If {otherType} is a member type of {abstractType}, return {true}.
  - Otherwise, return {false}.
```

benjie (changes requested):
Some bike-shedding suggestions :sweat_smile:

spec/Section 6 -- Execution.md
yaacovCR: corrected as [c15550f](https://github.com/graphql/graphql-spec/pull/973/commits/c15550f287539cf5b7009ca389e8b7a46495ede1)

in terms of the language of "sub-type", it is used elsewhere, e.g. within [IsValidImplementation](https://spec.graphql.org/draft/#IsValidImplementation()):

> field must return a type which is equal to or a sub-type of (covariant) the return type of implementedField field’s return type

yaacovCR: In fact, I think if #960 is adopted, {objectType} could actually be an Interface type, and then a portion of [IsValidImplementationFieldType](https://spec.graphql.org/draft/#IsValidImplementationFieldType()) could be replaced by IsSubType => as the difference there is that object types or interface types are currently inspected as to whether they are subtypes, and we are just inspecting object types. Maybe a different PR should abstract that out anyway, and even though the algorithm here only inspects object types, we could copy that behavior?

spec/Section 6 -- Execution.md
```diff
@@ -734,9 +734,21 @@ the {objectValue}.
...
+- Return {resolvedType}.
+
+IsSubType(abstractType, objectType):
+
+- If {abstractType} is an Interface type.
+  - If {objectType} implements {abstractType}, return {true}.
+  - Otherwise, return {false}.
+- If {abstractType} is a Union type.
+  - If {objectType} is a member type of {abstractType}, return {true}.
+  - Otherwise, return {false}.
```
benjie, line 751: Based on the [discussion here](https://github.com/graphql/graphql-spec/pull/973#discussion_r932110074) and careful reading of IsValidImplementationFieldType, I believe we can just use that existing algorithm:

```suggestion
- Assert {resolvedType} is not the same type as {abstractType}.
- Assert {IsValidImplementationFieldType(resolvedType, abstractType)}.
- Return {resolvedType}.
```

It may make sense to extract the relevant parts into a new algorithm as @yaacovCR suggests though, as the name of the algorithm isn't necessarily the most sensible name for this usage.

benjie (changes requested):

spec/Section 6 -- Execution.md
benjie: Actually I'm going to dismiss this; the proposed implementation is clearer.

spec/Section 6 -- Execution.md
yaacovCR: See #977

leebyron:
Merged the dependent editorial change so you can use that change here.

---

This feels like a real ambiguity in the spec that's getting resolved, nice catch! Can we confirm that these assertions already align to existing GraphQL.js behavior?

yaacovCR:
@leebyron 

This change is meant to capture existing behavior in graphql-js:

see fn ensureValidRuntimeType @
https://github.com/graphql/graphql-js/blob/a24a9f35b876bdd0d5050eca34d3020bd0db9a29/src/execution/execute.ts#L1234

yaacovCR:
@graphql/tsc  ready for review

yaacovCR:
pinging @graphql/tsc on this

benjie (approved):
Looks good to me! :raised_hands:

PR #977, opened 2022-08-02
separate out IsSubType from IsValidImplementationFieldType

yaacovCR:
Motivation: IsSubType could be re-used within ResolveAbstractType when its behavior is clarified. See: #973.


benjie (approved):

leebyron:
Nice!

Confirming that this aligns to the existing https://github.com/graphql/graphql-js/blob/main/src/type/schema.ts#L311 in the ref impl.

This `IsValidImplementationFieldType()` is in large part directly mapping to https://github.com/graphql/graphql-js/blob/a24a9f35b876bdd0d5050eca34d3020bd0db9a29/src/utilities/typeComparators.ts#L38 which maps well to the change you proposed

PR #979, opened 2022-08-03
field merging - field TYPES must not differ

rivantsov:
Minor correction, Editorial.

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -556,7 +556,7 @@ fragment safeDifferingArgs on Pet {
 ```
 
 However, the field responses must be shapes which can be merged. For example,
-scalar values must not differ. In this example, `someValue` might be a `String`
+field types must not differ. In this example, `someValue` might be a `String`
 or an `Int`:
 
 ```graphql counter-example


yaacovCR (approved):

benjie (changes requested):
This is not true; the field types do not have to be the same, they just have to be compatible. Please read http://spec.graphql.org/draft/#FieldsInSetCanMerge() carefully and consider fields with polymorphic types (unions and interfaces). The text as currently stated in the spec is, however, correct.

rivantsov:
> The text as currently stated in the spec is, however, correct.

so, "scalar values must not differ" - it is correct? which values? we are at early validation stage, we do not have any values yet. I thought it should say something about types; ok, maybe not require the exact same types, then pls help to formulate the proper short phrase describing 'type' requirements

benjie:
You could change the word "values" to "types" I suppose.

spec/Section 5 -- Validation.md
```diff
@@ -556,7 +556,7 @@ fragment safeDifferingArgs on Pet {
 ```
 
 However, the field responses must be shapes which can be merged. For example,
-scalar values must not differ. In this example, `someValue` might be a `String`
+field types must not differ. In this example, `someValue` might be a `String`
```
benjie, line 559: Previously suggested here: https://github.com/graphql/graphql-spec/pull/957#discussion_r889675186

```suggestion
leaf types must not differ. In this example, `someValue` might be a `String`
```

benjie (changes requested):

spec/Section 5 -- Validation.md
rivantsov: 'leaf types' is just a subgroup of types, not clear, related to fields themselves. How about: 

For example, for fields of leaf types, the fields' types must not differ.

benjie (approved):

mjmahone (approved):

PR #983, opened 2022-08-03
Remove "subscriptions is a significant change" sentence

rivantsov:
Explanation, current version: 

> Supporting subscriptions is a significant change for any GraphQL service.

who is changing what? why we assume certain order of building GraphQL service - queries mutations first, then subscriptions. what if I built all in parallel.

I think 'challenge' makes more sense in the context of the section.


michaelstaib (approved):

benjie:
From Dec WG: "Supporting subscriptions requires a different approach to queries and mutations." -- @mjmahone

rivantsov:
Matt's suggestion: 
> Supporting subscriptions requires a different approach to queries and mutations.

I am still not quite happy with this. Is it "to queries and mutations" or "COMPARED to queries and mutations"? 
A few things to point out I think that would help: 
1. This is an introductory sentence to a new topic. It does not have to be much meaningful or say something specific about subscriptions. We actually struggle against the unintended "meaning" we bring in - that it is a CHANGE, or a CHALLENGE (beware it is difficult) and so on. We can just say meaningless thing like "This chapter is about another feature - Subscriptions"
2. If we do want to add something "meat" - subscriptions are different. We can say that they use different interaction model, not classic HTTP request response. Client requests, server responds. We have to go outside HTTP, and use WebSockets.

spec/Section 6 -- Execution.md
```diff
@@ -224,9 +224,10 @@ must receive no more events from that event stream.
 
 **Supporting Subscriptions at Scale**
 
-Supporting subscriptions is a significant change for any GraphQL service. Query
-and mutation operations are stateless, allowing scaling via cloning of GraphQL
-service instances. Subscriptions, by contrast, are stateful and require
+Supporting subscriptions at scale for a GraphQL service requires taking into
+consideration the need to preserve state on the server. 
+Query and mutation operations are stateless, allowing scaling via cloning 
+of GraphQL service instances. Subscriptions, by contrast, are stateful and require
```
benjie, line 230: This seems redundant ("subscriptions [...] need to preserve state on the server" vs "Subscriptions [...] are stateful"); perhaps we can just remove the first sentence entirely:

```suggestion
Query and mutation operations are stateless, allowing scaling via cloning 
of GraphQL service instances. Subscriptions, by contrast, are stateful and require
```

benjie (changes requested):

spec/Section 6 -- Execution.md
rivantsov: agree, let's just remove it

rivantsov:
@benjie , can you pls help with this 'code style/pretier' failure, have no idea what to do about it

benjie:
If you have node installed, in the root of the repo run `npm install` and then `npm run format`

benjie (approved):
This looks good to me. @graphql/tsc can I get another +1?

Keweiqu (approved):

benjie:
Thanks @Keweiqu and @rivantsov 🙌

PR #987, opened 2022-08-04
Fix ambiguity around when schema definition may be omitted

benjie:
@rivantsov Pointed out in #978 that there's some ambiguity around when the `schema` keyword can be omitted from the SDL. Upon careful reading I've noticed that there is additional ambiguity around this topic.

> While any type can be the root operation type for a GraphQL operation, the type
> system definition language can omit the schema definition when the {`query`},
> {`mutation`}, and {`subscription`} root types are named {"Query"}, {"Mutation"},
> and {"Subscription"} respectively.

This seems to imply that all the root types are required in order to omit the schema definition. I've modified the text to indicate that the names only need to match for the root types that are actually present.

> Likewise, when representing a GraphQL schema using the type system definition
> language, a schema definition should be omitted if it only uses the default root
> operation type names.

Imagine we're doing biological research, tracking mutations in a virus. We might have a schema like:

```graphql
type Query {
  viruses: [Virus!]
}
type Virus {
  name: String!
  knownMutations: [Mutation!]!
}
type Mutation {
  name: String!
  geneSequence: String!
}
schema {
  query: Query
}
```

In this case we must not omit the schema definition when representing the schema using the SDL, because doing so would make it seem that the Mutation type was the root type for the mutation operation, when in fact the schema does not support a mutation operation.

I've clarified the wording to deal with this possibility too.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -122,7 +122,7 @@ RootOperationTypeDefinition : OperationType : NamedType
 
 A GraphQL service's collective type system capabilities are referred to as that
 service's "schema". A schema is defined in terms of the types and directives it
-supports as well as the root operation types for each kind of operation: query,
+supports as well as the _root operation type_ for each kind of operation: query,
 mutation, and subscription; this determines the place in the type system where
 those operations begin.
 
@@ -141,24 +141,24 @@ introspection system.
 
 ### Root Operation Types
 
-A schema defines the initial root operation type for each kind of operation it
-supports: query, mutation, and subscription; this determines the place in the
+:: A schema defines the initial _root operation type_ for each kind of operation
+it supports: query, mutation, and subscription; this determines the place in the
 type system where those operations begin.
 
-The {`query`} root operation type must be provided and must be an Object type.
+The {`query`} _root operation type_ must be provided and must be an Object type.
 
-The {`mutation`} root operation type is optional; if it is not provided, the
+The {`mutation`} _root operation type_ is optional; if it is not provided, the
 service does not support mutations. If it is provided, it must be an Object
 type.
 
-Similarly, the {`subscription`} root operation type is also optional; if it is
+Similarly, the {`subscription`} _root operation type_ is also optional; if it is
 not provided, the service does not support subscriptions. If it is provided, it
 must be an Object type.
 
 The {`query`}, {`mutation`}, and {`subscription`} root types must all be
 different types if provided.
 
-The fields on the {`query`} root operation type indicate what fields are
+The fields on the {`query`} _root operation type_ indicate what fields are
 available at the top level of a GraphQL query operation.
 
 For example, this example operation:
@@ -169,7 +169,8 @@ query {
 }
 ```
 
-is only valid when the {`query`} root operation type has a field named "myName":
+is only valid when the {`query`} _root operation type_ has a field named
+"myName":
 
 ```graphql example
 type Query {
@@ -177,8 +178,8 @@ type Query {
 }
 ```
 
-Similarly, the following mutation is only valid if the {`mutation`} root
-operation type has a field named "setName".
+Similarly, the following mutation is only valid if the {`mutation`} _root
+operation type_ has a field named "setName".
 
 ```graphql example
 mutation {
@@ -191,8 +192,8 @@ mutation {
 When using the type system definition language, a document must include at most
 one {`schema`} definition.
 
-In this example, a GraphQL schema is defined with both query and mutation root
-operation types:
+In this example, a GraphQL schema is defined with both a query and mutation
+_root operation type_:
 
 ```graphql example
 schema {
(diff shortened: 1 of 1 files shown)


benjie:
I've submitted a larger edit that I think improves accuracy and makes reading (and referencing) easier by using a definition.

spec/Section 3 -- Type System.md
```diff
@@ -211,14 +211,19 @@ type MyMutationRootType {
...
+respectively.
+
+The type system definition language can omit the schema definition when each
+root type present uses its respective _default root operation type name_ and no
+other type uses a default root operation type name.
 
 Likewise, when representing a GraphQL schema using the type system definition
-language, a schema definition should be omitted if it only uses the default root
-operation type names.
+language, a schema definition should be omitted if each root type present uses
```
yaacovCR, line 224: Given a schema with the mutation root type with name CustomMutationRootType, but query root type query, presumably when printing the schema (converting to string), the schema definition should only include the entry for overriding the default mutation type name and not respecify the default query root type name.

Assuming that is true, perhaps we need a sentence within this expanded section that says so explicitly.

rivantsov: > The type system definition language can omit ... 

That does not seem right. The 'language' omits something? The SDL file maybe?

benjie: @yaacovCR 

> Given a schema with the mutation root type with name CustomMutationRootType, but query root type query, presumably when printing the schema (converting to string), the schema definition should only include the entry for overriding the default mutation type name and not respecify the default query root type name.

No; the schema definition is either omitted or included, it's never included partially. Otherwise my Virus example above would be construed to support mutation operations (it does not).

benjie: @rivantsov That text is unchanged from the current spec, that is a topic for a separate PR.

yaacovCR: > No; the schema definition is either omitted or included, it's never included partially. Otherwise my Virus example above would be construed to support mutation operations (it does not).

@benjie i guess what I am asking is whether the schema definition should be only partially included __when it can __ — ie my example rather than yours.

I imagine this is in line with the idea of skipping it entirely when possible. I am not sure this is desirable, but it would work.

benjie: I don't think that's desirable, and either way is outside the scope of this PR.

benjie:
@leebyron "default root type name" read better in context than "default operation type name" so I've gone with that. If you approve, I think this is ready to merge 👍

benjie:
@graphql/tsc I'm going to add this to Thursday's WG, a review/merge before then would make the discussion shorter 😉

mjmahone (approved):

leebyron:
Marking this as a Proposal since this is a meaningful change. Two changes here:

1. The clarity around default type names and whether the schema definition needs to exist is great, that's an editorial change.
2. The insight that we must not omit the schema definition if a default type name exists in a non-root type is a new change and we need to ensure it's built into the reference impl.

The test case we need to see is your description above: when putting SDL in to in-memory, then back to SDL, does a non-root type named "Mutation" end up in the right place?

benjie:
@graphql/implementers Please could you add the following to your test suites, and report back if there are any issues:

1. Create a schema from the following SDL (or, if you don't support SDL, construct an equivalent schema)
2. Print that schema
3. Assert that the result matches the schema, namely that the schema has a query operation named `Query` but **does not** have a mutation operation (the `Mutation` type being just a regular type).

```graphql
type Query {
  viruses: [Virus!]
}

type Virus {
  name: String!
  knownMutations: [Mutation!]!
}

type Mutation {
  name: String!
  geneSequence: String!
}

schema {
  query: Query
}
```

I've raised a PR against GraphQL.js that you can use for inspiration: https://github.com/graphql/graphql-js/pull/3839

spawnia:
@benjie I have added the described test case in https://github.com/webonyx/graphql-php/pull/1303.

Constructing the schema from your example SDL worked fine, it did correctly not support mutations. I had to make the same fix as you did in `graphql-js` to correct the schema printer output.

leebyron:
Necessary code changes looks great, key decisions look correct, good feedback from implementers - this is now RFC 2

leebyron:
Would love feedback on my edits @benjie 

* I thought it would be more clear if we made _root operation type_ a definition as well to see how these two relate.
* I made some minor copy edits, please let me know if this improved clarity by your eye
* I added a simplified version of the test case above

spec/Section 3 -- Type System.md
```diff
@@ -234,6 +235,31 @@ type Query {
 }
 ```
 
+While this example describes a valid GraphQL schema which happens to contain a
+type called {"Mutation"} which is not the {`mutation`} _root operation type_.
```
benjie, line 239: This reads awkwardly, I think simply removing `While` would fix it.

benjie:
@leebyron Looks good to me! While we're at it... do you think we should rename "root operation type" to simply "root type"? That would fit better with "default root type name", and I don't think it loses clarity.

leebyron:
> While we're at it... 

Yes this would be clarifying but to keep this one from expanding further I'll do a follow up

spec/Section 3 -- Type System.md
leebyron: reworded

benjie: Much better; spot on 👍

PR #994, opened 2022-10-17
Possible reference mistake in subscription execution

Yogu:
I'm a bit confused about MapSourceToResponseEvent in the execution specification for subscriptions. It explains how to map a `sourceStream` to a `responseStream`. Instead of specifying the behavior in case `sourceStream` completes, it talks about what happens when `responseStream` completes, even though there is no way how `responseStream` could complete on its own.

Am I right that the specification part just confused `sourceStream` and `responseStream` here?

Also, why is the operation called `MapSourceToResponseEvent` when it maps *event streams*? Shouldn't it be called `MapSourceStreamToResponseEventStream`?

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -294,7 +294,7 @@ MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
   - Let {response} be the result of running
     {ExecuteSubscriptionEvent(subscription, schema, variableValues, event)}.
   - Yield an event containing {response}.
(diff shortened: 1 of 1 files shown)


benjie:
There's no EasyCLA on this PR, so I'm going to close and open it again. I agree with this change.

benjie:
Hi @Yogu thanks for this fix. Just to summarise the discussion last night: this is indeed the right fix, and the other route of closing the streams (when the user unsubscribes, also (optionally) close the source stream) is covered in the prose above the `Unsubscribe` algorithm.

PR #997, opened 2022-10-20
Expand @deprecated to Objects

fotoetienne:
# Problem

Take as a motivating example:

```graphql
type Query {
  animals: [Animal]
}

interface Animal {
  name: String
}

type Dog implements Animal {
  name: String
}

type Baiji implements Animal {
  name: String
}
```

The Baiji type corresponds to an Animal that is no longer known to exist, and the server will no longer return this type.
We would like to delete the code for this type and eventually remove from the schema, but first clients must remove all references of this type from their queries. Currently, there is no good way to indicate to clients that they should no longer spread on this type in their queries.

# Solution
Allow @deprecated on objects. Marking as deprecated indicates to clients that this type will no longer be returned from the server. This can indicate to client build tooling that references to this object should be removed from client queries.

```graphql
type Baiji implements Animal @deprecated {
  name: String
}
```

## Alternative Solutions
The most compelling use-case for deprecating types is when they are union members or interface implementations. A potential alternative would be to instead deprecate the membership/implementation instead of the type itself. The main challenge with this approach is with syntax, since it unclear how one would unambiguously annotate an interface implementation using the current `@deprecated` directive. Some possible alternatives:

*New directive location - @deprecated on UNION_MEMBER*
```graphql
union Animal = Dog | Cat | Baiji @deprecated
```

*New directive @deprecatedMembers on UNION*
```graphql
union Animal @deprecatedMembers(members: ["Baiji"]) = Dog | Cat | Baiji
```

*New directive @deprecatedImplementations on OBJECT*
```graphql
type Baiji implements Node & Animal @deprecatedImplementations(implementations: ["Animal"]) {
  id: ID!
  name: String
}
```

*New directive @deprecatedImplementations on INTERFACE*
```graphql
interface Animal @deprecatedImplementations(implementations: ["Baiji"]) {
  id: ID!
  name: String
}
```

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -2098,9 +2098,10 @@ type ExampleType {
 To deprecate a required argument or input field, it must first be made optional
 by either changing the type to nullable or adding a default value.
 
-The `@deprecated` directive is useful for object types that are included in unions or
-interfaces. Deprecating the type indicates to clients that the server will no longer be returning
-this type and clients should remove references to this type in their queries.
+The `@deprecated` directive is useful for object types that are included in
+unions or interfaces. Deprecating the type indicates to clients that the server
+will no longer be returning this type and clients should remove references to
+this type in their queries.
 
 ```graphql example
 type Query {
@@ -2114,7 +2115,8 @@ type DeprecatedType @deprecated {
 }
 ```
 
-The `@deprecated` directive should not appear on object types that are referenced by non-deprecated fields.
+The `@deprecated` directive should not appear on object types that are
+referenced by non-deprecated fields.
 
 ```graphql counter-example
 type Query {
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -2057,7 +2057,7 @@ condition is false.
 ```graphql
 directive @deprecated(
   reason: String = "No longer supported"
-) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
+) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE | OBJECT
 ```
 
 The `@deprecated` _built-in directive_ is used within the type system definition
@@ -2097,6 +2097,34 @@ type ExampleType {
 
 To deprecate a required argument or input field, it must first be made optional
 by either changing the type to nullable or adding a default value.
+
+The `@deprecated` directive is useful for object types that are included in unions or
+interfaces. Deprecating the type indicates to clients that the server will no longer be returning
+this type and clients should remove references to this type in their queries.
+
+```graphql example
+type Query {
+  returnsUnion: MyUnion
+}
+
+union MyUnion = ExampleType | DeprecatedType
+
+type DeprecatedType @deprecated {
+  id: ID
+}
+```
+
+The `@deprecated` directive should not appear on object types that are referenced by non-deprecated fields.
+
+```graphql counter-example
+type Query {
+  returnsDeprecatedType: DeprecatedType
+}
+
+type DeprecatedType @deprecated {
+  id: ID
+}
+```
 
 ### @specifiedBy
 
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -109,7 +109,7 @@ CommonMark-compliant Markdown renderer.
 **Deprecation**
 
 To support the management of backwards compatibility, GraphQL fields, arguments,
-input fields, and enum values can indicate whether or not they are deprecated
+input fields, enum values, and objects can indicate whether or not they are deprecated
 (`isDeprecated: Boolean`) along with a description of why it is deprecated
 (`deprecationReason: String`).
 
@@ -126,7 +126,7 @@ which are fully defined in the sections below.
 ```graphql
 type __Schema {
   description: String
-  types: [__Type!]!
+  types(includeDeprecated: Boolean = false): [__Type!]!
   queryType: __Type!
   mutationType: __Type
   subscriptionType: __Type
@@ -151,6 +151,8 @@ type __Type {
   ofType: __Type
   # may be non-null for custom SCALAR, otherwise null.
   specifiedByURL: String
+  isDeprecated: Boolean!
+  deprecationReason: String
 }
 
 enum __TypeKind {
@@ -236,6 +238,8 @@ Fields\:
 - `types` must return the set of all named types contained within this schema.
   Any named type which can be found through a field of any introspection type
   must be included in this set.
+  - Accepts the argument `includeDeprecated` which defaults to {false}. If
+    {true}, deprecated fields are also returned.
 - `directives` must return the set of all directives available within this
   schema including all built-in directives.
 
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -109,9 +109,9 @@ CommonMark-compliant Markdown renderer.
 **Deprecation**
 
 To support the management of backwards compatibility, GraphQL fields, arguments,
-input fields, enum values, and objects can indicate whether or not they are deprecated
-(`isDeprecated: Boolean`) along with a description of why it is deprecated
-(`deprecationReason: String`).
+input fields, enum values, and objects can indicate whether or not they are
+deprecated (`isDeprecated: Boolean`) along with a description of why it is
+deprecated (`deprecationReason: String`).
 
 Tools built using GraphQL introspection should respect deprecation by
 discouraging deprecated use through information hiding or developer-facing
@@ -142,7 +142,7 @@ type __Type {
   # must be non-null for OBJECT and INTERFACE, otherwise null.
   interfaces: [__Type!]
   # must be non-null for INTERFACE and UNION, otherwise null.
-  possibleTypes: [__Type!]
+  possibleTypes(includeDeprecated: Boolean = false): [__Type!]
   # must be non-null for ENUM, otherwise null.
   enumValues(includeDeprecated: Boolean = false): [__EnumValue!]
   # must be non-null for INPUT_OBJECT, otherwise null.
@@ -315,6 +315,8 @@ Fields\:
 - `description` may return a String or {null}.
 - `possibleTypes` returns the list of types that can be represented within this
   union. They must be object types.
+  - Accepts the argument `includeDeprecated` which defaults to {false}. If
+    {true}, deprecated fields are also returned.
 - All other fields must return {null}.
 
 **Interface**
@@ -336,6 +338,8 @@ Fields\:
   (if none, `interfaces` must return the empty set).
 - `possibleTypes` returns the list of types that implement this interface. They
   must be object types.
+  - Accepts the argument `includeDeprecated` which defaults to {false}. If
+    {true}, deprecated fields are also returned.
 - All other fields must return {null}.
 
 **Enum**


xuorig:
I'd love a mechanism like this to make GraphQL continuous evolution more powerful 👍 Although it doesn't necessarily require a change to the spec, I also like that this can also be used server-side, to indicate new functionality should not be added to a type that is going away.

Initial thoughts: Deprecating a type implies both deprecation all fields of that type and all spreads on that type. Should adding `@deprecated` on a type be invalid if `@deprecated` wasn't added on all fields for example? Should it be automatically inferred by clients? Automatically added by servers?

Are there weird edge cases we need to think about (All union members are deprecated, does that make the field deprecated?) Do we leave that decision to schema designers or to the client handling of deprecations or does something need to be added to the spec, etc.

xuorig:
@jturkel that's a good point, we've also seen the need for union member deprecation and you're right that it doesn't always imply  the full deprecation of that type. If we had that then type deprecation can be achieved through combination of field deprecations + membership deprecations? Clients directly relying on types through introspection (not through a query) for things like code generation could still gain from deprecation of the type itself I guess 🤔

xuorig:
I think the what @jturkel mentioned might be easier to get into the spec, at least at first, and most likely responds to the use case more directly. Maybe we can approach it with something like this.

#### Unions

Requires new directive location:

```graphql
union Animal = Dog @deprecated | Cat | Bird
```

Without new directive location

```graphql
union Animal @deprecatedMembers(members: ["Dog"]) = Dog | Cat | Bird
```

#### Interfaces

```graphql
type Dog implements Node & Animal @deprecatedImplementations(implementations: ["Node"]) {
  id: ID!
  name: String
}
```

Here we'd only be concerned with the interface implementation, and the `id` field would need to be deprecated separately if needed.

jturkel:
I guess the location of directives on type definitions makes it impossible to unambiguously annotate the interface implementation using the current `@deprecated` directive (unless it used prefix notation which would be confusingly inconsistent with other directives) i.e.:

```graphql
# Deprecates the type rather than the interface implementation
type Dog implements Animal & Node @deprecated {
  id: ID!
  name: String
}
```

Will this change require introducing new introspection types and fields to model union membership and interface implementation since `__Type.interfaces` and `__Type.possibleTypes` directly return the underlying types?

xuorig:
> Will this change require introducing new introspection types and fields to model union membership and interface implementation since __Type.interfaces and __Type.possibleTypes directly return the underlying types?

Maybe we could add `includeDeprecatedImplementations` and `includeDeprecatedMembers` to those introspection fields similar to the way we do it for fields at the moment.

jturkel:
> Maybe we could add includeDeprecatedImplementations and includeDeprecatedMembers to those introspection fields similar to the way we do it for fields at the moment.

That would work for including/excluding deprecated implementations/members but tooling would also want to know if a given implementation/members was deprecated or not for query linting, documentation generation, etc.

spec/Section 4 -- Introspection.md
```diff
@@ -151,6 +151,8 @@ type __Type {
   ofType: __Type
   # may be non-null for custom SCALAR, otherwise null.
   specifiedByURL: String
+  isDeprecated: Boolean!
```
fotoetienne, line 154: Hm. `__Type` includes non-objects (unions, interfaces, anything in `__TypeKind`). Maybe this field should be nullable like `possibleTypes`? Or defaults to `false` for any type that can't be deprecated?

xuorig:
> That would work for including/excluding deprecated implementations/members but tooling would also want to know if a given implementation/members was deprecated or not for query linting, documentation generation, etc.

Ah yes I see what you mean now. For backward compatibility we could start by including it alongside, but yeah 😐.

leebyron:
This would require adding a validation rule that a deprecated object can't be used from a non-deprecated field?

leebyron:
Note towards clarifying motivation: Use of fragment spreads of a type is why removing this directly without a deprecation direction is a breaking change and motivates having this tool

gwardwell:
I like the direction this proposal is taking. I would advocate for expanding the locations of `@deprecated` to include objects as well as handling deprecation of a specific use of an object or interface. Both are credible use cases.

In the case of deprecating an object, you _could_ imply that an object is deprecated by deprecating all the fields of the object. But for specifications built on top of GraphQL (for example, Apollo Federation or schema stitching) inferring that an object is deprecated breaks down once the schema is extended across several services. This creates a need for deprecating the object itself so all uses of the object, both across services/schema SDL and in client applications are caught.

I agree that specific uses of objects should also be able to be deprecated. The examples of deprecating a specific interface from a specific type or a specific type from a specific union should be possible.

TL;DR I would advocate for both solutions (expanding `@deprecated` locations and adding new directives).

fotoetienne:
I created a [PR](https://github.com/graphql/graphql-js/pull/3818) on graphql-js implementing the proposed solution.

gwardwell:
This should expand to also include:

- Deprecating scalars
- Deprecating Input objects
- Deprecating enums

```graphql
scalar DateTime @deprecated

input BookInput @deprecated {
  title: String!
}

enum Genre @deprecated {
  SCIENCE_FICTION
}
```

While I realize there aren't any client benefits to deprecating these schema types, there is a great deal of service-level benefit as it will allow communication of what should or shouldn't be used in a large-scale GraphQL implementation. IDE tools could be updated to provide all the expected warnings when trying to use a deprecated scalar, for example.

gwardwell:
Piling on here, union members and interface implementations should be able to be deprecated. Something like:

```graphql
directive @deprecatedMember(member: String!, reason: String) repeatable on UNION

directive @deprecatedImplementation(interface: String!, reason: String) repeatable on Object

union PrintedMedia @deprecatedMember(member: "EReader", reason: "EReaders aren't printed.") = Book | Magazine | EReader

interface FunStuff {
  name: String!
}

type Dentist implements FunStuff @deprecatedImplementation(interface: "FunStuff", reason: "Going to the dentist isn't fun") {
  name: String!
}
```

For clients, such deprecation would tell them when they are querying against on of these deprecated named fragments. For example, when querying unions:

```graphql
query GetPrintedMedia {
  printedMedia {
    ... on EReader {
      # This is still a valid named fragment, but client tooling could warn that
      # this member of the union is deprecated and will go away. This would allow
      # clients to move away from the deprecated union.
      # ...
    }
  }
}
```

Querying interfaces would behave the same way.

martinbonnin:
@gwardwell we could also explore something like:

```graphql
enum __DirectiveLocation {
   # ...
   UNION_MEMBER
   INTERFACE
}

directive @deprecated(
  reason: String = "No longer supported"
) on 
 # ...
 UNION_MEMBER | INTERFACE

union PrintedMedia  = Book | Magazine | @deprecated(reason: "EReaders aren't printed") EReader 
type Dentist implements @deprecated(reason: "Going to the dentist isn't fun") FunStuff
```

Could that work?

Also I'm curious if other languages allow this? I get the usage but haven't seen this pattern anywhere else.

gwardwell:
@martinbonnin That's an interesting idea. I was trying to avoid adding new directive locations, but I'm all for it if it makes sense. I think they would need to be `UNION_MEMBER` and `INTERFACE_IMPLEMENTATION` since there's already an `INTERFACE` location (unless I misunderstand what qualifies as the `INTERFACE` location).

Regarding other languages and the deprecation of union members or interface implementations, I'm not aware of any that do it. But I'm also not aware of any that expose the union and interfaces externally through an API like GraphQL does. The motivation here is to discourage the use of inline or named fragments that explicitly reference a deprecated member or implementer. I think that use case is pretty unique.

I also opened issue #997 to discuss a more broad changes to `@deprecated`, but I'd be happy to consolidate over here. My concern was that this PR had run out of steam due to lack of activity.

martinbonnin:
> INTERFACE_IMPLEMENTATION since there's already an INTERFACE location

Good callout, I edited to `IMPLEMENTED_INTERFACE` (not great but at least doesn't name classh with existing locations).

benjie:
I'm generally supportive of adding `@deprecated` to more locations, but proposing additional locations for `@deprecated` should be done through separate issues/PRs/RFCs (ideally one PR per location to keep the discussion on-topic). Please keep this discussion to feedback specifically related to adding `@deprecated` to object types.

dondonz:
Hello! I was thinking of object type deprecation today after seeing a case at work where we'd love to be able to deprecate an object type. Over the past year, a number of colleagues have asked me about it too.

I think starting with object type deprecation is a great idea. Is there anything else you need to get this draft spec change progressed? For example, are you looking to iron out edge cases, or do you want to trial this in some libraries first and collect more feedback (as we are doing for the new `@oneOf` directive)?

fotoetienne:
> Hello! I was thinking of object type deprecation today after seeing a case at work where we'd love to be able to deprecate an object type. Over the past year, a number of colleagues have asked me about it too.
> 
> I think starting with object type deprecation is a great idea. Is there anything else you need to get this draft spec change progressed? For example, are you looking to iron out edge cases, or do you want to trial this in some libraries first and collect more feedback (as we are doing for the new `@oneOf` directive)?

Hi @dondonz ! One of the [next steps](https://github.com/graphql/graphql-wg/blob/604627218a80534ea4d7e05d9c1b1edbdcbd4fb2/notes/2023/2023-06.md?plain=1#L130) to get this RFC to phase 2 is to add validations to graphql-js.

Would you be interested in adding relevant validations to graphql-js?

dondonz:
Hello,

I am sadly a JS/TS noob so it would be better if someone else implements the graphql-js version.

However, I am happy to implement it in graphql-java, if the working group wants to gather more feedback before finalising the spec wording. Would that help move this proposal to RFC2 stage?

I saw on your link to the working group minutes, Michael mentioned adding this proposal to Hot Chocolate. Do you know if any libraries previously implemented this proposal?

fotoetienne:
> Hello,
> 
> I am sadly a JS/TS noob so it would be better if someone else implements the graphql-js version.
> 
> However, I am happy to implement it in graphql-java, if the working group wants to gather more feedback before finalising the spec wording. Would that help move this proposal to RFC2 stage?
> 
> I saw on your link to the working group minutes, Michael mentioned adding this proposal to Hot Chocolate. Do you know if any libraries previously implemented this proposal?

That would be great to implement it in graphql-java! I don't know of any other implementations yet (besides the [PR](https://github.com/graphql/graphql-js/pull/3818/files#diff-0692ebfea743a0a525e35affba2414bb9ab2f1217c22c7b472acb99d705f8375) on graphql-js). It looks like the [Issue](https://github.com/ChilliCream/graphql-platform/issues/5531) on Hot Chocolate is still open

Thanks!

dondonz:
Ok it's on! I'll work on this in graphql-java then :D

If you'd like any other help (including turning up to a WG meeting), let me know, I'd love to see this in the spec.

spec/Section 4 -- Introspection.md
```diff
@@ -126,7 +126,7 @@ which are fully defined in the sections below.
 ```graphql
 type __Schema {
   description: String
-  types: [__Type!]!
+  types(includeDeprecated: Boolean = false): [__Type!]!
```
bbakerman:
```
includeDeprecated: Boolean = false
```

This was a major design flaw original when done like that on fields.  Why?

Because tooling will use introspection to get the field of a type t build local schemas and this leads to invalid types

```
type Foo {
   field : String @deprecated
}
```

When `field` is not included by default - its `type Foo {}` and thats illegal and tooling blows up

Now for consistency sake we will repeat the same design flaw

```
union Members = Jimmy | Hoffa

type Jimmy @deprecated {
   buried : String
}

type Hoffa @deprecated {
  giantsStaium : String
}
```

now when the introspection drops out `Jimmy` and `Hoffa` fro the results - the local schema will be invalid

Some will says "well set includeDeprecated:true" but tell that to graphiql and all the other tooling that hasn't done that.  The same mistake will be repeated again in the name of consistency

glen-84: These should be `Boolean!` now.

spec/Section 4 -- Introspection.md
```diff
@@ -109,9 +109,9 @@ CommonMark-compliant Markdown renderer.
 **Deprecation**
 
 To support the management of backwards compatibility, GraphQL fields, arguments,
-input fields, and enum values can indicate whether or not they are deprecated
-(`isDeprecated: Boolean`) along with a description of why it is deprecated
-(`deprecationReason: String`).
+input fields, enum values, and objects can indicate whether or not they are
+deprecated (`isDeprecated: Boolean`) along with a description of why it is
```
glen-84, line 113: `isDeprecated: Boolean!`

Edit: Actually, it's nullable for `__Type`, so not sure what to write here.

spec/Section 4 -- Introspection.md
glen-84:
```suggestion
  # must be non-null for OBJECT, otherwise null.
  isDeprecated: Boolean
```

spec/Section 3 -- Type System.md
```diff
@@ -2098,6 +2098,36 @@ type ExampleType {
...
+
+union MyUnion = ExampleType | DeprecatedType
+
+type DeprecatedType @deprecated {
+  id: ID
+}
+```
+
+The `@deprecated` directive should not appear on object types that are
+referenced by non-deprecated fields.
```
glen-84, line 2119: If this is an error, should it instead be `must not`?

PR #1005, opened 2023-01-12
Change 'server' to 'service' for consistency

benjie:
Outside of the introduction/overview, this is the only mention of the word "server" in the entire GraphQL spec. Since the paragraph opens talking about GraphQL services, let's use the term service here for consistency.


leebyron:
Agreed, this was the intent thanks for catching this

PR #1008, opened 2023-01-13
fix operator precedence in test:format

yaacovCR:
I used [this question and answer](https://unix.stackexchange.com/questions/88850/precedence-of-the-shell-logical-operators)  about the sometimes-surprising behavior of logical operator precedence in Bash as a reference.


package.json
```diff
@@ -16,7 +16,8 @@
     "test": "npm run test:spelling && npm run test:format && npm run test:build",
     "test:spelling": "cspell \"spec/**/*.md\" README.md",
     "format": "prettier --write \"**/*.{md,yml,yaml,json}\"",
-    "test:format": "prettier --check \"**/*.{md,yml,yaml,json}\" || echo \"\nTo resolve this, run: $(tput bold)npm run format$(tput sgr0)\" && exit 1",
+    "test:format": "prettier --check \"**/*.{md,yml,yaml,json}\" || npm run suggest:format",
+    "suggest:format": "echo \"\nTo resolve this, run: $(tput bold)npm run format$(tput sgr0)\" && exit 1",
```
benjie, line 20: I don't think anyone would ever want to run `suggest:format` so making it a top-level script doesn't seem desirable. Let's just add parens:

```suggestion
    "test:format": "prettier --check \"**/*.{md,yml,yaml,json}\" || ( echo \"\nTo resolve this, run: $(tput bold)npm run format$(tput sgr0)\" && exit 1 )",
```

```bash
$ (true || ( echo 'hi' && exit 1 ))
$ echo $?
0
$ (false || ( echo 'hi' && exit 1 ))
hi
$ echo $?
1
```

benjie:
Issue was introduced in https://github.com/graphql/graphql-spec/commit/b9fad9b5ff8aa0d86225711fb87061525399d98f

benjie (dismissed):

package.json
benjie: Actually, we don't use `yarn` so this is unsafe. Windows shell uses different syntax IIRC, and though yarn tries to make it consistent, npm doesn't.

benjie (approved):
Thanks for the quick fix @yaacovCR  🙌

leebyron:
Til. Thanks for the quick fix folks

PR #1015, opened 2023-02-09
RFC: Root Types clarity

leebyron:
This attempts to clarify some naming conventions in the spec and makes one meaningful syntax change.

This changes `SchemaDefinition` to have a zero-or-more list of `RootTypeDefinition` rather than one-or-more. It removes some language that refers to this "at least one" requirement as well.

Why? Because schema can be extended with `SchemaExtension`. This brings this into consistency with other elements of the type definition language where definitions allow providing zero items for aspects that require one at the point of schema validation. This was overlooked before since most people use the default names, but it is not required to do so, and sometimes you need to supply the schema definition for explicitness, but are within the context of a single file that is a subset of a full schema where other root types are defined elsewhere.

Specifically, this example used to not parse, and now should:

```graphql
schema {}
```

What has not changed is that schema are required to provide a query root type. A schema without this root type will not pass validation, and would later need something like this to become valid:

```graphql
extend schema { query: MyQueryType }
```

The bulk of this change edits naming of some concepts in an attempt to improve clarity:

- Renames `OperationType` to `OperationKind`. This was already inconsistently used throughout the spec doc, but now is consistently using "kind". This avoids overloading the term "type".

- Renames `RootOperationTypeDefinition` to `RootTypeDefinition`. "Root" here is well defined to provide the type for a root selection set of an operation, so this shorter name is less redundant and easier to read. It also creates consistency with "default root type name".

- Replaces "root field" with "root selection set". A root field is a holdover concept from a 10yr old version of GraphQL. More accurately today these are "fields on the root selection set". This makes the latter well defined. I searched for "root" to make sure it refers only to a well defined "root type" or "root selection set".


spec/Section 4 -- Introspection.md
```diff
@@ -69,33 +69,32 @@ underscores {"\_\_"}.
 ## Type Name Introspection
 
 GraphQL supports type name introspection within any selection set in an
-operation, with the single exception of selections at the root of a subscription
-operation. Type name introspection is accomplished via the meta-field
-`__typename: String!` on any Object, Interface, or Union. It returns the name of
-the concrete Object type at that point during execution.
+operation, with the single exception of a subscription _root selection set_.
```
benjie, line 72: I believe this introduces ambiguity. 

A [selection set](https://spec.graphql.org/draft/#sec-Selection-Sets) consists of field, fragment spread and inline fragment selections. For example:

```graphql
subscription {
  ...Frag
}
fragment Frag on Subscription {
  ...Frag2
}
fragment Frag2 on Subscription {
  myField
}
```

^ this has **three** selection sets. The "root" selection set is, arguably, `{ ...Frag }`. (Note we also use the term "top level selection set" for this: https://spec.graphql.org/draft/#sel-LALPHDDDDDBCBIBf2we)

We need a term for a "resolved" selection set, i.e. the kind of thing you would get from `CollectFields()`. We call that `groupedFieldSet` in some places.

Perhaps the term "root field set" would work, but we'd also need to define the term "field set" I think.

Arguably "root field" is actually more correct here - there must be exactly one root field (after resolving the selection sets).

benjie (changes requested):

spec/Section 4 -- Introspection.md
leebyron: Great find. Let me ponder on this

leebyron: Rephrasing this as "field collected from the root selection set"

benjie (changes requested):
I like this change in general, but I have one concern...

In English, "type" can very easily be interpreted to mean the same as "kind", in fact ["kind" shows up as part of the definition of "type" on Dictionary.com](https://www.dictionary.com/browse/type).

So having "operation type" and "kind of operation" as distinct concepts when they _feel_ the same to a casual reader doesn't sit quite right with me.

I'm in favour of using "root type" rather than "operation type" - I think "root type" and "root selection set" pair nicely together, and this removes the above confusion.

mjmahone:
The argument for this change has a symmetric argument for allowing empty type and interface definitions, a la:
```
type MyType {}
interface MyInterface {}
```

PR #1022, opened 2023-03-22
Strawman: add sameAs property to __Type to allow mapping to other schemas

saerdnaer:
related to https://github.com/w3c/graphql-rdf/issues/4

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -151,6 +151,8 @@ type __Type {
   ofType: __Type
   # may be non-null for custom SCALAR, otherwise null.
   specifiedByURL: String
+  # references to other schemas/ontologies via URI/IRI
+  sameAs: [String!]
 }
 
 enum __TypeKind {


benjie:
Hi @saerdnaer; maybe you want to bring this to a WG meeting for discussion? https://github.com/graphql/graphql-wg/tree/main/agendas/2023

benjie:
I know you proposed this for object types, but I feel this could be broadened to increase value. E.g. rather than being "same as" perhaps it could be "conforms to"? This would be particularly valuable for custom scalars, e.g.

```graphql
scalar Name
  # Treat Name as a String
  @conformsTo(url: "https://spec.graphql.org/September2025/#sec-String")
  # Note it's also a Name
  @conformsTo(url: "https://schema.org/name")
  # Note it's ALSO a GraphQL Name (i.e. constrained string)
  @conformsTo(url: "https://spec.graphql.org/September2025/#Name")
```

One potential issue with this framing is the input/output tension; e.g. on output it's fine to just treat it as an arbitrary `string` - you'll only get string values from it. But on input all of the constraints apply, so you can't send any arbitrary string value - the value you send must be `/^[A-Za-z_][A-Za-z0-9_]*$/` to comply with GraphQL `Name`. I think this is okay (and probably desired!) though.

For object types, if it conforms to something I think that means it implements _at least_ the specified fields, but that it may additionally implement additional fields. I also think that this is probably better suited to a GraphQL interface - the object implement the interface, and the interface conforms to the requirements of the external URL.

saerdnaer:
Yeah, I was actually thinking about the same thing – we should define one build-in directive + introspection property which works of all kind of GraphQL Base Entities, and maybe add a second parameter which clarifies how deep the relationship goes / both match with each other. For example, during working on LinkML schema, I came across following definitions:

* [exact](https://biolink.github.io/biolink-model/using-the-modeling-language/#exact_mappings): external concepts, predicates, or properties which are considered to be exact mappings to the class [or type] being defined.
  * These indicate semantic equivalence.
* [close](https://biolink.github.io/biolink-model/using-the-modeling-language/#close_mappings): external concepts, predicates, or properties that are considered to be close mappings (similar but not fully identical). 
  * Used when the match is strong but slightly looser than exact.
* [narrow](https://biolink.github.io/biolink-model/using-the-modeling-language/#narrow_mappings): external concepts, predicates, or properties which are narrower in scope than the class [or type] being defined.
  * Narrow mappings are more restrictive. If modeled, they would be children of the broader concept (e.g., same as).
* [broad](https://biolink.github.io/biolink-model/using-the-modeling-language/#broad_mappings): external concepts, predicates, or properties which are broader in scope than the class [or type] being defined.
  *  Broad mappings are more general. If modeled, they would be parents of the narrower concept (e.g., in complex with).
* [related](https://biolink.github.io/biolink-model/using-the-modeling-language/#related_mappings): external concepts, predicates, or properties that are related in meaning but not strictly broader, narrower, or exact.
  * Indicates semantic relatedness without strong hierarchical or equivalence implications.

[For other readers: You might also have a look at the parallel, more SDL driven, discussion in https://github.com/w3c/graphql-rdf/issues/4 ]

PR #1026, opened 2023-05-21
RFC: incremental delivery with deduplication + concurrent execution

yaacovCR:
These spec edits should correspond to the working implementation of incremental delivery with deduplication currently posted as PR stack:
https://github.com/graphql/graphql-js/pull/3894 => introduces Publisher
https://github.com/graphql/graphql-js/pull/3886 => bulk of effort, introduces deduplication
https://github.com/graphql/graphql-js/pull/3897 => adds pending

These spec edits do not currently include optional follow-on:
https://github.com/graphql/graphql-js/pull/3895 => payload consolidation

[The diff to main might be helpful, but this is built on top of the amazing #742 and so the diff [from that branch could be more useful](https://github.com/robrichard/graphql-spec/compare/incremental...yaacovCR:graphql-spec:deduplicate).]

IMPORTANT NOTE:

These are hopefully trending toward complete in terms of the algorithm -- more explanatory prose should definitely be added.

<hr>

### TLDR?

The implementation and spec changes show how one can start executing deferred fragments semi-immediately (i.e. after deferring in an implementation-specific way), rather than waiting for the entire initial result to be emitted. This is not required -- one could still be compliant with the spec by deferring all the way until the initial result completes! In fact, how one defers is not per se observable and so the spec cannot mandate much about it with great normative force. But -- and I think this is important -- this PR and the implementation PR provide an algorithm/implementation/spec changes that give servers the flexibility to do what they think is right in that regard, and that might be desirable.

As of this moment, I am fairly confident in the implementation PR over at `graphql-js`, and the spec PR should generally correspond, demonstrating:

 = the `Field Group` and `Defer Usage` record types that contain the information derived from the operation during field collection
  
 = the introduction of a distinction between `Incremental Data` records and `Subsequent Result` records. `Deferred Fragment` records and `Stream Items` records exemplify `Subsequent Result`s that are sent as a group to the client. But an individual `Deferred Fragment` Record may consist of a number of distinct `Deferred Grouped Field Set` records, which may overlap with other `Deferred Fragment` Records and should not be sent more than once. `Deferred Grouped Field Set` records are therefore a unit of `Incremental Data` and are tracked with a new record type. `Stream Items` records always contain a single unit of incremental data that is sent only once with little complication; they therefore represent both `Subsequent Result` and `Incremental Data` Records.

= the new `Publisher` construct, with a set of algorithms that create and manipulate Subsequent Result and Incremental Data records. `Mutation` of these records is not performed directly during execution, but only via interaction with the Publisher.

= the deferMap, which maps `Defer Usage` records to individual `Deferred Fragment` `Subsequent Result` records

** huge thanks to @urigo and @dotansimha of [the guild](https://the-guild.dev/) for sponsoring my open-source work on this. **

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -11,7 +11,6 @@ words:
   - openwebfoundation
   - parallelization
   - structs
-  - sublist
   - subselection
   # Fictional characters / examples
   - alderaan
--- a/cspell.yml
+++ b/cspell.yml
@@ -12,6 +12,7 @@ words:
   - parallelization
   - structs
   - subselection
+  - errored
   # Fictional characters / examples
   - alderaan
   - hagrid
--- a/cspell.yml
+++ b/cspell.yml
@@ -4,12 +4,14 @@ ignoreRegExpList:
   - /[a-z]{2,}'s/
 words:
   # Terms of art
+  - deprioritization
   - endianness
   - interoperation
   - monospace
   - openwebfoundation
   - parallelization
   - structs
+  - sublist
   - subselection
   # Fictional characters / examples
   - alderaan
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -2163,8 +2163,8 @@ fragment someFragment on User {
   omitted.
 - `label: String` - May be used by GraphQL clients to identify the data from
   responses and associate it with the corresponding defer directive. If
-  provided, the GraphQL Server must add it to the corresponding payload. `label`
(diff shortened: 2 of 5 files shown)


yaacovCR:
IMPORTANT NOTE:

These are hopefully trending toward complete with respect to the `Execution` section; the `Response` section must still be updated to match the new format.

yaacovCR:
closing in favor of #1034

PR #1034, opened 2023-07-12
RFC: incremental delivery with deduplication + concurrent execution

yaacovCR:
supersedes  #1026 

This version sets up an Incremental Publisher Record that includes:
(a) an Execution Event Queue input stream on the and
(b) a Subsequent Result output stream.

No mutations happen outside the algorithms that define any records.

With the following caveats:
1. The event queue is not considered to be "mutated" when different sub-algorithms push events into it.
2. The `CreateIncrementalPublisher()` algorithm is where the magic happens, a long algorithm where the Execution Event Handler and lazily executed Subsequent Result stream are managed concurrently. To increase readability, several subprocedures/macros/inner functions are defined that have access to the variables defined within the entire algorithm.

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -11,7 +11,6 @@ words:
   - openwebfoundation
   - parallelization
   - structs
-  - sublist
   - subselection
   # Fictional characters / examples
   - alderaan
--- a/cspell.yml
+++ b/cspell.yml
@@ -12,6 +12,7 @@ words:
   - parallelization
   - structs
   - subselection
+  - errored
   # Fictional characters / examples
   - alderaan
   - hagrid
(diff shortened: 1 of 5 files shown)


yaacovCR:
As before, these spec edits correspond to the working implementation of incremental delivery with deduplication currently posted as a PR stack, see https://github.com/graphql/graphql-js/pull/3886 for further discussion.

[The diff to main might be helpful, but this is built on top of the amazing https://github.com/graphql/graphql-spec/pull/742 and so the diff [from that branch could be more useful](https://github.com/robrichard/graphql-spec/compare/incremental...yaacovCR:graphql-spec:deduplicate2).]

yaacovCR:
And again: ** huge thanks to @Urigo and @dotansimha of [the guild](https://the-guild.dev/) for sponsoring my open-source work on this. **

yaacovCR:
Closing in favor of #1052

PR #1036, opened 2023-08-02
Add a note about omitting introspection type definitions

martinbonnin:
There is a similar note 

[for scalars](https://spec.graphql.org/draft/#sel-GAHXJHABAB_D4G):

```
When representing a GraphQL schema using the type system definition language, all built-in scalars 
must be omitted for brevity.
```

[for directives](https://spec.graphql.org/draft/#sel-FAHnBPLCAACCcooU):

```
When representing a GraphQL schema using the type system definition language any built-in directive 
may be omitted for brevity.
```

I didn't find anything for introspection types.

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -224,7 +224,7 @@ When representing a GraphQL schema using the type system definition language the
 _introspection type system definitions_ may be omitted for brevity.
 
 When introspecting a GraphQL service all provided _introspection type system
-definitions_ must be included in the set of returned directives.
+definitions_ must be included in the set of returned types.
 
 ### The \_\_Schema Type
 
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -220,6 +220,12 @@ enum __DirectiveLocation {
 }
 ```
 
+When representing a GraphQL schema using the type system definition language the
+_introspection type system definitions_ may be omitted for brevity.
+
+When introspecting a GraphQL service all provided _introspection type system
+definitions_, must be included in the set of returned directives.
+
 ### The \_\_Schema Type
 
 The `__Schema` type is returned from the `__schema` meta-field and provides all
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -224,7 +224,7 @@ When representing a GraphQL schema using the type system definition language the
 _introspection type system definitions_ may be omitted for brevity.
 
 When introspecting a GraphQL service all provided _introspection type system
-definitions_, must be included in the set of returned directives.
+definitions_ must be included in the set of returned directives.
 
 ### The \_\_Schema Type
 


spec/Section 4 -- Introspection.md
```diff
@@ -220,6 +220,12 @@ enum __DirectiveLocation {
 }
 ```
 
+When representing a GraphQL schema using the type system definition language the
+_introspection type system definitions_ may be omitted for brevity.
+
+When introspecting a GraphQL service all provided _introspection type system
+definitions_, must be included in the set of returned directives.
```
BoD, line 227: I think the `,` shouldn't be there

spec/Section 4 -- Introspection.md
```diff
@@ -220,6 +220,12 @@ enum __DirectiveLocation {
 }
 ```
 
+When representing a GraphQL schema using the type system definition language the
+_introspection type system definitions_ may be omitted for brevity.
+
+When introspecting a GraphQL service all provided _introspection type system
+definitions_ must be included in the set of returned directives.
```
benjie, line 227:
```suggestion
definitions_ must be included in the set of returned types.
```

benjie (changes requested):
Typo?

spec/Section 4 -- Introspection.md
martinbonnin: Ah yes, bad copy paste, thank you !

IvanGoncharov:
Based on Aug 3th WG discussion: I think we can untangle this change from scalars and direcives by saying that you have two options:
1. Status quo: Query root doesn't have `__schema` and `__type` so you shouldn't include Introspection types.
2. Query root has `__schema` and `__type` then you should include all the introspection type reachable from those two fields.

We can't tangle it with scalars because currently below is a valid behaviours for scalars:
You print scalars used only by introspection or standard directives even if you omitted those directives. Moreover `graphql-js` does that and we can't make such documents invalid or broken.

IvanGoncharov:
However, it creates a slightly different problem since introspection for the Query root type doesn't contain `__type` and `__schema`.
So technically you wouldn't be able to convert introspection to SDL with introspection types.

martinbonnin:
Thanks for circling back on this!

> Query root doesn't have __schema and __type so you shouldn't include Introspection types.

That sounds like a short term thing that doesn't have too many implications, I like this :) 👍. 

> You print scalars used only by introspection or standard directives even if you omitted those directives.

I'm confused. The spec [says](https://spec.graphql.org/draft/#sel-GAHXJHABAB_D4G):

```
When representing a GraphQL schema using the type system definition language, all built-in scalars 
must be omitted for brevity.
```

It's a "must" so my understanding is (in the current form at least), scalar type definitions must never be present in a SDL document?

`graphql-js` has the `printIntrospectionSchema()` that (I think) outputs them but it's an exception to the spec?

benjie:
Feels like we could do this by adding an `includeMeta`/similar option. We could either go with boolean like `includeDeprecated`, or we could be more granular to differentiate between whether `__typename` should or should not be included:

```diff
+ enum __IncludeMeta { NONE ALL ALL_EXCEPT_TYPENAME }
  type __Type {
    # ...
    fields(
      includeDeprecated: Boolean = false
+     includeMeta: __IncludeMeta = NONE
    ): [__Field!]
```

martinbonnin:
Closing in favor of https://github.com/graphql/graphql-spec/pull/1049

PR #1037, opened 2023-08-05
Add Appendix C - Built-in Definitions

martinbonnin:
As we dive into full schemas, introspection and more (see https://github.com/graphql/graphql-spec/pull/1036), I'd find it useful to have an authoritative source for built-in definitions that come with a GraphQL implementation (scalar, directives & introspection types).

It shows the built-in definition at a glance and makes it easy to diff, capture a change by looking at the git history. It's also nice for library authors because it uniformizes the descriptions.

The definitions are taken from graphql-js `printIntrospectionSchema()` with two small changes:

- I added the scalars definitions manually
- I carried over the `__Type.fields`, `__Type.interface`, etc... descriptions from the [current spec comments](https://spec.graphql.org/draft/#sel-IAJXNFAlCABAB8B3zK)

I was doing this for my own needs so figured out I might as well contribute it to the spec but I'm happy to move it somewhere else if you think this doesn't belong in the spec.


martinbonnin:
Would that qualify as an `editorial` change? 

The validation/execution behaviour is essentially unchanged. The main thing that could be considered a behaviour change is that the definitions proposed here contain descriptions. We could either:

* Remove descriptions 
* Add language that mandates the implementations use the same description as in the spec
* Add language that the implementations are free to change the description if they want
* Leave as is

spec/Section 4 -- Introspection.md
```diff
@@ -119,106 +119,10 @@ warnings.
...
-  INTERFACE
-  UNION
-  ENUM
-  ENUM_VALUE
-  INPUT_OBJECT
-  INPUT_FIELD_DEFINITION
-}
-```
+See [Appendix C](#sec-Introspection-types) for more details about the schema
+type definitions.
```
benjie, line 125: This should not be removed; it's heavily referenced whilst reading this section of the spec. It's okay to _repeat_ it in an appendix, like how we repeat the grammar in the appendix.

spec/Section 3 -- Type System.md
```diff
@@ -388,9 +388,10 @@ Scalar types represent primitive leaf values in a GraphQL type system. GraphQL
 responses take the form of a hierarchical tree; the leaves of this tree are
 typically GraphQL Scalar types (but may also be Enum types or {null} values).
 
-GraphQL provides a number of built-in scalars which are fully defined in the
-sections below, however type systems may also add additional custom scalars to
-introduce additional semantic meaning.
+GraphQL provides a number of
+[built-in scalars](#sec-Appendix-Built-in-Definitions.Scalars) which are fully
```
benjie, line 392: I don't think this should reference an appendix. Following the definitions below is fine. Note that Appendix B isn't actually referenced anywhere in the spec text.

benjie: Same for the below references, I think they also should be removed. The appendix doesn't _define_ the built in directives, it merely catalogues them in one place.

benjie (dismissed):
Let's keep this to just adding the appendix, it does not need to be referenced in the spec text (just as Appendix B is not referenced when grammar is mentioned).

spec/Section 4 -- Introspection.md
martinbonnin: Good callout 👍

martinbonnin:
I guess there's a distinction to be made between:

1. implementations MUST always support it and expose the definition in introspection: `@include`, `@defer`, etc...
2. implementations MAY support it but it may not always be present in introspection: `@defer`, `@disableErrorPropagation`, etc...

So I'm questioning wether `built-in definitions` is the proper term there. Maybe just `specified definitions`? Or just `definitions`? Thoughts?

spec/Appendix C -- Built-in Definitions.md
```diff
@@ -0,0 +1,366 @@
...
+an ID.
+"""
+scalar ID
+```
+
+## Directives
+
+```graphql
+"""
+Directs the executor to include this field or fragment only when the `if` argument is true
```
BoD, line 48:
```suggestion
Directs the executor to include this field or fragment only when the `if` argument is true.
```

spec/Appendix C -- Built-in Definitions.md
```diff
@@ -0,0 +1,366 @@
...
+"""
+directive @skip(
+  """
+  Skipped when true.
+  """
+  if: Boolean!
+) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+"""
+Marks the field, argument, input field or enum value as deprecated
```
BoD, line 68:
```suggestion
Marks the field, argument, input field or enum value as deprecated.
```

spec/Appendix C -- Built-in Definitions.md
```diff
@@ -0,0 +1,366 @@
...
+"""
+directive @deprecated(
+  """
+  The reason for the deprecation
```
BoD, line 72:
```suggestion
  The reason for the deprecation.
```

spec/Appendix C -- Built-in Definitions.md
```diff
@@ -0,0 +1,366 @@
...
+  """
+  must be non-null for INPUT_OBJECT, otherwise null.
+  """
+  inputFields(includeDeprecated: Boolean = false): [__InputValue!]
+  """
+  must be non-null for NON_NULL and LIST, otherwise null.
+  """
+  ofType: __Type
+  """
+  may be non-null for custom SCALAR, otherwise null.
```
BoD, line 159:
```suggestion
  Must be non-null for OBJECT and INTERFACE, otherwise null.
  """
  fields(includeDeprecated: Boolean = false): [__Field!]
  """
  Must be non-null for OBJECT and INTERFACE, otherwise null.
  """
  interfaces: [__Type!]
  """
  Must be non-null for INTERFACE and UNION, otherwise null.
  """
  possibleTypes: [__Type!]
  """
  Must be non-null for ENUM, otherwise null.
  """
  enumValues(includeDeprecated: Boolean = false): [__EnumValue!]
  """
  Must be non-null for INPUT_OBJECT, otherwise null.
  """
  inputFields(includeDeprecated: Boolean = false): [__InputValue!]
  """
  Must be non-null for NON_NULL and LIST, otherwise null.
  """
  ofType: __Type
  """
  May be non-null for custom SCALAR, otherwise null.
```

BoD (commented):
Added a few dots and capital letters.

BoD (approved):

benjie (dismissed):
I reproduced your work by starting from a clean slate and then extracting the relevant definitions from the relevant sections; doing so revealed a few discrepancies that have crept in since you authored this:

spec/Appendix C -- Built-in Definitions.md
```diff
@@ -0,0 +1,366 @@
...
+) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+"""
+Marks the field, argument, input field or enum value as deprecated
+"""
+directive @deprecated(
+  """
+  The reason for the deprecation
+  """
+  reason: String = "No longer supported"
```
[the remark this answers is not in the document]
martinbonnin: Good catch!

benjie (commented):
Let's discuss whether or not we need this, or whether it should be re-incorporated back into the other chapters of the spec... But if agreed that a new appendix is warranted then I confirm the current state of it as accurate :+1:

martinbonnin:
Feedback from May wg:

* Put everything in one giant block
* Add a script to get the block from graphql-js (update graphql-js if some things are out of order)

spec/Appendix C -- Built-in Definitions.md
```diff
@@ -0,0 +1,332 @@
...
+  name: String
+  description: String
+  specifiedByURL: String
+  fields(includeDeprecated: Boolean = false): [__Field!]
+  interfaces: [__Type!]
+  possibleTypes: [__Type!]
+  enumValues(includeDeprecated: Boolean = false): [__EnumValue!]
+  inputFields(includeDeprecated: Boolean = false): [__InputValue!]
+  ofType: __Type
+  isOneOf: Boolean
```
martinbonnin, line 129: Fun fact: I haven't found a version of GraphQL JS that matches the current draft:

* 16.11.0 has `@oneOf` but draft doesn't
* 16.8.0 doesn't have `@oneOf` but also doesn't have non-null `includeDeprecated` (which is in the draft)

spec/Section 4 -- Introspection.md
```diff
@@ -137,6 +137,8 @@ type __Type {
   kind: __TypeKind!
```
leebyron, line 137: drop changes to this file - update graphqljs

scripts/update-appendix-c.mjs
```diff
@@ -0,0 +1,28 @@
...
+
+The order of types, fields, arguments, values and directives is non-normative.
+
+\`\`\`graphql
+`
+
+const suffix = `
+\`\`\`
+`
+await writeFile(FILE, prefix + printSpecifiedScalars() + '\n\n' + introspectionSchema + suffix);
```
leebyron, line 28: add a line break at the end

PR #999, opened 2022-11-06
introduce ExecuteGroupedFieldSet, CollectRootFields and CollectSubfields

yaacovCR:
Rather than merging subSelectionSets of a field set using MergeSelectionSets and then calling CollectFields, this PR introduces CollectSubfields allows the field set's groupedSubfieldSet to be calculated directly.

The reference implementation already uses this algorithm so that this change actually aligns the specification to the reference implementation, and is ipso facto non-breaking.

Motivation: reformulating the specification in this manner may be helpful if the specification were ever to be altered such that additional state beyond the current selection set were to be required to calculate the response, i.e. if it were to be required to know the originating selectionSet of a given field within the fieldSet for determining when to communicate a [reference signal](https://github.com/graphql/graphql-spec/pull/998#issue-1435084489). In such a scenario, it may still be quite possible to merge the set of requested data from a field set's subSelectionSets, but it may not be possible to express that merged data as an equivalent selectionSet.

In particular, currently:

```graphql
{
  a {
    subfield1
  }
  ...ExampleFragment
}

fragment ExampleFragment on Query {
  a {
    subfield2
  }
  b
}
```

For the given set of fields:
```graphql
a {
  subfield1
}
a {
  subfield2
}
```

These can currently be trivially merged as:
```graphql
a {
  subfield1
  subfield2
}
```

However, the requested information for `a` in:
```graphql
{
  a {
    subfield1
  }
  Ref1 { completed } : ...ExampleFragment
}

fragment ExampleFragment on Query {
  a {
    subfield2
  }
  b
}
```

cannot be contained in a merged selection set under A, because some of those fields will be related to Ref1 and some will not. The requsted information can still be merged, but it cannot be expressed in selection set format.

PR #1039, opened 2023-08-21
Replace `ExecuteSelectionSet` with `ExecuteGroupedFieldSet`

benjie:
Essentially this PR raises the concept of "grouped field sets" to be a top-level concept, and replaces the `ExecuteSelectionSet` method with `ExecuteGroupedFieldSet` (which essentially just drops the first line of `ExecuteSelectionSet`, which was responsible for producing the grouped field set to be executed). It then refactors the rest of the spec to accommodate this change, reducing the repetition in `ExecuteQuery`,  `ExecuteMutation` and `ExecuteSubscriptionEvent`; and removing `MergeSelectionSets` (which generated a "virtual" selection set to accomodate the `ExecuteSelectionSet` method), instead adding a `CollectSubfields` algorithm which generates a grouped field set directly, ready for execution.

I extracted this common refactoring from a number of my attempts to write spec changes for the `@defer` and `@stream` directives - it turns out that this refactoring of the spec was always needed as a base for my changes. Similarly, @yaacovCR found similar in his attempts to address this same problem, and raised #999 extracted from his solution. This PR was introduced independently of #999 (other than using the `CollectSubfields` algorithm name) however there is significant alignment, so @yaacovCR suggested that I raise it as an alternative PR.

It may be easier to review this PR in "split" view rather than "unified" view.

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -321,31 +321,34 @@ Executing the root selection set works similarly for queries (parallel),
 mutations (serial), and subscriptions (where it is executed for each event in
 the underlying Source Stream).
 
+First, the selection set is turned into a grouped field set; then, we execute
+this grouped field set and return the resulting {data} and {errors}.
+
 ExecuteRootSelectionSet(variableValues, initialValue, objectType, selectionSet,
 serial):
 
 - If {serial} is not provided, initialize it to {false}.
-- Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
+- Let {groupedFieldSet} be the result of {CollectFields(objectType,
+  selectionSet, variableValues)}.
+- Let {data} be the result of running {ExecuteGroupedFieldSet(groupedFieldSet,
   objectType, initialValue, variableValues)} _serially_ if {serial} is {true},
   _normally_ (allowing parallelization) otherwise.
 - Let {errors} be the list of all _field error_ raised while executing the
   selection set.
 - Return an unordered map containing {data} and {errors}.
 
-## Executing Selection Sets
+## Executing a Grouped Field Set
 
-To execute a selection set, the object value being evaluated and the object type
-need to be known, as well as whether it must be executed serially, or may be
-executed in parallel.
+To execute a grouped field set, the object value being evaluated and the object
+type need to be known, as well as whether it must be executed serially, or may
+be executed in parallel.
 
-First, the selection set is turned into a grouped field set; then, each
-represented field in the grouped field set produces an entry into a response
-map.
+Each represented field in the grouped field set produces an entry into a
+response map.
 
-ExecuteSelectionSet(selectionSet, objectType, objectValue, variableValues):
+ExecuteGroupedFieldSet(groupedFieldSet, objectType, objectValue,
+variableValues):
 
-- Let {groupedFieldSet} be the result of {CollectFields(objectType,
-  selectionSet, variableValues)}.
 - Initialize {resultMap} to an empty ordered map.
 - For each {groupedFieldSet} as {responseKey} and {fields}:
   - Let {fieldName} be the name of the first entry in {fields}. Note: This value
@@ -363,8 +366,8 @@ is explained in greater detail in the Field Collection section below.
 
 **Errors and Non-Null Fields**
 
-If during {ExecuteSelectionSet()} a field with a non-null {fieldType} raises a
-_field error_ then that error must propagate to this entire selection set,
+If during {ExecuteGroupedFieldSet()} a field with a non-null {fieldType} raises
+a _field error_ then that error must propagate to this entire selection set,
 either resolving to {null} if allowed or further propagated to a parent field.
 
 If this occurs, any sibling fields which have not yet executed or have not yet
@@ -702,8 +705,9 @@ CompleteValue(fieldType, fields, result, variableValues):
     - Let {objectType} be {fieldType}.
   - Otherwise if {fieldType} is an Interface or Union type.
     - Let {objectType} be {ResolveAbstractType(fieldType, result)}.
-  - Let {subSelectionSet} be the result of calling {MergeSelectionSets(fields)}.
-  - Return the result of evaluating {ExecuteSelectionSet(subSelectionSet,
+  - Let {groupedFieldSet} be the result of calling {CollectSubfields(objectType,
+    fields, variableValues)}.
+  - Return the result of evaluating {ExecuteGroupedFieldSet(groupedFieldSet,
     objectType, result, variableValues)} _normally_ (allowing for
     parallelization).
 
@@ -750,9 +754,9 @@ ResolveAbstractType(abstractType, objectValue):
 
 **Merging Selection Sets**
 
-When more than one field of the same name is executed in parallel, their
-selection sets are merged together when completing the value in order to
-continue execution of the sub-selection sets.
+When more than one field of the same name is executed in parallel, during value
+completion their selection sets are collected together to produce a single
+grouped field set in order to continue execution of the sub-selection sets.
 
 An example operation illustrating parallel fields with the same name with
 sub-selections.
@@ -771,14 +775,19 @@ sub-selections.
 After resolving the value for `me`, the selection sets are merged together so
 `firstName` and `lastName` can be resolved for one value.
 
-MergeSelectionSets(fields):
+CollectSubfields(objectType, fields, variableValues):
 
-- Let {selectionSet} be an empty list.
+- Let {groupedFieldSet} be an empty map.
 - For each {field} in {fields}:
   - Let {fieldSelectionSet} be the selection set of {field}.
   - If {fieldSelectionSet} is null or empty, continue to the next field.
-  - Append all selections in {fieldSelectionSet} to {selectionSet}.
-- Return {selectionSet}.
+  - Let {subGroupedFieldSet} be the result of {CollectFields(objectType,
+    fieldSelectionSet, variableValues)}.
+  - For each {subGroupedFieldSet} as {responseKey} and {subfields}:
+    - Let {groupForResponseKey} be the list in {groupedFieldSet} for
+      {responseKey}; if no such list exists, create it as an empty list.
+    - Append all fields in {subfields} to {groupForResponseKey}.
+- Return {groupedFieldSet}.
 
 ### Handling Field Errors
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -131,12 +131,8 @@ ExecuteQuery(query, schema, variableValues, initialValue):
 - Let {queryType} be the root Query type in {schema}.
 - Assert: {queryType} is an Object type.
 - Let {selectionSet} be the top level Selection Set in {query}.
-- Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
-  queryType, initialValue, variableValues)} _normally_ (allowing
-  parallelization).
-- Let {errors} be the list of all _field error_ raised while executing the
-  selection set.
-- Return an unordered map containing {data} and {errors}.
+- Return {ExecuteRootSelectionSet(variableValues, initialValue, queryType,
+  selectionSet)}.
 
 ### Mutation
 
@@ -153,11 +149,8 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
 - Let {mutationType} be the root Mutation type in {schema}.
 - Assert: {mutationType} is an Object type.
 - Let {selectionSet} be the top level Selection Set in {mutation}.
-- Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
-  mutationType, initialValue, variableValues)} _serially_.
-- Let {errors} be the list of all _field error_ raised while executing the
-  selection set.
-- Return an unordered map containing {data} and {errors}.
+- Return {ExecuteRootSelectionSet(variableValues, initialValue, mutationType,
+  selectionSet, true)}.
 
 ### Subscription
 
@@ -301,12 +294,8 @@ ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
 - Let {subscriptionType} be the root Subscription type in {schema}.
 - Assert: {subscriptionType} is an Object type.
 - Let {selectionSet} be the top level Selection Set in {subscription}.
-- Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
-  subscriptionType, initialValue, variableValues)} _normally_ (allowing
-  parallelization).
-- Let {errors} be the list of all _field error_ raised while executing the
-  selection set.
-- Return an unordered map containing {data} and {errors}.
+- Return {ExecuteRootSelectionSet(variableValues, initialValue,
+  subscriptionType, selectionSet)}.
 
 Note: The {ExecuteSubscriptionEvent()} algorithm is intentionally similar to
 {ExecuteQuery()} since this is how each event result is produced.
@@ -321,6 +310,27 @@ the subscription.
 Unsubscribe(responseStream):
 
 - Cancel {responseStream}
+
+## Executing the Root Selection Set
+
+To execute the root selection set, the object value being evaluated and the
+object type need to be known, as well as whether it must be executed serially,
+or may be executed in parallel.
+
+Executing the root selection set works similarly for queries (parallel),
+mutations (serial), and subscriptions (where it is executed for each event in
+the underlying Source Stream).
+
+ExecuteRootSelectionSet(variableValues, initialValue, objectType, selectionSet,
+serial):
+
+- If {serial} is not provided, initialize it to {false}.
+- Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
+  objectType, initialValue, variableValues)} _serially_ if {serial} is {true},
+  _normally_ (allowing parallelization) otherwise.
+- Let {errors} be the list of all _field error_ raised while executing the
+  selection set.
+- Return an unordered map containing {data} and {errors}.
 
 ## Executing Selection Sets
 


spec/Section 6 -- Execution.md
```diff
@@ -761,14 +775,19 @@ sub-selections.
...
-MergeSelectionSets(fields):
+CollectSubfields(objectType, fields, variableValues):
 
-- Let {selectionSet} be an empty list.
+- Let {groupedFieldSet} be an empty map.
 - For each {field} in {fields}:
   - Let {fieldSelectionSet} be the selection set of {field}.
   - If {fieldSelectionSet} is null or empty, continue to the next field.
-  - Append all selections in {fieldSelectionSet} to {selectionSet}.
-- Return {selectionSet}.
```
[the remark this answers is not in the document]
benjie: Perhaps `childGroupedFieldSet`? `fieldGroupedFieldSet` sounds to me like a "field-grouped field set" rather than a "field's grouped-field-set".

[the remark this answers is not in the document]
yaacovCR: Hmm, child has the same problem, but maybe if we change fieldSelectionSet in tandem that would work anyway through some context signaling

robrichard (approved):

benjie:
(Rebased on `main`)

benjie:
(Rebased on `main`)

yaacovCR:
The only issue I noticed when working on incremental delivery, which depends on this, is the ordering of where the collect fields algorithm shows up — as it is now still in the original place, but it is used much earlier. This gets more complicated with incremental delivery as the incremental delivery plan building and execution has the same issue which exacerbates the ordering conundrum.

Might we consider moving collect fields earlier or into its own section as part of this PR?

benjie:
@yaacovCR Just to check I understand, you mean specifically reordering the sections in specification but keeping the algorithm steps the same, rather than reordering the steps within the algorithms, right?

Maybe you'd like to make a PR to this branch which does your proposed re-arrange without any extraneous edits?

yaacovCR:
https://github.com/graphql/graphql-spec/pull/1111

as just one option.

With incremental delivery, the Field Collection subsection may be supplemented by additional incremental-delivery specific subsections

@robrichard

JoviDeCroock (approved):

benjie:
@yaacovCR or @JoviDeCroock; please can you confirm this latest text still aligns with GraphQL.js.

@Keweiqu Since a "grouped field set" is only part of execution and doesn't have an expression in the language I've not defined it in chapter 2, but in chapter 6 instead.

yaacovCR (approved):

martinbonnin (approved):

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+    - If {skipDirective}'s {if} argument is {true} or is a variable in
+      {variableValues} with the value {true}, continue with the next {selection}
+      in {selectionSet}.
+  - If {selection} provides the directive `@include`, let {includeDirective} be
+    that directive.
+    - If {includeDirective}'s {if} argument is not {true} and is not a variable
+      in {variableValues} with the value {true}, continue with the next
+      {selection} in {selectionSet}.
+  - If {selection} is a {Field}:
+    - Let {responseKey} be the response key of {selection} (the alias if
```
mjmahone, line 414:
```suggestion
    - Let {responseName} be the response name of {selection} (the alias if
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+      defined, otherwise the field name).
+    - Let {groupForResponseKey} be the list in {groupedFields} for
+      {responseKey}; if no such list exists, create it as an empty list.
```
mjmahone, line 417:
```suggestion
      {responseName}; if no such list exists, create it as an empty list.
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+      in {selectionSet}.
+  - If {selection} provides the directive `@include`, let {includeDirective} be
+    that directive.
+    - If {includeDirective}'s {if} argument is not {true} and is not a variable
+      in {variableValues} with the value {true}, continue with the next
+      {selection} in {selectionSet}.
+  - If {selection} is a {Field}:
+    - Let {responseKey} be the response key of {selection} (the alias if
+      defined, otherwise the field name).
+    - Let {groupForResponseKey} be the list in {groupedFields} for
```
mjmahone, line 416:
```suggestion
    - Let {groupForResponseName} be the list in {groupedFields} for
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+      {responseKey}; if no such list exists, create it as an empty list.
+    - Append {selection} to the {groupForResponseKey}.
```
mjmahone, line 418:
```suggestion
    - Append {selection} to the {groupForResponseName}.
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+      {selectionSet}.
+    - Let {fragmentType} be the type condition on {fragment}.
+    - If {DoesFragmentTypeApply(objectType, fragmentType)} is {false}, continue
+      with the next {selection} in {selectionSet}.
+    - Let {fragmentSelectionSet} be the top-level selection set of {fragment}.
+    - Let {fragmentGroupedFieldSet} be the result of calling
+      {CollectFields(objectType, fragmentSelectionSet, variableValues,
+      visitedFragments)}.
+    - For each {fragmentGroup} in {fragmentGroupedFieldSet}:
+      - Let {responseKey} be the response key shared by all fields in
```
mjmahone, line 436:
```suggestion
      - Let {responseName} be the response name shared by all fields in
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+        {fragmentGroup}.
+      - Let {groupForResponseKey} be the list in {groupedFields} for
+        {responseKey}; if no such list exists, create it as an empty list.
```
mjmahone, line 439:
```suggestion
        {responseName}; if no such list exists, create it as an empty list.
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+      - Append all items in {fragmentGroup} to {groupForResponseKey}.
```
mjmahone, line 440:
```suggestion
      - Append all items in {fragmentGroup} to {groupForResponseName}.
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+    - Let {fragmentType} be the type condition on {selection}.
+    - If {fragmentType} is not {null} and {DoesFragmentTypeApply(objectType,
+      fragmentType)} is {false}, continue with the next {selection} in
+      {selectionSet}.
+    - Let {fragmentSelectionSet} be the top-level selection set of {selection}.
+    - Let {fragmentGroupedFieldSet} be the result of calling
+      {CollectFields(objectType, fragmentSelectionSet, variableValues,
+      visitedFragments)}.
+    - For each {fragmentGroup} in {fragmentGroupedFieldSet}:
+      - Let {responseKey} be the response key shared by all fields in
```
mjmahone, line 451:
```suggestion
      - Let {responseName} be the response name shared by all fields in
```

spec/Section 6 -- Execution.md
```diff
@@ -348,20 +337,151 @@ Unsubscribe(responseStream):
...
+        {fragmentGroup}.
+      - Let {groupForResponseKey} be the list in {groupedFields} for
+        {responseKey}; if no such list exists, create it as an empty list.
+      - Append all items in {fragmentGroup} to {groupForResponseKey}.
```
mjmahone, line 455:
```suggestion
      - Let {groupForResponseName} be the list in {groupedFields} for
        {responseName}; if no such list exists, create it as an empty list.
      - Append all items in {fragmentGroup} to {groupForResponseName}.
```

spec/Section 6 -- Execution.md
```diff
@@ -790,14 +806,22 @@ sub-selections.
...
+  - Let {fieldGroupedFieldSet} be the result of {CollectFields(objectType,
+    fieldSelectionSet, variableValues)}.
+  - For each {fieldGroupedFieldSet} as {responseKey} and {subfields}:
+    - Let {groupForResponseKey} be the list in {groupedFieldSet} for
+      {responseKey}; if no such list exists, create it as an empty list.
+    - Append all fields in {subfields} to {groupForResponseKey}.
+- Return {groupedFieldSet}.
+
+Note: All the {fields} passed to {CollectSubfields()} share the same _response
+key_.
```
mjmahone, line 824:
```suggestion
  - For each {fieldGroupedFieldSet} as {responseKey} and {subfields}:
    - Let {groupForResponseName} be the list in {groupedFieldSet} for
      {responseName}; if no such list exists, create it as an empty list.
    - Append all fields in {subfields} to {groupForResponseName}.
- Return {groupedFieldSet}.

Note: All the {fields} passed to {CollectSubfields()} share the same _response
name_.
```

mjmahone (approved):
Nits because we agreed to name things `responseName` instead of `responseKey`

spec/Section 6 -- Execution.md
```diff
@@ -357,19 +346,151 @@ Unsubscribe(responseStream):
...
+- Return an unordered map containing {data} and {errors}.
+
+### Field Collection
+
+:: A _grouped field set_ is a map where each entry is a list of field selections
+that share a _response name_ (the alias if defined, otherwise the field name).
+
+Before execution, the _selection set_ is converted to a _grouped field set_ by
+calling {CollectFields()}. This ensures all fields with the same response name
+(including those in referenced fragments) are executed at the same time.
```
benjie, line 382: These are the only lines that have actually been changed in Field Collection, everything else was just moved.

(To prove this, take the current `Field Collection` text and paste it over this PR, the diff should render as:

```
diff --git i/spec/Section 6 -- Execution.md w/spec/Section 6 -- Execution.md
index f81e38d..549930f 100644
--- i/spec/Section 6 -- Execution.md	
+++ w/spec/Section 6 -- Execution.md	
@@ -374,12 +374,11 @@ serial):
 
 ### Field Collection
 
-:: A _grouped field set_ is a map where each entry is a list of field selections
-that share a _response name_ (the alias if defined, otherwise the field name).
-
-Before execution, the _selection set_ is converted to a _grouped field set_ by
-calling {CollectFields()}. This ensures all fields with the same response name
-(including those in referenced fragments) are executed at the same time.
+Before execution, the _selection set_ is converted to a grouped field set by
+calling {CollectFields()}. Each entry in the grouped field set is a list of
+fields that share a _response name_ (the alias if defined, otherwise the field
+name). This ensures all fields with the same response name (including those in
+referenced fragments) are executed at the same time.
 
 As an example, collecting the fields of this selection set would collect two
 instances of the field `a` and one of field `b`:
```
)

benjie:
@leebyron I opted for a merge instead as it was less work, this is now good to review again.

(I diffed the diffs and only the expected changes have come through, I also searched for "response key" and `/^key/` to ensure no older keys still exist. This also addresses Matt's feedback I think.)

spec/Section 6 -- Execution.md
```diff
@@ -165,11 +161,8 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
 - Let {mutationType} be the root Mutation type in {schema}.
 - Assert: {mutationType} is an Object type.
 - Let {selectionSet} be the top level selection set in {mutation}.
-- Let {data} be the result of running {ExecuteSelectionSet(selectionSet,
-  mutationType, initialValue, variableValues)} _serially_.
-- Let {errors} be the list of all _execution error_ raised while executing the
-  selection set.
-- Return an unordered map containing {data} and {errors}.
+- Return {ExecuteRootSelectionSet(variableValues, initialValue, mutationType,
+  selectionSet, true)}.
```
leebyron, line 165: I don't believe a trailing optional boolean argument is very clear, an enum of strings might be easier to read

spec/Section 6 -- Execution.md
```diff
@@ -357,19 +346,151 @@ Unsubscribe(responseStream):
 
 - Cancel {responseStream}.
 
-## Executing Selection Sets
+## Executing the Root Selection Set
```
leebyron, line 349: this new header name seems like it should not be at `##` level anymore, but the sub-algos still matter. 

Perhaps this top level section should remain and explain the high level idea, and then have these be sub-headers

leebyron (commented):
Really love this cleanup, much closer to GraphQL.js

leebyron:
@benjie I took a pass at some re-ordering of the content here, I'd be curious for your feedback.

Also, I think "Grouped Field Set" has become a confusing name. Ideally this term "Collect" is what we use for this concept (instead of "group") and it's a bit weird this is called a "Set" when in fact its type is `Map<string, List<Field>>`. What do you think about renaming this? 

"Collected Field Map"?

benjie:
I support renaming it, but not convinced by that name.

:thinking: If I were putting it in a variable in my own code it would be `fieldsByResponseName`... But we wouldn't say "Executing a fieldsByResponseName"...

With "Collected Field Map" I don't like that "field" is singular - you might expect this to be `Map<ResponseName, Field>`. "Grouped Field Set" worked here because it's a Field Set (`Set<Field>`) that's then been grouped by response name to produce `Map<ResponseName, Set<Field>>` (where the `Set<Field>` inside the Map form partitions of the original `Set<Field>`). Each partition is an ordered set, but the distinction between ordered set and list is unimportant here. The fact it's ultimately a map is implied by the word "grouped".

:thinking:

You know what... I've talked myself out of it. I think we should leave it as is, or maybe rename it to "Grouped Field Sets" (i.e. pluralize it). I'd also be amenable to renaming it to "Collected Field Sets", though I think that's a less clear description of what the data actually _is_ (though it does surface from whence it originated).

Maybe we should leave the rename for a later PR?

benjie:
Scratch that. I'd be happy with "Collected Field**s** Map", but I think it's important "field" is plural, either via `fields` or via `list` or `set`.

> :: A _collected fields map_ is a map where each entry is a _response name_ and a
list of selected fields that share that _response name_ (the field alias if
defined, otherwise the field's name).

> ### Executing a Collected Fields Map
> 
> To execute a _collected fields map_, the object value being evaluated and the object type need to be known, as well as whether it must be executed serially, or may be executed in parallel (see [Normal and Serial Execution](#sec-Normal-and-Serial-Execution).
> 
> Each entry in the collected fields map represents a _response name_ which produces
an entry into a result map.

> ExecuteCollectedFieldsMap(collectedFieldsMap, objectType, objectValue,
variableValues):
> - Initialize {resultMap} to an empty ordered map.
> - For each {collectedFieldsMap} as {responseName} and {fields}:
>   - Let {fieldName} be the name of the first entry in {fields}. Note: This value
>     is unaffected if an alias is used.
>   - Let {fieldType} be the return type defined for the field {fieldName} of
>     {objectType}.
>   - If {fieldType} is defined:
>     - Let {responseValue} be {ExecuteField(objectType, objectValue, fieldType,
>       fields, variableValues)}.
>     - Set {responseValue} as the value for {responseName} in {resultMap}.
> - Return {resultMap}.

spec/Section 6 -- Execution.md
```diff
@@ -359,17 +348,212 @@ Unsubscribe(responseStream):
...
+executed, returning the resulting {data} and {errors}.
+
+ExecuteRootSelectionSet(variableValues, initialValue, objectType, selectionSet,
+executionMode):
 
 - Let {groupedFieldSet} be the result of {CollectFields(objectType,
   selectionSet, variableValues)}.
+- Let {data} be the result of running {ExecuteGroupedFieldSet(groupedFieldSet,
+  objectType, initialValue, variableValues)} _serially_ if {executionMode} is
+  {"serial"}, otherwise _normally_).
```
benjie, line 381: I think we should restore the parenthesis, they add clarity to what "normally" means.

```suggestion
  {"serial"}, otherwise _normally_ (allowing parallelization)).
```

spec/Section 6 -- Execution.md
```diff
@@ -359,17 +348,212 @@ Unsubscribe(responseStream):
...
+
+### Field Collection
+
+Before execution, the _selection set_ is converted to a _grouped field set_ by
+calling {CollectFields()}. This ensures all fields with the same response name,
+including those in referenced fragments, are executed at the same time.
+
+:: A _grouped field set_ is a map where each entry is a _response name_ and a
+list of selected fields that share that _response name_ (the field alias if
+defined, otherwise the field's name).
```
benjie, line 394: If we choose not to rename "grouped field set", we should at least define it in terms of a set:

```suggestion
:: A _grouped field set_ is a map where each entry is a _response name_ and an
ordered set of selected fields that share that _response name_ (the field alias
if defined, otherwise the field's name).
```

spec/Section 6 -- Execution.md
```diff
@@ -359,17 +348,212 @@ Unsubscribe(responseStream):
...
+
+Note: The steps in {CollectFields()} evaluating the `@skip` and `@include`
+directives may be applied in either order since they apply commutatively.
+
+**Merging Selection Sets**
+
+When more than one field of the same name is executed in parallel, during value
+completion each related _selection set_ is collected together to produce a
+single _grouped field set_ in order to continue execution of the sub-selection
+sets.
```
benjie, line 500: Text like this already exists in the spec as it stands today, but it is confusing. The "in parallel" sounds like it applies to serial versus parallel execution, but it doesn't since this still holds for mutation fields. Further the "same name" is also not quite right, because it's the _response name_ that matters, not the actual field name - `{ cat: pet(id: 1) { name } dog: pet(id: 2) { age } }` would not merge selection sets.

The thing is, we don't execute _a field_ (in the GraphQL request document AST sense[^1] - i.e. a "field selection") we actually execute _a field set_ (i.e. a "set of field selections").

[^1]: We do execute a field in the GraphQL schema sense. Using "field" for both of these is quite confusing, I find. Maybe we should be more careful to use "field selection" rather than "field".

How about:

```suggestion
When a field is executed, during value completion the _selection set_ of each of
the related field selections with the same response name are collected together
to produce a single _grouped field set_ in order to continue execution of the
sub-selection sets.
```

benjie (commented):
Thanks for the edits @leebyron! I'm generally in favour, but have a few suggestions.

spec/Section 6 -- Execution.md
benjie: If we accept this, we should also make a few other changes such as:

```diff
-Initialize groupedFields to an empty ordered map of lists.
+Initialize groupedFields to an empty ordered map of ordered sets.
```

robrichard: There's no functional change between a list or a set in this case since the algorithm never attempts to add the same item to the set/list?

robrichard:
Worth mentioning here that the data structure returned by CollectFields will have to change to support `@defer`. It's currently modeled as:

```ts
type FieldDetails = { field: Field, deferUsage: DeferUsage }
type GroupedFieldSet = Map<string, List<FieldDetails>>
```

spec/Section 6 -- Execution.md
benjie: Yes indeed, list or ordered set is equivalent here since it will never include duplicates. The main reason to change it to set is because we call it a "grouped field set"; a "field set" being `OrderedSet<Field>`, and grouped being a map grouped by response key: `Map<string, OrderedSet<Field>>`.

benjie: This might be a simpler solve actually:

```suggestion
:: A _grouped field set_ is a map where each entry is a _response name_ and a
list of selected fields that share that _response name_ (the field alias if
defined, otherwise the field's name).

Note: No entry in the list of selected fields will be duplicated, hence we call
it a "field set" even though in the algorithms we use a list structure for it.
```

benjie (commented):
LGTM :+1:

benjie:
@leebyron To reflect your discussion above, I've renamed "ExecuteGroupedFieldSet" to "ExecuteCollectedFields" and that definitely flows better.

However, I think that "grouped field set" is still the right term to refer to the structure and actually defining a "field set" in general adds a lot of clarity to a number of algorithms (e.g. `ExecuteField()`) so I've both reworked the introduction to field collection in this PR and raised a separate PR that changes "grouped field set" to actually be `Map<string, OrderedSet<FieldSelection>>` rather than containing lists:

https://github.com/graphql/graphql-spec/pull/1161

This makes the algorithms a lot clearer to me:

```diff
-CollectSubfields(objectType, fields, variableValues):
+CollectSubfields(objectType, fieldSet, variableValues):
```
^ Knowing that a "field set" is a set of field selections that share the same response key makes this algorithms purpose a lot clearer

```diff
-ExecuteField(objectType, objectValue, fieldType, fields, variableValues):
+ExecuteField(objectType, objectValue, fieldType, fieldSet, variableValues):
```
^ A lot more intuitive what `fieldSet` is (field selections representing the field to be executed) rather than `fields` (could be interpreted as the the fields requested on this field's selection set).

```diff
-CompleteValue(fieldType, fields, result, variableValues):
+CompleteValue(fieldType, fieldSet, result, variableValues):
```
^ Same

spec/Section 6 -- Execution.md
```diff
@@ -359,17 +348,227 @@ Unsubscribe(responseStream):
...
+set via {CollectFields()} or from the selection sets of a _field set_ via
+{CollectSubfields()}.
+
+:: A _field set_ is a list of selected fields that share the same _response
+name_ (the field alias if defined, otherwise the field's name).
+
+Note: The order of field selections in a _field set_ is significant, hence the
+algorithms in this specification model it as a list. Any later duplicated field
+selections in a field set will not impact its interpretation, so using an
+ordered set would yield equivalent results.
```
benjie, line 404: # :new:

spec/Section 3 -- Type System.md
```diff
@@ -757,7 +757,7 @@ type Person {
 }
 ```
 
-Valid operations must supply a nested field set for any field that returns an
+Valid operations must supply a selection of fields for any field that returns an
```
martinbonnin, line 760: Not directly related to this PR but a more precise wording would be to use `selectionSet` here as fragments are also valid, not only fields.

```suggestion
Valid operations must supply a valid sub _selection set_ for any field that returns an
```

leebyron:
@benjie I made some substantial editorial changes here, especially to one of the term names. Please review!

Issue #53, opened 2015-07-14
Having both isDeprecated and deprecationReason is redundant


swolchok:
Just define it so that fields are deprecated iff they have a non-null `deprecationReason`.

leebyron:
The current reason for this is that `deprecationReason` is optional, regardless of if a field is deprecated or not, allowing a field to be tersely deprecated and avoiding less good cases like using empty strings to indicate deprecation or something along those lines.

I also felt like it was more clear to have the `isDeprecated` field, otherwise it seemed easy to misread the list of fields and assume deprecation was always possible.

If we decide to change this though, I would expect that decision to also include making `description` non-nullable. It's non-nullable for the same reason this semi-redundancy exists.

leebyron:
Closing this aging issue.

martinbonnin:
_necromancer_mode_on_

Can we revive this discussion? 

> allowing a field to be tersely deprecated and avoiding less good cases like using empty strings to indicate deprecation or something along those lines

Since the [current `@deprecated` directive definition](https://spec.graphql.org/draft/#sec--deprecated) has a default value, feels like terseness is not a problem anymore? One can always deprecate using just the directive:

```graphql
type Query {
  foo: Int @deprecated
}
```

We could make `reason` non-nullable in SDL:

```graphql
directive @deprecated(
  # note how reason is non-nullable here
  reason: String! = "No longer supported"
) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
```

This will also forbid such possibly confusing declarations:

```graphql
type Query {
  foo: Int @deprecated(reason: null) 
}
```

Introspection would keep nullable `deprecationReason`:

```graphql
type __Field {
  name: String!
  description: String
  args(includeDeprecated: Boolean = false): [__InputValue!]!
  type: __Type!
  # remove (or deprecated 😄  `isDeprecated`)
  # isDeprecated: Boolean!

  # keep the current behaviour for deprecationReason
  "the deprecation reason if the field is deprecated or null if the field is not deprecated"
  deprecationReason: String
}
```
Thoughts?

benjie:
@martinbonnin I've re-opened this discussion; but I think the correct way forward would be to add a PR that specifically changes the `reason` argument to be non-nullable (and nothing else).

PR #1040, opened 2023-08-23
Make the reason argument in `@deprecated` non-nullable

martinbonnin:
Follow up from https://github.com/graphql/graphql-spec/issues/53#issuecomment-1688335159

Make `reason` non-nullable:

```graphql
directive @deprecated(
  reason: String! = "No longer supported"
) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
```

This is technically a breaking change for someone that does this:

```graphql
type Foo {
  bar: String! @deprecated(reason: null)
}
```

But feels like this shouldn't be allowed in the first place? 

Fixes https://github.com/graphql/graphql-spec/issues/53

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -2099,7 +2099,7 @@ condition is false.
 
 ```graphql
 directive @deprecated(
-  reason: String = "No longer supported"
+  reason: String! = "No longer supported"
 ) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
 ```
 
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -110,7 +110,7 @@ CommonMark-compliant Markdown renderer.
 
 To support the management of backwards compatibility, GraphQL fields, arguments,
 input fields, and enum values can indicate whether or not they are deprecated
-(`isDeprecated: Boolean`) along with a description of why it is deprecated
+(`isDeprecated: Boolean!`) along with a description of why it is deprecated
 (`deprecationReason: String`).
 
 Tools built using GraphQL introspection should respect deprecation by
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -424,7 +424,8 @@ Fields\:
   this field.
 - `isDeprecated` returns {true} if this field should no longer be used,
   otherwise {false}.
-- `deprecationReason` optionally provides a reason why this field is deprecated.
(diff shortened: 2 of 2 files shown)


benjie (approved):

JoviDeCroock:
Partially related PR in GraphQL JS https://github.com/graphql/graphql-js/pull/4006 where the `@deprecated(reason: null)` currently doesn't work at all

martinbonnin:
Can this be moved to RFC1 or do we need anything else? 

Planning to work on a graphql-js implementation for the next wg.

benjie:
From the meeting:

- https://spec.graphql.org/draft/#sel-DAJXGTHLAAAEVBAA-7D (and similar locations throughout) < "deprecationReason: optionally provides a reason why this field is deprecated" should be updated - this is not optional now? (I didn't even realise we used "optional" when referring to output values - hence my confusion in yesterday's WG! In fact, these might be the only places we do?)
- If someone does `@deprecated` without a reason, how should this show up in SDL: with or without the default reason? What if they explicitly set the reason to exactly match the default reason?


Other sections of the spec will need to be addressed too, for example: https://spec.graphql.org/draft/#sel-FAJXLDCAACECx6V should actually have two bangs? Looks like the `isDeprecated: Boolean` is already incorrect there?

martinbonnin:
```
deprecationReason: optionally provides a reason why this field is deprecated
```

> this is not optional now?

I think it has to stay ~optional~ nullable for the cases where `isDeprecated` is `false`? If not what value do we put there?

> I didn't even realise we used "optional" when referring to output values - hence my confusion in yesterday's WG! In fact, these might be the only places we do?

Same! I find using optional for output positions is confusing. Proposal to change the wording to:

```
deprecationReason: the reason why the field is deprecated or `null` if the field is not deprecated.
```

We could even go as far as "deprecating `isDeprecated`" 🙃 :

```
  isDeprecated: Boolean! @deprecated(reason: "use deprecationReason != null instead")
  deprecationReason: String
```

benjie:
> it has to stay nullable

100% correct, sorry if I implied otherwise - it was not my intent.

> deprecationReason: the reason why the field is deprecated, or `null` if the field is not deprecated.

I added a comma, but yes looks good to me! (Follow the pattern of the surrounding text if you haven't already.)

> We could even go as far as "deprecating isDeprecated" 🙃 :

I think we should leave `isDeprecated` alone... for now :wink:  (But seriously, it's more efficient over the wire than an arbitrarily long string, so we shouldn't deprecate it.)

martinbonnin:
👍 Makes a lot of sense. And thanks for the comma! Will dive a bit more into this in the upcoming month

benjie (approved):

Issue #899, opened 2021-11-05
Question about BlockString term


sungam3r:
See http://spec.graphql.org/October2021/#sec-Descriptions : 
> To allow GraphQL service designers to easily publish documentation alongside the
capabilities of a GraphQL service, GraphQL descriptions are defined using the
Markdown syntax (as specified by [CommonMark](https://commonmark.org/)). In the
type system definition language, these description strings (often {BlockString})
occur immediately before the definition they describe.

{BlockString} is used like a reference to some term but I don't see its definition. I can find only single usage of `BlockString` through the spec. I see `BlockStringCharacter` term and `BlockStringValue` function. I suppose that `{BlockString}` term here is used by mistake and should be changed to just `block string`.

benjie:
Good question @sungam3r; it's a `StringValue` (http://spec.graphql.org/October2021/#StringValue) but that covers more than just the block string that the content is aiming for. I think it would probably be beneficial to alter the grammar to specifically call out `BlockString`; so:

```md
StringValue ::
  - `""` [lookahead != `"`]
  - `"` StringCharacter+ `"`
  - BlockString

BlockString :: `"""` BlockStringCharacter* `"""`
```

sungam3r:
It makes sense. Then I will use the `BlockString` term in error messages.

PR #1042, opened 2023-08-29
Add explicit definition for BlockString

benjie:
As pointed out by @sungam3r in #899, we state:

> In the type system definition language, these description strings (often {BlockString}) occur immediately before the definition they describe.

However, we don't have a definition for `BlockString`.

I've elevated `BlockString` to be its own entity, and `StringValue` now uses that definition instead.

GraphQL.js already has a BlockString token type: https://github.com/graphql/graphql-js/blob/b12dcffe83098922dcc6c0ec94eb6fc032bd9772/src/language/tokenKind.ts#L27 so I think this is just an edit to reflect reality and doesn't really require the RFC process.

@graphql/tsc Can I get some +1s? I'll also add it to the Sept 7 WG in case anyone feels we shouldn't do this.

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -100,7 +100,7 @@ StringValue ::
 
 - `""` [lookahead != `"`]
 - `"` StringCharacter+ `"`
-- `"""` BlockStringCharacter\* `"""`
+- BlockString
 
 StringCharacter ::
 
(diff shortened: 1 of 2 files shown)


fotoetienne (approved):
Makes sense

Shane32 (approved):

mjmahone (approved):

robrichard (approved):

calvincestari (approved):

martinbonnin (approved):

benjie:
With all these approvals I'm going to go ahead and merge this; we can always undo it if we need to :+1: 

Thanks everyone!

PR #466, opened 2018-06-13
 Add description to Schema

IvanGoncharov:
Fist comment is just put `Description` section before `Schema` section.
Changes to review: bd5cdd8

This PR adds description to schema both in SDL and introspection.

PR #1047, opened 2023-09-27
Relax non-inclusion of schema definition.

ddebrunner:
"should" seems to strong here given a schema can have a description and directives applied to it.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -221,7 +221,7 @@ _root operation type_ uses its respective _default root type name_ and no other
 type uses any _default root type name_.
 
 Likewise, when representing a GraphQL schema using the type system definition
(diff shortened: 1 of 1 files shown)


ddebrunner:
and no description ...

benjie:
We should keep the `should` for consistency - best that everyone generates the same SDL for the same schema where possible.

So I support @martinbonnin's proposal that we further narrow this case. I think the description was an oversight of https://github.com/graphql/graphql-spec/pull/466 - that's the main thing to include right now.

Note that there are no directives that would be visible on a `schema` keyword currently when the schema is printed (via introspection results) as SDL. The only directives currently included when representing a schema as SDL are `@deprecated` and `@specifiedBy`, though this may change when #300 is addressed. (This is not true for the _source_ SDL, but that's not what we're talking about here.)

PR #1048, opened 2023-10-04
RFC: Null-Only-On-Error / Semantically-Non-Null type (asterisk)

benjie:
This PR replaces #1046. Here's an updated description:

Inspired by @captbaritone's True Nullability Schema discussion (https://github.com/graphql/graphql-wg/discussions/1394) and following @fotoetienne's excellent talk at GraphQLConf and feedback on my resulting PR #1046, I am proposing that we introduce a new wrapper type, the "Null-Only-On-Error" type, represented via asterisk `*`. This new type walks the line between the default nullable types and the existing Non-Null type wrapper - it states that the value may only be {null} if a field error has been raised - i.e. it represents the "true nullability" of the field, whilst still acting as an error boundary to avoid null propagation.

Critically, this type would "evaporate" for legacy clients, appearing the same as a nullable field. (This is enabled via the `includeNullOnlyOnError` argument to the `__Field.type` field, which defaults to `false`.)

This is a much more complete RFC that #1046 was, and happily leverages a lot of the existing behavior of the spec.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -1239,6 +1239,11 @@ NonNullType :
 - NamedType !
 - ListType !
 
+NullOnlyOnErrorType :
+
+- NamedType \*
+- ListType \*
+
 GraphQL describes the types of data expected by arguments and variables. Input
 types may be lists of another input type, or a non-null variant of any other
 input type.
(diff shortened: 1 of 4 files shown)


benjie:
See also [directive for disabling null bubbling](https://github.com/graphql/graphql-spec/pull/1050) built on top of this.

benjie:
Closed in favour of https://github.com/graphql/graphql-spec/pull/1065

PR #1049, opened 2023-10-04
Introduce SourceSchemaDocument and FullSchemaDocument

martinbonnin:
Supersedes https://github.com/graphql/graphql-spec/pull/1036

See also https://github.com/graphql/graphql-wg/pull/1401 for the glossary.

This PR is a draft of what spec edits might look like to give a more precise idea of what it could end up like. Terminology and style to be fine tuned.

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -238,7 +238,12 @@ Directives[Const] : Directive[?Const]+
 
 Directive[Const] : @ Name Arguments[?Const]?
 
-TypeSystemDocument : TypeSystemDefinition+
+SourceSchemaDocument : TypeSystemDefinitionOrExtension+
+
+TypeSystemDefinitionOrExtension :
+
+- TypeSystemDefinition
+- TypeSystemExtension
 
 TypeSystemDefinition :
 
@@ -246,17 +251,12 @@ TypeSystemDefinition :
 - TypeDefinition
 - DirectiveDefinition
 
-TypeSystemExtensionDocument : TypeSystemDefinitionOrExtension+
-
-TypeSystemDefinitionOrExtension :
-
-- TypeSystemDefinition
-- TypeSystemExtension
-
 TypeSystemExtension :
 
 - SchemaExtension
 - TypeExtension
+
+FullSchemaDocument : TypeSystemDefinition+
 
 SchemaDefinition : Description? schema Directives[Const]? {
 RootOperationTypeDefinition+ }
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -242,16 +242,18 @@ Definition :
 - ExecutableDefinition
 - TypeSystemDefinitionOrExtension
 
+A GraphQL Document describes a complete file or request string operated on by a
+GraphQL service or client. A document contains multiple definitions, either
+executable or representative of a GraphQL type system.
+
+### Executable document
+
 ExecutableDocument : ExecutableDefinition+
 
 ExecutableDefinition :
 
 - OperationDefinition
 - FragmentDefinition
-
-A GraphQL Document describes a complete file or request string operated on by a
-GraphQL service or client. A document contains multiple definitions, either
-executable or representative of a GraphQL type system.
 
 Documents are only executable by a GraphQL service if they are
 {ExecutableDocument} and contain at least one {OperationDefinition}. A Document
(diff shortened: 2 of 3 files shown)


spec/Appendix B -- Grammar Summary.md
```diff
@@ -258,6 +260,8 @@ TypeSystemExtension :
 - SchemaExtension
 - TypeExtension
 
+FullSchemaDocument : TypeSystemDefinition+
```
mjmahone, line 263: Interesting that we are disallowing extensions in the FullSchemaDocument. I'm not opposed but this is interesting.

That feels a little odd, as the Relay Compiler tends to overload extension to mean client-defined (this is probably wrong and we should switch to using a directive).

spec/Section 2 -- Language.md
```diff
@@ -275,6 +277,90 @@ operations, each operation must be named. When submitting a Document with
...
+}
+```
+
+### Full schema document
+
+FullSchemaDocument : TypeSystemDefinition+
+
+A {FullSchemaDocument} describes all the types in the schema, including the
+built-in ones. Tools that are not implementations may use a {FullSchemaDocument}
+to validate an operation.
```
mjmahone, line 356: I'm not sure this singular purpose captures the depth of why FullSchemaDocument is useful. It's validation, but also code generation, IDE features, and more.

spec/Section 2 -- Language.md
```diff
@@ -275,6 +277,90 @@ operations, each operation must be named. When submitting a Document with
...
+type Mutation {
+  name: String
+}
+```
+
+### Full schema document
+
+FullSchemaDocument : TypeSystemDefinition+
+
+A {FullSchemaDocument} describes all the types in the schema, including the
```
mjmahone, line 354: "describes all the types in the schema" I'm not sure this is guaranteed to be true. It describes all the types that are visible to the consumer.

I know this is pedantic, but I'm currently using an equivalent of the "full schema document" in a way that says, for this library, what do we want to make available for codegen and validation, **even if other types/fields exist** and are exposed on the server.

spec/Section 2 -- Language.md
```diff
@@ -275,6 +277,90 @@ operations, each operation must be named. When submitting a Document with
...
+built-in ones. Tools that are not implementations may use a {FullSchemaDocument}
+to validate an operation.
+
+All _built-in definitions_ must be included in a {SourceSchemaDocument}.
```
mjmahone, line 358: FullSchemaDocument

Though I'm not sure this is true either. Any built-in definitions not included will be unavailable to tooling, and imply the schema **is not spec compliant** which is not the same as "must be included".

But for instance if there are no Float usages, does the full schema need to include the Float type?

spec/Section 2 -- Language.md
```diff
@@ -275,6 +277,90 @@ operations, each operation must be named. When submitting a Document with
...
+
+A {SourceSchemaDocument} must exactly one {SchemaDefinition}. That
```
mjmahone, line 360:
```suggestion
A {FullSchemaDocument} must include exactly one {SchemaDefinition}. That
```

spec/Section 3 -- Type System.md
```diff
@@ -139,6 +130,8 @@ All types and directives defined within a schema must not have a name which
 begins with {"\_\_"} (two underscores), as this is used exclusively by GraphQL's
 introspection system.
 
+A document must include at most one {SchemaDefinition}.
```
mjmahone, line 133:
```suggestion
A document may include at most one {SchemaDefinition}.
```

mjmahone (commented):
I am still a fan of this change

spec/Appendix B -- Grammar Summary.md
martinbonnin: I went with `no extension` by default to make a full schema as simple and as easy to read as possible. 

A "full schema" is probably machine written and probably machine consumed as well. It's the source of truth for a service. 

Another typical use case besides validation would be someone clicking their way in IntelliJ to find a type definition. I'd argue that having the extensions merged at that point would be nicer because you don't have to remember to "merge" things.

That being said, I think it'd be OK to "extend" a full schema:

`client full schema` = `server full schema` + `type extensions`

We'll probably end up encouraging this in Apollo Kotlin, possibly for stuff like `@nullOnlyOnError`

spec/Section 2 -- Language.md
martinbonnin: Fair point 👍 . I guess it depends how much we want to bind this to introspection but I think it'd have some value having the `FullSchemas` be a super set of the introspection schema.

spec/Section 2 -- Language.md
martinbonnin: 💯 I'll reword

spec/Section 2 -- Language.md
martinbonnin: > FullSchemaDocument

Indeed, thanks!

> Though I'm not sure this is true either. Any built-in definitions not included will be unavailable to tooling, and imply the schema is not spec compliant which is not the same as "must be included".

Reason I went with this requirement is we're plagued with duplicate directive definitions. Latest example to date is https://github.com/apollographql/apollo-ios/issues/3287. Right now, we've been adding directive definitions in order to validate documents but this really fragile to cases where the directive definitions do not match like this one.

> if there are no Float usages, does the full schema need to include the Float type?

Good point. I guess it's not really required. The rule would then be "all used built-in definitions must be included", which is more flexible but also more prone to interpretation... But could be cool indeed.

martinbonnin (commented):
Thanks for the review 💙

PR #1052, opened 2023-11-06
RFC: deduplicated incremental delivery

yaacovCR:
Iterating on #1034 and #1026, removed mutation of internal state, event stream management, and the need for subprocedures.

YieldSubsequentPayloads now is passed the entirety of the "state," pending futures, etc, monitors for any changes to pending futures, rebuilds a new "state" representation based on any changes, yields a single result as necessary, and then recursively calls itself to yield remaining results. 

[The diff to main might be helpful, but this is built on top of the amazing #742 and so the diff [from that branch could be more useful](https://github.com/robrichard/graphql-spec/compare/incremental..yaacovCR:graphql-spec:deduplicate3).]

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -4,6 +4,7 @@ ignoreRegExpList:
   - /[a-z]{2,}'s/
 words:
   # Terms of art
+  - deprioritization
   - endianness
   - interoperation
   - monospace
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1941,6 +1941,11 @@ by a validator, executor, or client tool such as a code generator.
 
 GraphQL implementations should provide the `@skip` and `@include` directives.
 
+GraphQL implementations are not required to implement the `@stream` directive.
+If the directive is implemented, it must be implemented according to this
+specification. GraphQL implementations that do not support the `@stream`
+directive must not make it available via introspection.
+
 GraphQL implementations that support the type system definition language must
 provide the `@deprecated` directive if representing deprecated portions of the
 schema.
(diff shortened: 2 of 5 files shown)


yaacovCR:
warning: has bugs, will change, hopefully will be fixed prior to next week's meeting. :)

yaacovCR:
extracted out #1054 from this PR to simply a bit

also extracted out a separate commit for deduplication of parent-child relationships, to be included within this PR

yaacovCR:
Conceptually, I've restructured it a bit so that the changes can be summarized as follows:
1. introduced an "Executing a Field Plan" layer above "Executing a SelectionSet/GroupedFieldSet" where the "Field Plan" refers to the original grouped field set vs. any possible new grouped field sets that need be executed separately (whether they require additional deferral or not)*
2. {CollectFields()} is changed to:
(a) {AnalyzeSelectionSet()} which is responsible for extracting the data from visited fields, new deferred fragments
(b) {GetGroupedFieldSets()} which determines the way in which any new deferred grouped field sets must be created, which can be performed prior to execution, same across all members of a list(c) and {GetFieldPlan} which combines the results of these calls.
3. During execution, actual concrete new deferred fragments and incremental data records are created within {GetPathDependentFieldPlan()}
4. introduced the {YieldSubsequentPayloads()} algorithm that updates the "incrementalState" internal just to this algorithm so that no illegal mutation is performed
5. rebased the PR on top of Benjie's https://github.com/graphql/graphql-spec/pull/1039 which changes ExecuteSelectionSet => ExecuteGroupedFieldSet
6. separated out the necessary edits from @stream and @defer into separate commits, which I think considerably helps with review of the Execution section.
7. squashed the considerable hx from https://github.com/graphql/graphql-spec/pull/742 to aid in above

yaacovCR:
Updated to reflect simplification from https://github.com/graphql/graphql-js/pull/3994

yaacovCR:
closing in favor of #1077, although that PR still describes only algorithmic changes, and additional work from #742 and this PR would have to be pulled in if the approach is adopted.

PR #1053, opened 2023-11-09
Implementations may not deprecate a field that the interface hasn't deprecated

benjie:
This PR addresses a spec validation omission; if an interface field is not deprecated then any implementation of that interface field should also not be deprecated. I.e. the following schema _should_ be invalid, but before this PR it is valid:

```graphql
interface Node {
  id: ID!
}

type Foo implements Node {
  id: ID! @deprecated(reason: "...")
}

type Query {
  foo: Foo
}
```

To solve this, either the deprecation should be removed, or the `id` field on `Node` should _also_ be deprecated:

```graphql
interface Node {
  id: ID! @deprecated(reason: "...")
}
```

Relevant action item: https://github.com/graphql/graphql-wg/issues/1331

GraphQL.js Implementation: https://github.com/graphql/graphql-js/pull/3986

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -947,6 +947,8 @@ IsValidImplementation(type, implementedType):
        2. Let {implementedFieldType} be the return type of {implementedField}.
        3. {IsValidImplementationFieldType(fieldType, implementedFieldType)} must
           be {true}.
+    6. If the `@deprecated` directive is not applied to {implementedField}, then
+       the `@deprecated` directive must not be applied to {field}.
 
 IsValidImplementationFieldType(fieldType, implementedFieldType):
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -947,8 +947,7 @@ IsValidImplementation(type, implementedType):
        2. Let {implementedFieldType} be the return type of {implementedField}.
        3. {IsValidImplementationFieldType(fieldType, implementedFieldType)} must
           be {true}.
-    6. If the `@deprecated` directive is not applied to {implementedField}, then
-       the `@deprecated` directive must not be applied to {field}.
+    6. If {field} is deprecated then {implementedField} must also be deprecated.
 
 IsValidImplementationFieldType(fieldType, implementedFieldType):
 


fotoetienne:
Should the graphql-js impl be part of a major release?

mjmahone (approved):

mjmahone:
One potential wrinkle to this: we don't really have deprecated interface implementations.

Like, maybe I'm deprecating a field because I will *not* be implementing an interface in the future. What I'm really trying to say is:
```
interface Node {
  id: ID!
}

type Foo implements Node @deprecated(reason: "Node is terrible for this purpose") {
  id: ID! 
}

type Query {
  foo: Foo
}
```

spec/Section 3 -- Type System.md
```diff
@@ -947,6 +947,8 @@ IsValidImplementation(type, implementedType):
        2. Let {implementedFieldType} be the return type of {implementedField}.
        3. {IsValidImplementationFieldType(fieldType, implementedFieldType)} must
           be {true}.
+    6. If the `@deprecated` directive is not applied to {implementedField}, then
+       the `@deprecated` directive must not be applied to {field}.
```
leebyron, line 951:
```suggestion
    6. If {field} is deprecated then {implementedField} must also be deprecated.
```

leebyron (approved):
suggest rewriting to:

- avoid directly referencing directives, but just the deprecated status directly.
- if / then to avoid the negative

michaelstaib (approved):

fotoetienne:
> maybe I'm deprecating a field because I will not be implementing an interface in the future.

I think @mjmahone has a valid point. Perhaps we don't want clients to do this anymore:

```graphql
... on Foo {
    id
}
```

but this is still ok
```graphql
... on Node {
    id
}
```

Marking `id` on foo as deprecated would indicate this.

fotoetienne (changes requested):
On further reflection, I'm thinking that this validation isn't needed. It doesn't bring enough benefit to warrant removing some deprecation capabilities.

benjie:
@fotoetienne I think the general intent of deprecation (and specifically the `includeDeprecated` feature in introspection) is that you should be able to represent a schema with all deprecated entities removed and that schema should still be valid. Without this rule, this is not the case - if you remove the deprecated fields then `Foo` no longer implements `Node`, and the "not deprecated" schema is now invalid.

If and when we support deprecation on interface implementation as Matt suggest, this rule can be changed such that _either_ the interface field is deprecated _or_ the implementation of this interface is deprecated.

benjie:
Discussed at tonights WG; we boiled it down to this:

**If you remove everything deprecated from a GraphQL schema, should the resulting schema be valid?**

Consensus was yes, @michaelstaib expressed it as something along the lines of: if you're building a new client to a GraphQL schema, you don't want to use any of the deprecated fields so you should be able to just use a schema with everything deprecated removed.

Do you agree with this guiding principle to deprecation, @fotoetienne?

If so, but you feel this is removing the deprecation capability of deprecating a field where you plan to remove the type from an interface, then perhaps we need to pair this with a spec change that enables deprecation on object type interface implementation?

fotoetienne:
👍That makes sense. Once we support object/interface deprecation, we can revisit the use case.

JoviDeCroock (approved):

benjie:
It breaks the guiding principle for deprecation which is that if you remove everything that is deprecated then the resulting schema should still be valid. Implementing this rule should not break any client applications, and is easy to address within the schema by either deprecating the interface field or un-deprecating the object field depending on your intent.

martinbonnin:
Principles are good of course but I also like it when changes are rooted in real life use cases, helps making a case that the principles are still up to date, solving concrete issues, etc... 
In all cases, LGTM, was mostly curious.

BoD (approved):

mjmahone (approved):

PR #1054, opened 2023-11-09
deferred fragments should not change ordering of fields within initial response

yaacovCR:
builds on #1052


yaacovCR:
Requires rebasing with the changes to #1052 => and confirmation from the WG that this would be the desired behavior!

yaacovCR:
feedback i got offline (was it at graphql conf 2024?) was that it is fine if it changes the ordering, as long as it does in a predictable way.

PR #1055, opened 2023-11-09
Fix heading level for Required Arguments validation rule

goto-bus-stop:
Currently in the rendered spec, Required Arguments shows up as a sub-rule of "Argument Uniqueness". It's an independent rule so it should show up next to it instead.
image


benjie (approved):
Thanks for spotting!

benjie:
(You need to sign the CLA before we can merge this fix.)

benjie:
Congrats on your first merge; and thanks for the fix! :raised_hands:

PR #1058, opened 2023-11-09
[RFC] List coercion algorithm

### LINKED ISSUE #1002: Field error from list arg with nullable variable entry (nullable=optional clash?)
Consider the following _currently valid_ schema and query:

```graphql
type Query {
  sum(numbers:[Int!]!): Int
}

query Q ($number: Int = 3) {
  sum(numbers: [1, $number, 3])
}
```

With the following variables:

```json
{
  "number": null
}
```

Now following through [6.1.2 Coercing Variable Values](https://spec.graphql.org/draft/#sec-Coercing-Variable-Values):

1. Let coercedValues be an empty unordered Map. ✅ 
2. Let variablesDefinition be the variables defined by operation. ✅ 
3. For each variableDefinition in variablesDefinition: ✅ 
   1. Let variableName be the name of variableDefinition. ✅ 
   2. Let variableType be the expected type of variableDefinition. ✅ 
   3. Assert: [IsInputType](https://spec.graphql.org/draft/#IsInputType())(variableType) must be true. ✅ 
   4. Let defaultValue be the default value for variableDefinition. ✅ 
   5. Let hasValue be true if variableValues provides a value for the name variableName. ✅ 
   6. Let value be the value provided in variableValues for the name variableName. ✅ 
   7. If hasValue is not true and defaultValue exists (including null): ✅ 
      1. Add an entry to coercedValues named variableName with the value defaultValue. ✅ 
   8. Otherwise if variableType is a Non-Nullable type, and either hasValue is not true or value is null, raise a [request error](https://spec.graphql.org/draft/#request-error). ✅ 
   9. Otherwise if hasValue is true: ✅ 
      1. If value is null: ✅ 
         1. Add an entry to coercedValues named variableName with the value null. ✅

For the variable definition `$number: Int = 3`:

- `variableType` is `Int` (NOT `Int!`).
- `defaultValue` is `3`
- `hasValue` is `true` since `number` is provided in variables (even though it's null)
- `value` is `null`

Thus `coercedValues` becomes `{ number: null }`.

When it comes to executing the field, this results in [CoerceArgumentValues()](https://spec.graphql.org/draft/#CoerceArgumentValues()) raising a field error at 5.j.iii.1 (since `[1, null , 3]` cannot be coerced to `[Int!]!`).

Had the query have been defined with `($number: Int! = 3)` then this error could not have occurred; but we explicitly allow the nullable Int with default in [IsVariableUsageAllowed()](https://spec.graphql.org/draft/#IsVariableUsageAllowed()). Presumably this is so that we can allow a default value to apply in the list, except this is a `ListValue` so there cannot be a default value there.

We've discussed "optionality" versus "nullability" a few times, but I'd like to get the WG's read on this.

Reproduction:

```js
const { graphqlSync, GraphQLSchema, GraphQLList, GraphQLNonNull, GraphQLInt, GraphQLObjectType, validateSchema } = require('graphql');

  name: 'Query',
  fields: {
    sum: {
      type: GraphQLInt,
      args: {
        numbers: {
          type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLInt))),
        }
      },
      resolve(_, args) {
        let total = 0;
        for (const n of args.numbers) {
          total += n;
        }
        return total;
      }
    }
  }
});

  query: Query
});

if (errors.length > 0) {
  throw errors[0];
}

{
  console.log('Testing query with variables and null');
  const result = graphqlSync({
    schema,
    source: `query Q($number: Int = 3) {sum(numbers:[1,$number,3])}`,
    variableValues: {
      number: null
    }
  });
  console.dir(result);
}
```

benjie:
Fixes #1002.

Previously, list coercion does not detail what to do with variables at all, and that could lead to either a null pointer exception, or to double-coercion of the variable value if you're only following the spec.

Consider the following valid schema:

```graphql
type Query {
  sum(numbers:[Int!]!): Int
}
```

and the query that is valid against this schema:

```graphql
query Q ($number: Int = 3) {
  sum(numbers: [1, $number, 3])
}
```

**NOTE**: We're using the variable in a list item position!

If you issue this to the GraphQL server with variables `{"number": null}` then `CoerceVariableValues` will give you `{"number": null}` and when you fast-forward to `CoerceArgumentValues` you'll go in to 5.j.iii.1:

https://spec.graphql.org/draft/#sel-NANTHHCJFTDFBBCAACGB0yS

- Let {coercedValues} be an empty unordered Map. `coercedValues = {}`
- Let {argumentValues} be the argument values provided in {field}. `argumentValues = { numbers: [1, $number, 3] }`
- Let {fieldName} be the name of {field}. `fieldName = 'sum'`
- Let {argumentDefinitions} be the arguments defined by {objectType} for the
  field named {fieldName}. `argumentDefinitions = { numbers: ... }`
- For each {argumentDefinition} in {argumentDefinitions}:
  - Let {argumentName} be the name of {argumentDefinition}. `argumentName = 'numbers'`
  - Let {argumentType} be the expected type of {argumentDefinition}. `argumentType = [Int!]!`
  - Let {defaultValue} be the default value for {argumentDefinition}. `defaultValue = undefined`
  - Let {hasValue} be {true} if {argumentValues} provides a value for the name
    {argumentName}. `hasValue = true`
  - Let {argumentValue} be the value provided in {argumentValues} for the name
    {argumentName}. `argumentValue = [1, $number, 3]`
  - If {argumentValue} is a {Variable}: `NOPE`
    - Let {variableName} be the name of {argumentValue}.
    - Let {hasValue} be {true} if {variableValues} provides a value for the name
      {variableName}.
    - Let {value} be the value provided in {variableValues} for the name
      {variableName}.
  - Otherwise, let {value} be {argumentValue}. `value = [1, $number, 3]`
  - If {hasValue} is not {true} and {defaultValue} exists (including {null}):  `NOT TRIGGERED`
    - Add an entry to {coercedValues} named {argumentName} with the value
      {defaultValue}.
  - Otherwise if {argumentType} is a Non-Nullable type, and either {hasValue} is
    not {true} or {value} is {null}, raise a _field error_.  `NOT TRIGGERED`
  - Otherwise if {hasValue} is true: `Yes, it is`
    - If {value} is {null}: `It is not, it is a list`
      - Add an entry to {coercedValues} named {argumentName} with the value
        {null}.
    - Otherwise, if {argumentValue} is a {Variable}: `It is not, it is a list`
      - Add an entry to {coercedValues} named {argumentName} with the value
        {value}.
    - Otherwise: `YES`
      - If {value} cannot be coerced according to the input coercion rules of
        {argumentType}, raise a _field error_. **TIME TO VISIT LIST COERCION**
      - Let {coercedValue} be the result of coercing {value} according to the
        input coercion rules of {argumentType}.
      - Add an entry to {coercedValues} named {argumentName} with the value
        {coercedValue}.
- Return {coercedValues}.

## Time to visit [list coercion](https://spec.graphql.org/draft/#sec-List)

We need to coerce the value `[1, $number, 3]` to the non-nullable type `[Int!]!`.

Step 1: handle the non-null. It's not null. Great!

Now we need to coerce the value `[1, $number, 3]` to the list type `[Int!]`.

Here's what the spec says about input coercion for lists:

> **When expected as an input, list values are accepted only when each item in the list can be accepted by the list’s item type.**
>
> If the value passed as an input to a list type is not a list and not the null value, then the result of input coercion is a list of size one, where the single item value is the result of input coercion for the list’s item type on the provided value (note this may apply recursively for nested lists).
>
> This allows inputs which accept one or many arguments (sometimes referred to as “var args”) to declare their input type as a list while for the common case of a single value, a client can just pass that value directly rather than constructing the list.

We have a list, so we only care about the bold line.

This line seems to miss a bunch of situations.

For example: if we were coercing to `[Int]` the value `[1, $number, 3]` with variables `{}` then _is_ $number (which is undefined, since it wasn't provided in the variables) "accepted by the list's item type"? Really we must coerce this to `null`, but that doesn't seem to be detailed. In fact this entire section doesn't mention variables at all.

We're actually coercing to `[Int!]`, so the question is: is `$number` accepted by the list's item type? `$number` itself is a variable, so...

---

I've attempted to solve this problem by being much more explicit about the input coercion for lists, inspired by the input coercion for input objects. I've also added a non-normative note highlighting the risk of a null variable being fed through into a non-nullable position, why that can occur (validation) and what we do about it (field error). I've also expanded the table with both variables and many more examples to cover many more edge cases.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1780,7 +1780,9 @@ Following are examples of input coercion with various list types and values:
 | `[Int]`       | `1`              | `[1]`                       |
 | `[Int]`       | `null`           | `null`                      |
 | `[[Int]]`     | `[[1], [2, 3]]`  | `[[1], [2, 3]]`             |
-| `[[Int]]`     | `[1, 2, 3]`      | Error: Incorrect item value |
+| `[[Int]]`     | `[1, 2, 3]`      | `[[1], [2], [3]]`           |
+| `[[Int]]`     | `[1, null, 3]`   | `[[1], null, [3]]`          |
+| `[[Int]]`     | `[[1], ["b"]]`   | Error: Incorrect item value |
 | `[[Int]]`     | `1`              | `[[1]]`                     |
 | `[[Int]]`     | `null`           | `null`                      |
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1792,13 +1792,25 @@ CoerceListItemValue(itemValue, itemType):
 
 - If {itemValue} is {null}, return {null}.
 - Otherwise, if {itemValue} is a Variable:
-  - Let {runtimeValue} be the runtime value of that variable, or {null} if no
-    runtime value is provided.
-  - If {runtimeValue} is {null} and {itemType} is a non-null type, a _field
+  - If the variable provides a runtime value:
+    - Let {coercedItemValue} be the runtime value of the variable.
+  - Otherwise, if the variable definition provides a default value:
+    - Let {coercedItemValue} be this default value.
+  - Otherwise:
+    - Let {coercedItemValue} be {null}.
+  - If {coercedItemValue} is {null} and {itemType} is a non-null type, a _field
     error_ must be raised.
-  - Return {runtimeValue}.
+  - Return {coercedItemValue}.
 - Otherwise, return the result of coercing {itemValue} according to the input
   coercion rules for {itemType}.
+
+Note: When a default value exists for a variable definition, the type of the
+variable is allowed to be nullable even if it is used in a non-nullable
+position, see
+[Allowing Optional Variables When Default Values Exist](#sec-All-Variable-Usages-Are-Allowed.Allowing-Optional-Variables-When-Default-Values-Exist)
+in Validation. If the value for such a variable is explicitly {null} and is used
+as the value for a list item of non-nullable type then a _field error_ will be
+raised.
 
 Following are examples of input coercion with various list types and values:
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1773,18 +1773,18 @@ constructing the list.
 
 Following are examples of input coercion with various list types and values:
 
-| Expected Type | Provided Value   | Coerced Value               |
-| ------------- | ---------------- | --------------------------- |
-| `[Int]`       | `[1, 2, 3]`      | `[1, 2, 3]`                 |
-| `[Int]`       | `[1, "b", true]` | Error: Incorrect item value |
-| `[Int]`       | `1`              | `[1]`                       |
-| `[Int]`       | `null`           | `null`                      |
-| `[[Int]]`     | `[[1], [2, 3]]`  | `[[1], [2, 3]]`             |
-| `[[Int]]`     | `[1, 2, 3]`      | `[[1], [2], [3]]`           |
-| `[[Int]]`     | `[1, null, 3]`   | `[[1], null, [3]]`          |
-| `[[Int]]`     | `[[1], ["b"]]`   | Error: Incorrect item value |
-| `[[Int]]`     | `1`              | `[[1]]`                     |
-| `[[Int]]`     | `null`           | `null`                      |
+| Expected Type | Literal Value    | Variable Values | Coerced Value               |
+| ------------- | ---------------- | --------------- | --------------------------- |
+| `[Int]`       | `[1, 2, 3]`      | `{}`            | `[1, 2, 3]`                 |
+| `[Int]`       | `[1, "b", true]` | `{}`            | Error: Incorrect item value |
+| `[Int]`       | `1`              | `{}`            | `[1]`                       |
+| `[Int]`       | `null`           | `{}`            | `null`                      |
+| `[[Int]]`     | `[[1], [2, 3]]`  | `{}`            | `[[1], [2, 3]]`             |
+| `[[Int]]`     | `[1, 2, 3]`      | `{}`            | `[[1], [2], [3]]`           |
+| `[[Int]]`     | `[1, null, 3]`   | `{}`            | `[[1], null, [3]]`          |
+| `[[Int]]`     | `[[1], ["b"]]`   | `{}`            | Error: Incorrect item value |
+| `[[Int]]`     | `1`              | `{}`            | `[[1]]`                     |
+| `[[Int]]`     | `null`           | `{}`            | `null`                      |
 
 ## Non-Null
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1771,20 +1771,69 @@ This allows inputs which accept one or many arguments (sometimes referred to as
 single value, a client can just pass that value directly rather than
 constructing the list.
 
+The result of coercion of a value {value} to a list type {listType} is
+{CoerceListValue(value, listType)}.
+
+CoerceListValue(value, listType):
+
+- If {value} is {null}, return {null}.
+- Let {itemType} be the inner type of {listType}.
+- Let {coercedList} be an empty list.
+- If {value} is a list:
+  - For each {itemValue} in {value}:
+    - Let {coercedItemValue} be {CoerceListItemValue(itemValue, itemType)}.
+    - Append {coercedItemValue} to {coercedList}.
+- Otherwise:
+  - Let {coercedItemValue} be {CoerceListItemValue(value, itemType)}.
+  - Append {coercedItemValue} to {coercedList}.
+- Return {coercedList}.
+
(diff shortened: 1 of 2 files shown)


mjmahone (approved):
This feels like it brings list coercion to the same (currently buggy due to nullable-with-default-values) state as all the other variable coercion so this feels right to me

spec/Section 3 -- Type System.md
```diff
@@ -1771,18 +1771,82 @@ This allows inputs which accept one or many arguments (sometimes referred to as
...
+- Return {coercedList}.
+
+CoerceListItemValue(itemValue, itemType):
+
+- If {itemValue} is {null}, return {null}.
+- Otherwise, if {itemValue} is a Variable:
+  - If the variable provides a runtime value:
+    - Let {coercedItemValue} be the runtime value of the variable.
+  - Otherwise, if the variable definition provides a default value:
+    - Let {coercedItemValue} be this default value.
```
martinbonnin, line 1798: Is there a specific reason this doesn't use the "pre-coerced" variable values?

benjie: The reason was that this is essentially copied from the input coercion for input objects (but via an algorithm to make it clearer): https://spec.graphql.org/draft/#sec-Input-Objects.Input-Coercion

But it's a good question. I guess the reason is that `coercedVariableValues` is not explicitly made available in section 3 of the spec.

martinbonnin: I see... It may be an issue:

Ultra synthetic example:

```graphql
type Query {
  a(b: [[Int]]): Int
}
```

Operation:

```graphql
query Foo($c: [Int]) {
  a(b: [$c])
}
```

Runtime Variables:
```json
{
  "c": 42
}
```

If we're saying the runtime value is the "value that is sent over the wire", we end up with `b = [42]` (incompatible) instead of `b = [[42]]` if we coerce the variable to a list first (I think?)
If we're saying the runtime value is the "pre-coerced" value, then there's no need to mention defaultValue

JoviDeCroock (approved):
This one is also in line with GraphQL JS we deeply traverse and perform the same logic as during normal coercion on each value we find deeply

spec/Section 3 -- Type System.md
JoviDeCroock: That should work in GraphQL-JS as well from what I can tell

- We see that `b` expects a List https://github.com/graphql/graphql-js/blob/16.x.x/src/utilities/valueFromAST.ts#L77
- We see that `b` gets a list as an AST-Argument https://github.com/graphql/graphql-js/blob/16.x.x/src/utilities/valueFromAST.ts#L79
- We go into the for loop, the first element is a variable, we recurse https://github.com/graphql/graphql-js/blob/16.x.x/src/utilities/valueFromAST.ts#L90
- Inside of the recursion we see that this value is expected to be an array again https://github.com/graphql/graphql-js/blob/16.x.x/src/utilities/valueFromAST.ts#L77
- The received value is not a list, but a variable, we get the variable and return it in an array https://github.com/graphql/graphql-js/blob/16.x.x/src/utilities/valueFromAST.ts#L99-L103

JoviDeCroock: That understanding is not what I'm saying though, notice how we return [result] rather than result? Meaning we do coerce into a list as that's the expected type of the argument, regardless of the variable type, so the variable coerced or not seems irrelevant

martinbonnin: On line 103 there's coercion indeed:

```js
  if (isListType(type)) {
    // ....
    return [coercedValue];
  } 
```

But my understanding is that line 103 is not reached because we should enter the branch in line 49 first?:
```js
  if (valueNode.kind === Kind.VARIABLE) {
    // ..
    return variableValue;
  }
```

Not super familiar with the codebase so apologies if that's absolutely not how it works but it _feels_ like it should work like so?

benjie: This is a really good catch @martinbonnin. The issue here is I copied the phrasing from the input objects section which made a similar mistake. To address this I've pushed up a fix that:

1. defines the term "coerced runtime value" for a variable
2. uses this term instead of "runtime value"
3. removes the redundant text about applying the variable's default (since that will already have been handled as you point out)
4. applies similar fixes to input object coercion
5. changes some incorrect `should` to `must` in the input object prose

If we were to reference `variableValues` in section 3 explicitly, we should do that throughout section 3 I think. I've avoided that for now.

martinbonnin: > If we were to reference variableValues in section 3 explicitly, we should do that throughout section 3 I think. I've avoided that for now.

Fair enough 👍 I think the spec would gain in clarity in doing so but agree this is a separate topic.

leebyron:
@benjie here's my alternative proposal draft as discussed in today's WG meeting: https://github.com/graphql/graphql-spec/pull/1179

That counter-proposal is more of a superset that includes input object coercion, but also I feel makes the list coercion easier to follow since null and variable checking is handled elsewhere

Issue #1062, opened 2023-11-20
On specifying ordered vs unordered enum definitions


cdaringe:
# Problem

There is no specification on whether [EnumTypeDefinition](https://spec.graphql.org/draft/#EnumTypeDefinition) are sorted by `Name`, and it causes friction for GQL clients who presume there is order.

# Context

- [EnumTypeDefinition](https://spec.graphql.org/draft/#EnumTypeDefinition) makes no indication of ordering
- Ordering very is significant to clients who convert the stringly-GQL DSL to typed client SDKs
  - example 1 (typescript): https://github.com/dotansimha/graphql-code-generator/issues/6935#issuecomment-1474084101
  - example 2 (ios/apollo gen):
     - given an enum `enum { Xx Zz Yy }` 
        - enum => union: `union Foo = Xx | Zz | Yy`
        - struct: `public struct AsXxOrZzOrYy: GraphQLSelectionSet {`

As observed from the above, ordering matters during codegen to GQL consumers. If a schema maintainer changed the enum to `enum { Yy Xx Zz }`, some clients would see this as a breaking change in the status quo.

# Discussion

- To help consumers/clients of GQL succeed, the spec should disambiguate ordering on `EnumTypeDefinition`.

Personal recommendation: `enum`s and tagged unions alike are `Set`s, thus intrinsically unordered. We should state as much. Equipped with this assertion, client libraries whom seek to do type generation should apply stable sorting.

benjie:
Good catch. In my opinion, like fields, they should be in the order in which they were defined, which effectively means tools are free to make their own decisions. Either way, the order should be stable in my opinion:

1. introspect a GraphQL schema
2. build a schema from these introspection results
3. introspect this new schema using the same introspection query

The results of 1 and 3 should be identical.

cdaringe:
@benjie! long time, no-GitHub-see :)

> Either way, the order should be stable in my opinion ...

Just to clarify, are you asserting that you think, in the spec, `EnumTypeDefinition` should specify that entries are ordered?

martinbonnin:
We've been [occasionaly hit](https://github.com/apollographql/apollo-kotlin/issues/4659) by such sorting issues in Apollo Kotlin and the more I think about it the more I think the "good" order for tools is the schema order. Everything else is surprising and error prone. 

[...]

PR #1063, opened 2023-11-21
[RFC] Type system ordering of: object interfaces, directive arguments, input object fields, enum values

benjie:
**UPDATE 2025-06-26**: The motivation of this PR is **to make introspection stable** (and, related, to make `printSchema()` stable, and to make `printSchema(buildSchema(printSchema(schema))) === printSchema(schema)`).

**UPDATE 2024-04-05**: Following [last night's WG meeting](https://github.com/graphql/graphql-wg/blob/main/notes/2024/2024-04.md#ordering-of-schema-elements-10m-benjie), this PR has been split into 3 parts, and is now PR 3 of 3 in the "data collections" trilogy stack

1. #1102
2. #1092
3. #1063 (this one)

---

As raised by @cdaringe in #1062, enum values don't dictate an explicit order. On scanning through the spec in more detail, it turns out this is true of a few other things too.

Everything in SDL is implicitly ordered (since it's a textual representation, one thing after another) and everything in introspection is implicitly ordered (because it's represented via lists). E.g. in introspection, [`enumValues` is a list](https://spec.graphql.org/draft/#sec-The-__Type-Type.Enum), and a list is inherently ordered.

I feel it's an unwritten rule that GraphQL introspection should be stable (i.e. introspect the exact same schema twice with the exact same introspection query and the results should be the same). Thus, there should be an order (dictated by the schema designer), and I'd like to make that more explicit.

I researched the current status, and I think we can start to fix this with the few minor edits I made to the spec in this PR, in particular:

1. changing from using the word `set` (which is _generally_ perceived as unordered) to the word `list` (which is always ordered),
2. specifying list for things that are currently ambiguous.

Generally this was achieved by copying text from similar things, e.g. the directive arguments copied from field arguments; input object fields copied from object fields.

I know that @IvanGoncharov has been very careful in graphql-js to ensure that ordering is stable, I believe he ensures that introspection -> SDL -> introspection always results in the same results.

# Status before this PR (emphasis mine)

## Object fields: ordered :white_check_mark: 

> GraphQL Objects represent a **list** of named fields, each of which yield a value of a specific type.

## Object field arguments: ordered :white_check_mark: 

> Object field arguments are defined as a **list** of all possible argument names and their expected input types

## Object interfaces: not declared? :disappointed: 

> An object type may declare that it implements one or more unique interfaces.

(I couldn't find anything in Section 3 declaring set/list.)

## Input object fields: a "set" :x: 

> A GraphQL Input Object defines a **set** of input fields

## Enum values: a "set" :x: 

> However Enum types describe the **set** of possible values

## Directive arguments: not declared? :disappointed: 

(I couldn't find anything in Section 3 declaring set/list.)

### DIFF
--- a/spec/GraphQL.md
+++ b/spec/GraphQL.md
@@ -92,10 +92,14 @@ Conformance requirements expressed as algorithms can be fulfilled by an
 implementation of this specification in any way as long as the perceived result
 is equivalent. Algorithms described in this document are written to be easy to
 understand. Implementers are encouraged to include equivalent but optimized
-implementations.
+implementations. Similarly, data collections such as _list_, _set_ and _map_
+also introduce conformance requirements. Implementers are free to use
+alternative data collections as long as the perceived result remains equivalent.
 
 See [Appendix A](#sec-Appendix-Notation-Conventions) for more details about the
-definition of algorithms and other notational conventions used in this document.
+definition of algorithms and other notational conventions used in this document,
+and [Appendix A: Data Collections](#sec-Data-Collections) for specifics of data
+collections and their ordering.
 
 **Non-Normative Portions**
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -921,8 +921,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          returns {true}.
       4. If argument type is Non-Null and a default value is not defined:
          1. The `@deprecated` directive must not be applied to this argument.
-3. An object type may declare that it implements an ordered set of one or more
-   unique interfaces.
+3. An object type may declare that it implements one or more unique interfaces,
+   these interfaces form an ordered set.
 4. An object type must be a super-set of all interfaces it implements:
    1. Let this object type be {objectType}.
    2. For each interface declared implemented as {interfaceType},
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -921,8 +921,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          returns {true}.
       4. If argument type is Non-Null and a default value is not defined:
          1. The `@deprecated` directive must not be applied to this argument.
-3. An object type may declare that it implements a set of one or more unique
-   interfaces.
+3. An object type may declare that it implements an ordered set of one or more
+   unique interfaces.
 4. An object type must be a super-set of all interfaces it implements:
    1. Let this object type be {objectType}.
    2. For each interface declared implemented as {interfaceType},
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -310,8 +310,8 @@ GraphQL supports two abstract types: interfaces and unions.
 
 An `Interface` defines an ordered set of fields; `Object` types and other
 Interface types which implement this Interface are guaranteed to implement those
-fields. Whenever a field claims it will return an Interface type, it will return
-a valid implementing Object type during execution.
+fields (in any order). Whenever a field claims it will return an Interface type,
+it will return a valid implementing Object type during execution.
 
 A `Union` defines a set of possible types; similar to interfaces, whenever the
 type system claims a union will be returned, one of the possible types will be
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1350,10 +1350,10 @@ UnionMemberTypes :
 - UnionMemberTypes | NamedType
 - = `|`? NamedType
 
-GraphQL Unions represent an object that could be one of a set of GraphQL Object
-types, but provides for no guaranteed fields between those types. They also
-differ from interfaces in that Object types declare what interfaces they
-implement, but are not aware of what unions contain them.
+GraphQL Unions represent an object that could be one of an ordered set of
+GraphQL Object types, but provides for no guaranteed fields between those types.
+They also differ from interfaces in that Object types declare what interfaces
+they implement, but are not aware of what unions contain them.
 
 With interfaces and objects, only those fields defined on the type can be
 queried directly; to query other fields on an interface, typed fragments must be
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -302,17 +302,18 @@ enumerable. GraphQL offers an `Enum` type in those cases, where the type
 specifies the space of valid responses.
 
 Scalars and Enums form the leaves in response trees; the intermediate levels are
-`Object` types, which define a set of fields, where each field is another type
-in the system, allowing the definition of arbitrary type hierarchies.
+`Object` types, which define an ordered set of fields, where each field is
+another type in the system, allowing the definition of arbitrary type
+hierarchies.
 
 GraphQL supports two abstract types: interfaces and unions.
 
-An `Interface` defines a list of fields; `Object` types and other Interface
-types which implement this Interface are guaranteed to implement those fields.
-Whenever a field claims it will return an Interface type, it will return a valid
-implementing Object type during execution.
+An `Interface` defines an ordered set of fields; `Object` types and other
+Interface types which implement this Interface are guaranteed to implement those
+fields. Whenever a field claims it will return an Interface type, it will return
+a valid implementing Object type during execution.
 
-A `Union` defines a list of possible types; similar to interfaces, whenever the
+A `Union` defines a set of possible types; similar to interfaces, whenever the
 type system claims a union will be returned, one of the possible types will be
 returned.
 
@@ -674,11 +675,11 @@ GraphQL operations are hierarchical and composed, describing a tree of
 information. While Scalar types describe the leaf values of these hierarchical
 operations, Objects describe the intermediate levels.
 
-GraphQL Objects represent a list of named fields, each of which yield a value of
-a specific type. Object values should be serialized as ordered maps, where the
-selected field names (or aliases) are the keys and the result of evaluating the
-field is the value, ordered by the order in which they appear in the _selection
-set_.
+GraphQL Objects represent an ordered set of named fields, each of which yield a
+value of a specific type. Object values should be serialized as ordered maps,
+where the selected field names (or aliases) are the keys and the result of
+evaluating the field is the value, ordered by the order in which they appear in
+the _selection set_.
 
 All fields defined within an Object type must not have a name which begins with
 {"\_\_"} (two underscores), as this is used exclusively by GraphQL's
@@ -920,7 +921,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          returns {true}.
       4. If argument type is Non-Null and a default value is not defined:
          1. The `@deprecated` directive must not be applied to this argument.
-3. An object type may declare that it implements one or more unique interfaces.
+3. An object type may declare that it implements a set of one or more unique
+   interfaces.
 4. An object type must be a super-set of all interfaces it implements:
    1. Let this object type be {objectType}.
    2. For each interface declared implemented as {interfaceType},
@@ -982,7 +984,7 @@ InputValueDefinition : Description? Name : Type DefaultValue? Directives[Const]?
 
 Object fields are conceptually functions which yield values. Occasionally object
 fields can accept arguments to further specify the return value. Object field
-arguments are defined as a list of all possible argument names and their
+arguments are defined as an ordered set of all possible argument names and their
 expected input types.
 
 All arguments defined within a field must not have a name which begins with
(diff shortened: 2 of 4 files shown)


spec/Section 3 -- Type System.md
```diff
@@ -1473,7 +1474,7 @@ EnumValuesDefinition : { EnumValueDefinition+ }
 EnumValueDefinition : Description? EnumValue Directives[Const]?
 
 GraphQL Enum types, like Scalar types, also represent leaf values in a GraphQL
-type system. However Enum types describe the set of possible values.
+type system. However Enum types describe the list of possible values.
```
benjie, line 1477: I considered "However Enum types describe as a list the set of possible values." but it didn't seem worth the extra words.

martinbonnin (approved):

cdaringe:
The prior presence of “set” to me unambiguously implied unordered. Your switch to “list” i think is likely? a weaker implicit association with ordering for serialized content via introspection.

This proposal definitely works (happy to support it), but i think we needed top level or inline definitions/assertions about what terms (eg list) apply ordered or not.

benjie:
> The prior presence of “set” to me unambiguously implied unordered.

I don't think the term "set" itself implies ordered or unordered, "set" is independent of ordering. For example in JavaScript a `Set` is ordered, in Java a TreeSet is ordered but a HashSet is not, in mathematics you can have ordered and partially ordered sets. Whether the sets were ordered or not was not stated.

> i think we needed top level or inline definitions/assertions about what terms (eg list) apply ordered or not.

I agree. The term "list" itself is not as unamiguously "ordered" as I at first thought; for example here's an excerpt from an A-level Computing syllabus in the UK:

> A list is a number of items in an ordered or unordered structure.
> -- https://en.wikibooks.org/wiki/A-level_Computing/AQA/Problem_Solving,_Programming,_Operating_Systems,_Databases_and_Networking/Programming_Concepts/Lists

Similarly HTML has both `<ol>` and `<ul>` tags for ordered and unordered lists respectively.

---

An alternative fix would be to leave the existing wording as-is (but maybe slightly edited for consistency) and instead indicate that all lists and sets in the spec are ordered unless otherwise stated. We currently leave the order to be "implementation defined". Essentially, we are asking the question "are the two following schemas the same?":

```graphql
enum E {
  A
  Z
}
type Query {
  a: E
}
```

```graphql
enum E {
  Z
  A
}
type Query {
  a: E
}
```

I'd argue that they are not, and that the order of the enum is significant. A change in that order would trigger a change in the printed SDL that you might version control, a change in any generated documentation, a change in browsing it in GraphiQL, and could have many other consequences elsewhere (for example code generation). We should definitely be more explicit about this.

benjie:
I've added a note to the beginning of the spec :+1:

spec/GraphQL.md
```diff
@@ -122,6 +122,11 @@ explicitly in prose (e.g. "Note: ") or are set apart in a note block, like this:
 
 Note: This is an example of a non-normative note.
 
+**Lists Are Ordered**
+
+Unless otherwise stated (for example, "an unordered list"), any mention of the
+term "list" in this document indicates an ordered collection.
```
martinbonnin, line 128: Almost feels like we need to differentiate between syntax order and semantic (~execution?) order. In the syntax, I think I always want determined order. At execution, sometimes the order is unimportant like for input object literals. 
Don't want to bikeshed this more than needed though. This PR is a net improvement to clarity so I'm all for it 👍 .

benjie: The coercion of input objects is explicitly unordered: https://spec.graphql.org/draft/#sec-Input-Objects.Input-Coercion; the order of a _literal_ is arguably important though, e.g. `{f(o:{a:1,b:2})}` and `{f(o:{b:2,a:1})}` are concretely different documents even though the coerced interpretation of the `o` argument is equivalent.

cdaringe, line 128: 👍 . Thanks--this catch-all I think does a lot of important work 😃

cdaringe (approved):

leebyron:
Following up on this after the discussion in the WG meeting today, my suggestions:

1. Add an appendix section describing the data structure terms we use more specifically along with a note about preserving observable order for semantically unordered data when possible.

2. In the introductory section on conformance, add in references to "data collections" alongside "algorithms" where relevant. What we say there about allowing differences in implementation as long as outcomes are observably correct remains true for what data structures implementors choose to use.

3. Add in a non-normative note in the introspection section about how order "should" be preserved for the unordered sets when possible (but remove redefining set to list)


This was a bit of a riff, so edit away:

```markdown
## A.6 Data Collections

This specification describes the semantic properties of data collections 
using the terms with defined properties. This describes observable data 
collections such as the result of applying a grammar and the inputs and 
outputs of algorithms. It also describes unobservable data collections 
such as temporary data internal to an algorithms. Each data collection 
type defines the operations available, and whether values unique or ordered.

**List**

The term "list" describes a sequence of values which may not be unique. 
A list is ordered unless explicitly stated otherwise (as an "unordered 
list"). For clarity the term "ordered list" may be used when an order 
is semantically important.

**Set**

The term "set" describes a unique collection of values, where each value 
is considered a "member" of that set. A set is unordered unless explicitly 
stated otherwise (as an "ordered set"). For clarity the term "unordered set" 
may be used when the lack of an order is semantically important.

**Map**

The term "map" describes a collection of "entry" key and value pairs, where 
the set of keys across all entries is unique but the values across all 
entries may repeat. A map is unordered unless explicitly stated otherwise 
(as an "ordered map"). For clarity the term "unordered map" may be used 
when the lack of an order is semantically important.

Note: To improve legibility, when possible implementations should preserve
observable order for unordered data collections. For example, if an applied 
grammar to an input string results in an unordered set, serializing that 
set (to a string or other observable output) should produce the same order 
found in the original input string.
```

benjie:
@leebyron I've changed list, set and map into definitions and have also changed the final non-normative note into a normative statement since it uses the word `should` (equivalent to `we recommend`, which is true - we do recommend that) and thus does not add a requirement; but otherwise I've not much changed your section A.6 suggestion other than the final sentence.

I've also attempted to implement your other suggestions, and have then gone through and corrected the types throughout section 3, for example an object doesn't define a list of fields but more an ordered set of fields...

... except, is it really a map where the field names are the key? I'll add this to the WG next month.

michaelstaib (approved):
This clarification is nice!

benjie:
This has been split into 3 PRs, it's now the top PR on top of:

- #1091
- #1092

spec/Section 3 -- Type System.md
```diff
@@ -302,17 +302,18 @@ enumerable. GraphQL offers an `Enum` type in those cases, where the type
...
+another type in the system, allowing the definition of arbitrary type
+hierarchies.
 
 GraphQL supports two abstract types: interfaces and unions.
 
-An `Interface` defines a list of fields; `Object` types and other Interface
-types which implement this Interface are guaranteed to implement those fields.
-Whenever a field claims it will return an Interface type, it will return a valid
-implementing Object type during execution.
+An `Interface` defines an ordered set of fields; `Object` types and other
```
mjmahone, line 311: IMO Interface/Object is not really an ordered set? Maybe it is? It feels a little bit annoying for maintainers to feel like they must maintain order (does the order of fields in the object need to match the order in the interface)?

spec/Section 3 -- Type System.md
```diff
@@ -920,7 +921,8 @@ of rules must be adhered to by every Object type in a GraphQL schema.
          returns {true}.
       4. If argument type is Non-Null and a default value is not defined:
          1. The `@deprecated` directive must not be applied to this argument.
-3. An object type may declare that it implements one or more unique interfaces.
+3. An object type may declare that it implements a set of one or more unique
```
mjmahone, line 924: Wouldn't this also be an ordered set of interfaces?

spec/Section 3 -- Type System.md
```diff
@@ -1347,7 +1350,7 @@ UnionMemberTypes :
 - UnionMemberTypes | NamedType
 - = `|`? NamedType
 
-GraphQL Unions represent an object that could be one of a list of GraphQL Object
+GraphQL Unions represent an object that could be one of a set of GraphQL Object
```
mjmahone, line 1353: Why is this not ordered?

spec/Section 3 -- Type System.md
mjmahone: Basically, IMO every "set" we have is in fact an ordered set, so specifying ordered set vs. just set is more, not less, confusing.

Am I wrong and is there in fact any unordered set? Or should we just use set and use unordered set when we don't expect you to be able to preserve order?

spec/Section 3 -- Type System.md
```diff
@@ -302,17 +302,18 @@ enumerable. GraphQL offers an `Enum` type in those cases, where the type
 specifies the space of valid responses.
 
 Scalars and Enums form the leaves in response trees; the intermediate levels are
-`Object` types, which define a set of fields, where each field is another type
-in the system, allowing the definition of arbitrary type hierarchies.
+`Object` types, which define an ordered set of fields, where each field is
```
SimonSapin, line 305: Shouldn’t objects (and interfaces, and input objects) have an ordered _map_ of fields? The field names are unique, not the field definitions as a whole. `foo: Int, foo: String` would be a set of two different fields that happen to share a name.

spec/Section 3 -- Type System.md
SimonSapin: A set in set theory is unordered, and calling a programming data structure a set outside of this spec generally implies it to be unordered. Even if it wouldn’t be exactly wrong for the spec to only say in one place "all sets in this document are implicitly ordered" it’s not great spec-writing. Specifying "ordered set" explicitly every time makes readers less likely to make a wrong assumption.

spec/Section 3 -- Type System.md
```diff
@@ -982,7 +984,7 @@ InputValueDefinition : Description? Name : Type DefaultValue? Directives[Const]?
 
 Object fields are conceptually functions which yield values. Occasionally object
 fields can accept arguments to further specify the return value. Object field
-arguments are defined as a list of all possible argument names and their
+arguments are defined as an ordered set of all possible argument names and their
 expected input types.
```
SimonSapin, line 988: Similar to fields (comment above): rather than a set, shouldn’t field definitions (and directive definitions) have an ordered map of argument names to argument definitions?

And the values of this map are more than expected types: an argument definition may also include a default value and directives)

spec/Section 3 -- Type System.md
benjie: @SimonSapin 

> A set in set theory is unordered

Veering a little off topic, but that's not technically true, more accurate might be to say that pure set theory is "order agnostic" - the concept of order is entirely absent from the definition of a set. Typically set theory and order theory go hand in hand, and a set can be ordered, unordered, or even partially ordered.

That said, I agree with your conclusion that explicit > implicit.

---

@mjmahone 

> IMO Interface/Object is not really an ordered set? Maybe it is? It feels a little bit annoying for maintainers to feel like they must maintain order (does the order of fields in the object need to match the order in the interface)?

Great catch. IMO an interface defines an ordered set of fields (order is significant, and must be reflected in introspection); however we should state that implementations can implement this set of fields in any order - I've added a commit 7170d82 to address this.

spec/Section 3 -- Type System.md
benjie: Yep; great catch - 602d319

spec/Section 3 -- Type System.md
benjie: Agree ce8c732

spec/Section 3 -- Type System.md
benjie: This is something I raised in the WG actually; but that change is beyond the scope of this PR.

spec/Section 3 -- Type System.md
benjie: Here can be interpreted as "(an ordered set of all possible argument names) (and their expected input types)" so the set is fine; but I hate ambiguity and would rather define it as a map too. I think that's currently out of scope for this PR though.

spec/Section 3 -- Type System.md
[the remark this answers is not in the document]
benjie: How do you feel about a change in the wording:

```diff
-3. An object type may declare that it implements an ordered set of one or more
-   unique interfaces.
+3. An object type may declare that it implements one or more unique interfaces,
+   these interfaces form an ordered set.
```

Personally I think that the wording already covers this ("one or more unique interfaces" implies each can only be specified once, and then "an ordered set of <that>" makes it into an ordered set afterwards); but perhaps this more awkward wording is more precise?

BoD (approved):

PR #1066, opened 2023-12-07
enhance(ResolveFieldValue): note that list items may be async

yaacovCR:
and some baseline collection language for comparison

extracted from #742

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -672,8 +672,9 @@ ResolveFieldValue(objectType, objectValue, fieldName, argumentValues):
 Note: It is common for {resolver} to be asynchronous due to relying on reading
 an underlying database or networked service to produce a value. This
 necessitates the rest of a GraphQL executor to handle an asynchronous execution
-flow. In addition, an implementation for collections may leverage asynchronous
-iterators or asynchronous generators provided by many programming languages.
+flow. In addition, for fields that have a return type that is a List type, each
+value in a collection of values returned by {resolver} may be retrieved
+asynchronously.
 
 ### Value Completion
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -656,6 +656,12 @@ As an example, this might accept the {objectType} `Person`, the {field}
 {"soulMate"}, and the {objectValue} representing John Lennon. It would be
 expected to yield the value representing Yoko Ono.
 
+List values are resolved similarly. For example, {ResolveFieldValue} might also
+accept the {objectType} `MusicBand`, the {field} {"members"}, and the
+{objectValue} representing the Beatles. It would be expected to yield a
+collection of values representing John Lennon, Paul McCartney, Ringo Starr and
+George Harrison.
+
 ResolveFieldValue(objectType, objectValue, fieldName, argumentValues):
 
 - Let {resolver} be the internal function provided by {objectType} for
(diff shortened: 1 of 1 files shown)


benjie (dismissed):
Please restore the paragraph breaks which will both minimise the diff and restore the formatting of the note (which has special handling in spec-md).

spec/Section 6 -- Execution.md
```diff
@@ -660,7 +666,8 @@ ResolveFieldValue(objectType, objectValue, fieldName, argumentValues):
 Note: It is common for {resolver} to be asynchronous due to relying on reading
 an underlying database or networked service to produce a value. This
 necessitates the rest of a GraphQL executor to handle an asynchronous execution
-flow.
+flow. In addition, an implementation for collections may leverage asynchronous
+iterators or asynchronous generators provided by many programming languages.
```
benjie, line 670: This would introduce the term "generator" and "iterator" to the spec, which isn't currently used. Let's keep the language simpler:

```suggestion
flow. In addition, for fields that have a return type that is a List type, each
value in a collection of values returned by {resolver} may itself be
asynchronous.
```

benjie (changes requested):

spec/Section 6 -- Execution.md
yaacovCR: What about something very similar:

```suggestion
flow. In addition, for fields that have a return type that is a List type, each
value in a collection of values returned by {resolver} may be
retrieved asynchronously.
```

This just avoids saying that the value itself is asynchronous, which I think you might mean to just be a more concise version of the above?

@robrichard may have input, this was extracted from #742

spec/Section 6 -- Execution.md
```diff
@@ -656,6 +656,12 @@ As an example, this might accept the {objectType} `Person`, the {field}
 {"soulMate"}, and the {objectValue} representing John Lennon. It would be
 expected to yield the value representing Yoko Ono.
 
+List values are resolved similarly. For example, {ResolveFieldValue} might also
```
benjie, line 659: I do not like "similarly" because technically it follows the exact same algorithm, whereas "similarly" implies that there _is_ a difference. Let's try alternate wording:

```suggestion
For an example relating to list fields, {ResolveFieldValue} might
```

Though I'm not convinced we need this paragraph.

spec/Section 6 -- Execution.md
```diff
@@ -666,7 +672,9 @@ ResolveFieldValue(objectType, objectValue, fieldName, argumentValues):
 Note: It is common for {resolver} to be asynchronous due to relying on reading
 an underlying database or networked service to produce a value. This
 necessitates the rest of a GraphQL executor to handle an asynchronous execution
-flow.
+flow. In addition, for fields that have a return type that is a List type, each
+value in a collection of values returned by {resolver} may be retrieved
+asynchronously.
```
benjie, line 677: Alternative wording suggestion:

```suggestion
flow. If the field is of a list type, each
value in the collection of values returned by {resolver} may itself be retrieved
asynchronously.
```

I don't like mixing a non-normative note with "may" (which is usually normative); but I think this is okay.

benjie (changes requested):
Minor wording suggestions

spec/Section 6 -- Execution.md
yaacovCR: I just removed the whole paragraph.

spec/Section 6 -- Execution.md
yaacovCR: Sounds good!

benjie (approved):
Looks good to me, and despite being non-normative this already aligns with the behavior of GraphQL.js. I think we can flag this as an editorial note.

benjie:
@graphql/tsc This PR extends the non-normative note about resolvers being asynchronous to also note that each entry in a list returned from a resolver can itself be asynchronous. Please either approve or express your concerns - thanks!

robrichard (approved):

fotoetienne (approved):

PR #1068, opened 2023-12-08
3.10 Input Objects - clarify lists are permitted as Input fields

msolomon:
A minor omission in 3.10 Input Objects confused me briefly today by suggesting lists might not be permitted as the type of an input field. However, they are (and are even used in the examples), so add it in.

Alternatively, the kinds of permitted fields could be described explicitly as they are under Field Arguments: https://spec.graphql.org/draft/#sel-GAHZjCZABABDukZ

I believe this is simply an editing mistake.


benjie (changes requested):
Good find! Since "list" is actually a wrapping type, and you can't have a "list of unions" here, we should be more specific. I've made an alternative suggestion:

benjie (approved):
LGTM! @graphql/tsc another +1 and we'll merge :+1:

PR #1069, opened 2023-12-14
Enforce consistent punctuation in algorithms

benjie:
Follow up to #1067 with thorough check through all algorithm steps, and a CI check to ensure there's no regressions.

Using consistent formatting within the spec will make it easier to read, easier to search through, and easier for automated tools to parse it.

cc @yaacovCR

PR #1070, opened 2023-12-14
Fix formatting and typo in `MapSourceToResponseEvent`

benjie:
Based on https://github.com/graphql/graphql-spec/pull/1069; that PR needs merging first (hence this is marked as a draft). If reviewing this, **just look at the final commit only**.

Lee did an editorial edit to the [subscriptions PR](https://github.com/graphql/graphql-spec/pull/305) ("Spec: Formal Subscriptions Definition") before merging, and added a line [(in this commit)](https://github.com/robzhu/graphql/commit/47404cb687ec5a7a1d74e02d9a97fb19d1617e77#diff-0f02d73330245629f776bb875e5ca2b30978a716732abca136afdd028d5cd33cR260) to `MapSourceToResponseEvent`:

```diff
 MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
 
   * Return a new event stream {responseStream} which yields events as follows:
   * For each {event} on {sourceStream}:
     * Let {response} be the result of running
       {ExecuteSubscriptionEvent(subscription, schema, variableValues, event)}.
     * Yield an event containing {response}.
+  * When {responseStream} completes: complete this event stream.
```

The "this event stream" is unclear; I would expect it to refer to `responseStream` (the stream that we've created) but `When {responseStream} completes: complete {responseStream}.` doesn't make sense.

I figured that it _might_ be a deliberate way of saying that {sourceStream} should be closed if {responseStream} completes, however that's already handled in the Unsubscribe algorithm text:

> #### Unsubscribe
> 
> Unsubscribe cancels the Response Stream when a client no longer wishes to
> receive payloads for a subscription. This may in turn also cancel the Source
> Stream. This is also a good opportunity to clean up any other resources used by
> the subscription.
> 
> Unsubscribe(responseStream):
> 
> - Cancel {responseStream}.

Note "This may in turn also cancel the Source Stream." - i.e. *may*, we are not enforcing this. So `MapSourceToResponseEvent` should not enforce it.

I believe this was a simple typo, and should have been `When {sourceStream} completes: complete {responseStream}.` so I have made this edit.

Also, we now enforce consistent formatting in the spec (via https://github.com/graphql/graphql-spec/pull/1069), so algorithm steps after a `:` should be indented; I've fixed this also.

### DIFF
--- a/.github/algorithm-format-check.mjs
+++ b/.github/algorithm-format-check.mjs
@@ -1,0 +1,114 @@
+import { readFile, readdir } from "node:fs/promises";
+
+const SPEC_DIR = new URL("../spec", import.meta.url).pathname;
+
+process.exitCode = 0;
+const filenames = await readdir(SPEC_DIR);
+for (const filename of filenames) {
+  if (!filename.endsWith(".md")) {
+    continue;
+  }
+  const markdown = await readFile(`${SPEC_DIR}/${filename}`, "utf8");
+
+  /**
+   * Not strictly 'lines' since we try and group indented things together as if
+   * they were one line. Close enough though.
+   */
+  const lines = markdown.split(/\n(?=[\S\n]|\s*(?:-|[0-9]+\.) )/);
+
+  for (let i = 0, l = lines.length; i < l; i++) {
+    const line = lines[i];
+
+    // Check algorithm is consistently formatted
+    {
+      // Is it an algorithm definition?
+      const matches = line.match(/^([a-z0-9A-Z]+)(\s*)\(([^)]*)\)(\s*):(\s*)$/);
+      if (matches) {
+        const [, algorithmName, ns1, _args, ns2, ns3] = matches;
+        if (ns1 || ns2 || ns3) {
+          console.log(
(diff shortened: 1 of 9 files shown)


yaacovCR:
If the line was originally supposed to refer to the source stream, or even if we are simply just changing it to refer to the source stream, we cannot say “when the source stream completes “, we have to say instead “if the source stream completes. “

benjie:
I don't think it's necessary but nonetheless I've changed it to "if and when" to make clear both the conditional ("if") and timely ("when") elements.

I'd argue `thing.on('error', e => console.log(e))` can be expressed as "when an error occurs, log it" since the "when" _can_ introduce the condition whilst also making the "at the time of" explicit; I don't think it matters that the "when" may never occur. However it's arguably more precise (though a little pedantic) to say "if and when an error occurs, log it"; and what is a specification if not pedantically precise? :laughing:

benjie:
This was later spotted by @Yogu and their fix has since been merged; so this PR is no longer needed. Ref: https://github.com/graphql/graphql-spec/pull/994

PR #1072, opened 2024-01-03
incremental: maintain defer fragment ordering

yaacovCR:
depends on #1052

Maintains ordering of defers from operation, only sending a child deferred fragment after the parent is sent.  By contrast, #1052 will send a deferred fragment once it can be merged into the final reconcilable object.

since it depends on #1052, this is probably the more instructive diff: https://github.com/yaacovCR/graphql-spec/compare/deduplicate3...deduplicate5

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -4,6 +4,7 @@ ignoreRegExpList:
   - /[a-z]{2,}'s/
 words:
   # Terms of art
+  - deprioritization
   - endianness
   - interoperation
   - monospace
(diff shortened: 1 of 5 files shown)


yaacovCR:
https://github.com/yaacovCR/graphql-spec/compare/deduplicate3...deduplicate5

yaacovCR:
pulling this into #1052 based on feedback at https://github.com/graphql/defer-stream-wg/discussions/80

PR #1073, opened 2024-01-25
Be strict about error paths format

martinbonnin:
Replace `should` with `must` in the description of error paths: `This field must be a list of path segments starting...`

```json
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [{ "line": 6, "column": 7 }],
      "path": ["hero", "heroFriends", 1, "name"]
    }
```

Anything else will make it impossible to handle errors in error-aware clients. This is especially important for [strict-nullability](https://github.com/graphql/graphql-wg/discussions/1410) or [semantic-non-nullability](https://github.com/graphql/graphql-spec/pull/1065)

### DIFF
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -109,12 +109,12 @@ must contain an entry with the key `path` that details the path of the response
 field which experienced the error. This allows clients to identify whether a
 `null` result is intentional or caused by a runtime error.
 
-This field should be a list of path segments starting at the root of the
-response and ending with the field associated with the error. Path segments that
-represent fields should be strings, and path segments that represent list
-indices should be 0-indexed integers. If the error happens in an aliased field,
-the path to the error should use the aliased name, since it represents a path in
-the response, not in the request.
+This field must be a list of path segments starting at the root of the response
+and ending with the field associated with the error. Path segments that
+represent fields must be strings, and path segments that represent list indices
+must be 0-indexed integers. If the error happens in an aliased field, the path
+to the error must use the aliased name, since it represents a path in the
+response, not in the request.
 
 For example, if fetching one of the friends' names fails in the following
 operation:
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -109,12 +109,12 @@ must contain an entry with the key `path` that details the path of the response
 field which experienced the error. This allows clients to identify whether a
 `null` result is intentional or caused by a runtime error.
 
-If present, this field must be a list of path segments starting at the root of the response
-and ending with the field associated with the error. Path segments that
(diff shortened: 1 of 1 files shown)


benjie (commented):
This is technically a breaking change; but it seems more of a bugfix to me - this should (haha) have been a `must` all along, and I'm not aware of any implementation that is breaking these rules. I'm not going to mark this as an approval until we've discussed at the WG, but I am heavily in favour of it.

@leebyron Was this deliberately a `should`, or was it a case where natural language accidentally made its way in rather than RFC2119 language?

@martinbonnin Please add this to an upcoming WG for discussion.

cc @graphql/implementers  - is anyone's implementation going to break if we change this from `should` to `must`?

captbaritone:
cc @mjmahone in case you have context on why this is not "must". You mentioned that this might have been an intentional choice to allow servers to use errors to silently handle things like permissions?

I can see a parallel here with things like returning 404 errors when you try to access a route that you don't have permission to view in order to prevent you from using the status code to determine if a private entity exists or not.

mjmahone:
@captbaritone I was misunderstanding the context. I don't have an objection to the `path` being required to be a list of strings/ints if it exists.

mjmahone (approved):

spec/Section 7 -- Response.md
```diff
@@ -109,12 +109,12 @@ must contain an entry with the key `path` that details the path of the response
 field which experienced the error. This allows clients to identify whether a
 `null` result is intentional or caused by a runtime error.
 
-This field should be a list of path segments starting at the root of the
-response and ending with the field associated with the error. Path segments that
-represent fields should be strings, and path segments that represent list
-indices should be 0-indexed integers. If the error happens in an aliased field,
-the path to the error should use the aliased name, since it represents a path in
-the response, not in the request.
+This field must be a list of path segments starting at the root of the response
```
benjie, line 112:
```suggestion
If present, this field must be a list of path segments starting at the root of the response
```

benjie (approved):

leebyron:
@graphql/implementers - please review before we accept

leebyron:
Radiating intent here (and action item): We intend to accept this and merge it next month barring any feedback.

benjie (approved):

andimarek:
From a GraphQL Java POV we are ok with that

mjmahone (approved):

robrichard (approved):

Keweiqu (approved):

mjmahone:
At the Mar 7 WG meeting we established that we intend to merge this PR after the April WG meeting, unless there are objections.

If you implement a GraphQL server or client and do *not* implement the error paths array as specified, please let us know!

michaelstaib (approved):

Issue #1076, opened 2024-02-11
Composite types is not clearly defined


andimarek:
The spec uses the word "composite" 5 times, but it is never clearly defined.

While it is quite clear that composite for output types means Interface, Object or Union the usage for input types is in my experience much less common. The only input composite type is Input Object and the spec uses it in this way too.

We should clearly define the meaning of composite type.

martinbonnin:
+1 👍 

Another option could be to just remove the "composite" term and be explicit instead? (i.e. use "object, interface or union") 

Looks like the main place where the term is used in normative context is [in field merging](https://spec.graphql.org/draft/#sel-IALRFDHCLCAACEa0ne):

> 5. If typeA or typeB is Scalar or Enum: ...
>   a. If typeA and typeB are the same type return true, otherwise return false.
> 6. Assert: typeA and typeB are both composite types.

That last sentence could be replaced by

> 6. Assert: typeA is Object or Union or Interface and typeB is Object or Union or Interface.

It's more of a mouthful but the spec isn't really meant to be read out loud? :)

PR #1078, opened 2024-02-13
chore: add clarifying note for composite and expand term

JoviDeCroock:
Resolves https://github.com/graphql/graphql-spec/issues/1076

This PR adds a clarifying note for the term composite in prior versions of the spec and expands the term into `Object, Interface or Union type` wherever it was found

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -352,7 +352,7 @@ Fields\:
 
 **Input Object**
 
-Input objects are composite types defined as a list of named input values. They
+Input objects are types defined as a list of named input values. They
 are only used as inputs to arguments and variables and cannot be a field return
 type.
 
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -352,7 +352,7 @@ Fields\:
 
 **Input Object**
 
-Input objects are types defined as a list of named input values. They
+Input objects are composite types defined as a list of named input values. They
 are only used as inputs to arguments and variables and cannot be a field return
 type.
 
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -446,7 +446,8 @@ SameResponseShape(fieldA, fieldB):
 - If {typeA} or {typeB} is Scalar or Enum:
   - If {typeA} and {typeB} are the same type return {true}, otherwise return
     {false}.
-- Assert: {typeA} and {typeB} are both Object, Interface or Union types.
+- Assert: {typeA} is an object, union or interface type.
+- Assert: {typeB} is an object, union or interface type.
 - Let {mergedSet} be the result of adding the selection set of {fieldA} and the
   selection set of {fieldB}.
 - Let {fieldsForName} be the set of selections with a given response name in
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -910,7 +910,7 @@ fragment inlineNotExistingType on Dog {
 }
 ```
 
-#### Fragments on object, interface or union Types
+#### Fragments on Object, Interface or Union Types
 
 **Formal Specification**
 
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -455,6 +455,9 @@ SameResponseShape(fieldA, fieldB):
   - If {SameResponseShape(subfieldA, subfieldB)} is {false}, return {false}.
 - Return {true}.
 
+Note: composite is used to signal a type that is either an object-, interface-
+or union-type.
+
 **Explanatory Text**
 
 If multiple field selections with the same response names are encountered during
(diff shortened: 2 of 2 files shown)


spec/Section 4 -- Introspection.md
```diff
@@ -352,7 +352,7 @@ Fields\:
 
 **Input Object**
 
-Input objects are composite types defined as a list of named input values. They
+Input objects are types defined as a list of named input values. They
```
benjie, line 355: I think the `composite` improves the clarity of the text. I think that the term "composite type" is fairly well understood in prose, e.g. [Wikipedia: composite data type](https://en.wikipedia.org/wiki/Composite_data_type). I suggest we revert this change.

```suggestion
Input objects are composite types defined as a list of named input values. They
```

spec/Section 5 -- Validation.md
```diff
@@ -910,7 +910,7 @@ fragment inlineNotExistingType on Dog {
 }
 ```
 
-#### Fragments on Composite Types
+#### Fragments on object, interface or union Types
```
benjie, line 913: See the [style guide on titles](https://github.com/graphql/graphql-spec/blob/main/STYLE_GUIDE.md#title-case):

```suggestion
#### Fragments on Object, Interface or Union Types
```

benjie (changes requested):
I support the replacement of "composite types" in section 5 with more explicit language.

benjie (dismissed):
I think this helps clarity :+1:

benjie:
@graphql/tsc Can we get another approval, then we'll wait a week or two for any more feedback and if no-one has concerns I think we should merge :+1:

leebyron:
I’m OK with this change however I’m pretty sure the term composite type
shows up all over the graphql-js repository as well

Keweiqu (approved):
Makes sense! Thanks for doing this

JoviDeCroock:
As suggested in yesterdays WG this has been replaced with a note about composite

martinbonnin:
I was thinking we would keep "composite" out of the normative sections (initial PR) and add a non-normative note that "composite" may be found in other contexts:

```
Note: implementers may use the term composite type for a type that is either an object, interface
or union.
```

martinbonnin (approved):

benjie:
@JoviDeCroock I suggest you undo your force-push and then add a note along the lines of what you just added; essentially the change should be to remove "composite" from most places (preferring explicitness) and then to add a non-normative note indicating that previous versions of the spec used and some implementers use the term "composite type" to refer to Object, Interface and Union types. 

Also, the letter after `Note: ` should be a capital since that is what is rendered in the note box (the `Note: ` prefix is stripped out).

spec/Section 5 -- Validation.md
```diff
@@ -455,6 +455,9 @@ SameResponseShape(fieldA, fieldB):
   - If {SameResponseShape(subfieldA, subfieldB)} is {false}, return {false}.
 - Return {true}.
 
+Note: composite is used to signal a type that is either an object-, interface-
```
JoviDeCroock, line 458:
```suggestion
Note: Composite is used to signal a type that is either an object-, interface-
```

JoviDeCroock:
@benjie the force push was just to update this branch with `main` I re-added the `composite` rephrasing and updated the note. Thank you for the review 🙌

spec/Section 5 -- Validation.md
```diff
@@ -455,6 +455,9 @@ SameResponseShape(fieldA, fieldB):
   - If {SameResponseShape(subfieldA, subfieldB)} is {false}, return {false}.
 - Return {true}.
 
+Note: In prior versions of the spec the term "composite" wass used to signal
+a type that is either an object-, interface- or union-type.
```
benjie, line 459:
```suggestion
Note: In prior versions of the spec the term "composite" was used to signal
a type that is either an Object, Interface or Union type.
```

benjie (changes requested):
Looking good!

martinbonnin (approved):
👍

spec/Section 5 -- Validation.md
```diff
@@ -446,7 +446,7 @@ SameResponseShape(fieldA, fieldB):
 - If {typeA} or {typeB} is Scalar or Enum:
   - If {typeA} and {typeB} are the same type return {true}, otherwise return
     {false}.
-- Assert: {typeA} and {typeB} are both composite types.
+- Assert: {typeA} and {typeB} are both Object, Interface or Union types.
```
benjie, line 449: This reads ambiguously; it could be interpretted as `(typeA is object and typeB is object) or (typeA is union and typeB is union) or (typeA is an interface and typeB is an interface)` rather than `(typeA is object or union or interface) and (typeB is object or union or interface)`. Perhaps:

```suggestion
- Assert: {typeA} is an object, union or interface type.
- Assert: {typeB} is an object, union or interface type.
```

benjie (changes requested):
This one got undone accidentally I think!

benjie (approved):
Looks great to me :+1:

benjie:
I believe @Keweiqu and @leebyron's previous reviews are still valid with the latest state, though they may want to review the added note.

mjmahone (approved):
It's nice to have a word to use for "Types that are schema-defined vertex types in your graph" but Composite isn't a great word for that, agree.

robrichard (approved):

PR #1010, opened 2023-01-19
Fragment Arguments: Spec Implementation

mjmahone:
This fully specifies the Fragment Arguments changes required, as evidenced by the implementation changes made to `graphql-js` in https://github.com/graphql/graphql-js/pull/3835

This is an update to a years-old RFC PR https://github.com/graphql/graphql-spec/pull/865, but given that PR had a lot of discussion that has been resolved, it should be easier for reviewers to start with this PR as a cleaner slate. I've tried to incorporate all discussion from previous PRs and Working Group discussion into the initial commit here.

Please use this PR for reviewing and verifying the accuracy of the PR, with inline comments when possible.

**For discussions about Fragment Arguments as an idea, missing features or open questions, please open or respond to a thread in the graphql-wg discussion space: https://github.com/graphql/graphql-wg/discussions/1239**

PR #1081, opened 2024-02-16
Fragment arguments/variables (syntax/validation/execution)

### LINKED ISSUE #204: Parameterized fragments
A [recent talk about Relay 2](https://speakerdeck.com/wincent/relay-2-simpler-faster-more-predictable) suggested adding arguments to query fragments via directives. 

I think first-class support for arguments in fragments would give a few benefits: 

- Smaller variable scope: instead of checking each usage of a fragment (which may be deeply nested) to determine what variables are available, a fragment would be its own scope.
- More flexible reuse: since a fragment wouldn't depend on its query's variables, fragments could be applied to queries whose variables had different names. 

For example, a fragment might be defined with variables: 

```graphql 
fragment avatarFields($thumbSize: Int = 40) on Image {
  thumbURL(size: $thumbSize)
  uploadedAt
}
```

The spread with different values: 

```graphql 
{
  person(id: 1) {
    avatar {
      ... avatarFields(thumbSize: 80)
    }
  }
}
```

```graphql 
query findPerson($personId: Int!, $avatarSize: Int!) {
  person(id: $personId) {
    avatar {
      ... avatarFields(thumbSize: $avatarSize)
    }
  }
}
```

To be honest, I don't have a use case. But this seems like a really solid feature, and as a library maintainer, I'd rather add this language feature than implement some new directives! (Or, would those directives be "compiled out"?)

Technically, this could even be backwards compatible. Fragments could _also_ use variables from their outer scope.

What do you think, would this be an improvement? Have I overlooked any downsides or complexities?

JoviDeCroock:
This spec contains amendments to https://github.com/graphql/graphql-spec/pull/1010, a diffed view is available at https://github.com/mjmahone/graphql-spec/pull/3.

These amendments are made from comments on the implementation PR and alterations from the new implementation

- Changing fragment-key to just be fragment-name https://github.com/graphql/graphql-js/pull/3835#discussion_r1101825832
- [Removing the substitution logic and going closer to the `coerce` logic we use in the general flow](https://github.com/JoviDeCroock/graphql-js/pull/2/commits/3397a8ac2525a259dabd32af3b7c50dc8a3cef84)

In general the biggest changes are that we introduce `fragmentVariableValues` which will be on the `groupedFieldSet`, these are derived from the arguments in scope of the fragmentDefinition where this field is used.

We introduce `localFragmentVariables` which as we are traversing down fragment-spreads are a coerced set of variables i.e.

```gql
query($b: String!, $c: String!) {
  field3(b: $b) ## uses b from variable-values
  ...A(a: "A", b: "B")
}

fragment A($a: String!, $b: String!) on X {
  field(a: $a) ## uses $a from localVariableValues
  field2(c: $c) ## we have access to variableValues so we take c from there
  ...B(b: $b) ## uses $b from localVariableValues
}

fragment B($b: String!) on X {
  ## This fragment has access to $b and $c but not to $a
  field4(b: $b) ## uses $b from localVariableValues
}
```

Last but not least we introduce `getArgumentValuesFromSpread` which looks at the spread and fragment-definition and establishes a coerced set of localVariableValues.

---

- Fixes https://github.com/graphql/graphql-spec/issues/204


mjmahone (commented):
This is clear to me :)

Very excited to see progress, and thank you for cleaning up and clarifying the PR!

spec/Section 5 -- Validation.md
```diff
@@ -570,6 +577,50 @@ fragment conflictingDifferingResponses on Pet {
...
+  doesKnowCommand(dogCommand: $command)
+}
+
+fragment potentiallyConflictingArguments(
+  $commandOne: DogCommand!
+  $commandTwo: DogCommand!
+) on Dog {
+  ...commandFragment(command: $commandOne)
+  ...commandFragment(command: $commandTwo)
+}
```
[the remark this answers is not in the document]
JoviDeCroock: I think it's good to allow for `...commandFragment ...commandFragment` (two argumentless fragment-spreads) for backwards compatability and `...commandFragment(a: $a) ...commandFragment(a: $a)` because they are the same which in my mind ties intuitively into the former case while the different arguments should be disallowed for ease of reasoning otherwise we'd have to go into the fragment-spreads and analyse whether this changed variable results in conflicting fields due to arguments on the field or `@include/@skip`/...

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,70 @@ size `60`:
...
 **Variable Use Within Fragments**
 
-Variables can be used within fragments. Variables have global scope with a given
-operation, so a variable used within a fragment must be declared in any
-top-level operation that transitively consumes that fragment. If a variable is
-referenced in a fragment and is included by an operation that does not define
-that variable, that operation is invalid (see
+Variables can be used within fragments. Operation-defined variables have global
+scope with a given operation, so a variable used within a fragment must either
+be declared in any top-level operation that transitively consumes that fragment,
```
Keweiqu, line 1220: The sentence would flow better if we do something like this --

Operation-defined variables have... Fragment-defined variables have ... So a variable used within a fragment must ...

Thoughts?

spec/Section 5 -- Validation.md
```diff
@@ -418,8 +418,15 @@ fragment directFieldSelectionOnUnion on CatOrDog {
 
 FieldsInSetCanMerge(set):
 
+- Let {visitedSelections} be the selections in {set} including visiting
+  fragments and inline fragments and applying any supplied fragment spread
+  arguments.
+- Let {spreadsForName} be the set of fragment spreads with a given name in
+  {visitedSelections}.
+- Given each pair of members {spreadA} and {spreadB} in {spreadsForName}:
+  - {spreadA} and {spreadB} must have identical sets of arguments.
```
Keweiqu, line 427: Should we specify "must have identical sets of arugments and values?"

Keweiqu: and directives?

JoviDeCroock: I actually hadn't thought about directives, that's a great point!

spec/Section 2 -- Language.md
JoviDeCroock: That sounds great, what do you think about https://github.com/graphql/graphql-spec/pull/1081/commits/5a7354371b14a3fcc42b5b376217283669869396

spec/Section 5 -- Validation.md
```diff
@@ -1563,12 +1636,36 @@ fragment HouseTrainedFragment on Query {
 }
 ```
 
+Likewise, it is valid for both an operation and a fragment to define a variable
```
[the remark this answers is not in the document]
JoviDeCroock: Addressed in https://github.com/graphql/graphql-spec/pull/1081/commits/48fa841ae59036d7721b3b5d2838f01a9f346eda, does that sound better?

JoviDeCroock:
@graphql/tsc is there anything I can do to move this forward?

spec/Section 6 -- Execution.md
```diff
@@ -562,9 +564,35 @@ DoesFragmentTypeApply(objectType, fragmentType):
...
+
 Note: The steps in {CollectFields()} evaluating the `@skip` and `@include`
 directives may be applied in either order since they apply commutatively.
 
+Note: The unset value used to replace unset fragment-defined arguments in
+{SubstituteFragmentArguments()} must not be a variable defined by any operation
+that includes the fragment. An example would be to use a variable with a
+reserved prefix, like `$__UNSET`, to replace all unset fragment-defined
+variables.
+
```
JoviDeCroock, line 595:
```suggestion
```

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,71 @@ size `60`:
 
 **Variable Use Within Fragments**
 
-Variables can be used within fragments. Variables have global scope with a given
-operation, so a variable used within a fragment must be declared in any
-top-level operation that transitively consumes that fragment. If a variable is
-referenced in a fragment and is included by an operation that does not define
-that variable, that operation is invalid (see
+Variables can be used within fragments. Operation-defined variables have global
+scope with a given operation. Fragment-defined variables have local scope within the
```
benjie, line 1219:
```suggestion
scope within a given operation. Fragment-defined variables have local scope within the
```

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,71 @@ size `60`:
...
+fragment definition they are defined in. A variable used within a fragment must either
```
benjie, line 1220:
```suggestion
fragment definition in which they are defined. A variable used within a fragment must either
```

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,71 @@ size `60`:
...
+be declared in any top-level operation that transitively consumes that fragment,
```
benjie, line 1221:
```suggestion
be declared in each top-level operation that transitively consumes that fragment,
```

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,71 @@ size `60`:
...
+## Fragment Variable Definitions
+
+Fragments may define locally scoped variables. This allows fragments to be
+reused while enabling the caller to specify the fragment's behavior.
+
+For example, the profile picture may need to be a different size depending on
+the parent context:
+
+```graphql example
+query withFragmentArguments {
```
benjie, line 1236:
```suggestion
query userAndFriends {
```

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,71 @@ size `60`:
...
+```
+
+In this case the `user` will have a larger `profilePic` than those found in the
+list of `friends`.
+
+A fragment-defined variable is scoped to the fragment that defines it.
+Fragment-defined variables are allowed to shadow operation-defined variables.
+
+```graphql example
+query withShadowedVariables($size: Int) {
```
benjie, line 1259:
```suggestion
query withShadowedVariables($size: Int!) {
```

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,71 @@ size `60`:
...
+  ...dynamicProfilePic(size: $size)
+}
+
+fragment dynamicProfilePic($size: Int!) on User {
+  profilePic(size: $size)
+}
+```
+
+The profilePic for `user` will be determined by the variables set by the
+operation, while `secondUser` will always have a profilePic of size 10. In this
```
benjie, line 1278:
```suggestion
operation, while `secondUser` will always have a `profilePic` of size `10`. In this
```

spec/Section 2 -- Language.md
```diff
@@ -1215,13 +1215,71 @@ size `60`:
...
+case, the fragment `variableProfilePic` uses the operation-defined variable,
+while `dynamicProfilePic` uses the value passed in via the fragment spread's
+argument `size`.
```
benjie, line 1281:
```suggestion
`size` argument.
```

Or alternatively: "the fragment spread's argument, `size`."

benjie (changes requested):
It feels like section 2 should comment on the fact that shadowed variables may not be referenced implicitly by child fragment spreads. (This is probably not the right way of saying that!) E.g. in the following query, `Forbidden` is spread inside of `F` and attempts to use variable `$a` which is ambiguous, and hence not permitted.

```graphql
query Q($a: Int!, $b: Int!) {
  ...F(a: 7)
}
fragment F($a: Int!) {
  ...Fine
  ...Forbidden
}
fragment Fine {
  b: echo(input: $b)
}
fragment Forbidden {
  a: echo(input: $a)
}
```

JoviDeCroock:
@benjie from a purely technical perspective I agree, I was approaching that more from a backwards compatibility perspective. The Fragments variables are only applicable within the context of a `FragmentDefinitiion` so we don't carry nested values. A fragment is always looked at in isolation and not from a call-stack point of view.

i.e.

```
query ($a: String!) {
  operation {
    ...fields(a: 'some-value')
  }
}

fragment fields($a: String) on Type {
  field(x: $a) ## this carries 'some-value'
  ...moreFields
}

fragment moreFields on Type {
  field2(x: $a) ## this would carry the value passed into the operation
}
```

spec/Section 5 -- Validation.md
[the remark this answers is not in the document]
benjie: I agree with Matt - this should use the same logic as field usage. This would thus be invalid for the same reason that this query (and a query using any two of these three fields) would be invalid:

```graphql
query ($id1: ID! = "ZmlsbXM6MQ==", $id2: ID! = "ZmlsbXM6MQ==") {
  node(id: "ZmlsbXM6MQ==") { id }
  node(id: $id1) { id }
  node(id: $id2) { id }
}
```

[Example on SWAPI](http://graphql.github.io/swapi-graphql/?query=query%20(%24id1%3A%20ID!%20%3D%20%22ZmlsbXM6MQ%3D%3D%22%2C%20%24id2%3A%20ID!%20%3D%20%22ZmlsbXM6MQ%3D%3D%22)%20%7B%0A%20%20node(id%3A%20%22ZmlsbXM6MQ%3D%3D%22)%20%7B%20id%20%7D%0A%20%20node(id%3A%20%24id1)%20%7B%20id%20%7D%0A%20%20node(id%3A%20%24id2)%20%7B%20id%20%7D%0A%7D)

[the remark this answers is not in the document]
JoviDeCroock: Made it more clear in https://github.com/graphql/graphql-spec/pull/1081/commits/03ba25598f7520c05b3dfb27697a84de76829c2e

spec/Section 5 -- Validation.md
```diff
@@ -675,16 +728,34 @@ fragment argOnRequiredArg on Dog {
...
+The following is invalid since `command` is not defined on
+`Dog.doesKnowCommand`.
 
 ```graphql counter-example
 fragment invalidArgName on Dog {
 }
 ```
 
+and this is also invalid as the variable `dogCommand` is not defined on fragment
```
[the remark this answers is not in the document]
benjie: Oh no wait this is that weird thing where we're calling them variables like operation variables rather than arguments... In which case, we should reference "the variable `$dogCommand`"?

I think calling them arguments makes more sense here at least. We'll have to be careful with this.

spec/Section 5 -- Validation.md
```diff
@@ -1563,12 +1636,36 @@ fragment HouseTrainedFragment on Query {
 }
 ```
 
+Likewise, it is valid for a fragment to define a variable with a name that
+is also defined on an operation:
+
+```graphql example
+query C($atOtherHomes: Boolean) {
+  ...HouseTrainedFragment
```
benjie, line 1628: This fragment is not defined?

benjie (changes requested):
I haven't got as far as section 6 yet.

spec/Section 5 -- Validation.md
```diff
@@ -647,23 +698,25 @@ query directQueryOnObjectWithSubFields {
...
 **Formal Specification**
 
 - For each {argument} in the document:
   - Let {argumentName} be the Name of {argument}.
-  - Let {argumentDefinition} be the argument definition provided by the parent
-    field or definition named {argumentName}.
+  - If the parent is a field or directive:
+  - Let {argumentDefinition} be the argument or variable definition named
+    {argumentName} provided by the parent field definition, directive definition
+    or fragment definition.
```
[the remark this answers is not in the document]
JoviDeCroock: This looks all right to me, like the other blocks that go multi-line

spec/Section 5 -- Validation.md
```diff
@@ -675,16 +728,34 @@ fragment argOnRequiredArg on Dog {
...
 }
+
+fragment withFragmentArg($command: DogCommand) on Dog {
+  doesKnowCommand(dogCommand: $command)
+}
+
+fragment usesFragmentArg on Dog {
+  ...withFragmentArg(command: DOWN)
+}
 ```
```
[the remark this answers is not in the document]
JoviDeCroock: Added in https://github.com/graphql/graphql-spec/pull/1081/commits/969f692fdcc466e9b9c730a624f567361080b6cf#diff-607ee7e6b71821eecadde7d92451b978e8a75e23d596150950799dc5f8afa43eR719

spec/Section 5 -- Validation.md
JoviDeCroock: Added!

spec/Section 5 -- Validation.md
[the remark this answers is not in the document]
benjie: Probably a bug in Refined GitHub, they do mess up the rendering of diffs from time to time. Closing.

spec/Section 5 -- Validation.md
```diff
@@ -1636,13 +1734,14 @@ query takesCatOrDog($catOrDog: CatOrDog) {
     transitively.
   - For each {fragment} in {fragments}:
     - For each {variableUsage} in scope of {fragment}, variable must be in
-      {operation}'s variable list.
+      {fragment}'s or {operation}'s variable list.
```
benjie, line 1737: Might be worth making it clear this is a "logical or" (allowing both to be true) rather than an "exclusive or" (XOR) which is more commonly how a layman interprets the word `or`.

spec/Section 5 -- Validation.md
```diff
@@ -1669,9 +1768,10 @@ query variableIsNotDefined {
 ${atOtherHomes} is not defined by the operation.
 
 Fragments complicate this rule. Any fragment transitively included by an
-operation has access to the variables defined by that operation. Fragments can
-appear within multiple operations and therefore variable usages must correspond
-to variable definitions in all of those operations.
+operation has access to the variables defined by that operation and defined on
+the fragment. Fragments can appear within multiple operations and therefore
```
benjie, line 1772: This is kind of fine as-is but reads a little ambiguously if the read is not careful (as if the variable has to be defined in both places). Adding `also those` seems to clarify this:

```suggestion
operation has access to the variables defined by that operation and also those defined on
the fragment. Fragments can appear within multiple operations and therefore
```

spec/Section 5 -- Validation.md
```diff
@@ -1828,6 +1928,29 @@ fragment isHouseTrainedWithoutVariableFragment on Dog {
 }
 ```
 
+Fragment arguments can shadow operation variables: fragments that use an
+argument are not using the operation-defined variable of the same name.
+
+Likewise, it would be invalid if the variable was shadowed by a fragment
+argument:
```
benjie, line 1935: This is really hard to express. Perhaps something like:

```suggestion
As such, it would be invalid if the operation defined a variable and variables of that name were used exclusively inside fragments that define a variable with the same name:
```

spec/Section 5 -- Validation.md
```diff
@@ -1853,6 +1976,39 @@ fragment isHouseTrainedFragment on Dog {
...
 variable.
 
+### All Fragment Variables Used
+
+**Formal Specification**
+
+- For every {fragment} in the document:
+  - Let {variables} be the variables defined by that {fragment}.
+  - Each {variable} in {variables} must be used at least once in the fragment's
+    scope.
```
benjie, line 1986: I think `scope` might be poorly defined; e.g. it could be interpretted to include fragments spread within it. Perhaps:
```suggestion
  - Each {variable} in {variables} must be used at least once transitively within the fragment's selection set excluding traversal of named fragment spreads.
```

benjie (changes requested):
Great work Jovi! Here's a few comments, but this was only a quick review so it's not exhaustive.

spec/Section 5 -- Validation.md
```diff
@@ -677,14 +715,42 @@ fragment argOnOptional on Dog {
 }
 ```
 
-the following is invalid since `command` is not defined on `DogCommand`.
+The above is also applicable to fragment-definitions and fragment-spreads, each
+variable must be defined by the fragment-definition before it can be inserted as
+an argument by the fragment-spread.
```
JoviDeCroock, line 720:
```suggestion
The above is also applicable to fragment definitions and fragment spreads, each
variable must be defined by the fragment definition before it can be inserted as
an argument by the fragment spread.
```

spec/Section 5 -- Validation.md
```diff
@@ -570,6 +577,35 @@ fragment conflictingDifferingResponses on Pet {
...
+fragment potentiallyConflictingArguments(
+  $commandOne: DogCommand!
+) on Dog {
+  ...commandFragment(command: $commandOne)
+  ...commandFragment(command: $commandTwo)
+}
+
+fragment safeFragmentArguments on Dog {
+  ...potentiallyConflictingArguments(commandOne: SIT, commandTwo: DOWN)
+}
```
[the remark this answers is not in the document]
JoviDeCroock: I tried my best to make it a bit better 😅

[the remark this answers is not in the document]
benjie: Here we call the fragment with two arguments:
```graphql
query {
  pet {
    ...potentiallyConflictingArguments(commandOne: SIT, commandTwo: DOWN)
  }
}
```

The fragment defines only one variable, so the other variable, `$commandTwo`, is assumed to be global. This does not align with how it is called above.

```graphql
fragment potentiallyConflictingArguments(
  $commandOne: DogCommand!
) on Dog {
  ...commandFragment(command: $commandOne)
  ...commandFragment(command: $commandTwo)
}
```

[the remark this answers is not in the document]
JoviDeCroock: Oh gotcha, sorry I missed that

spec/Section 5 -- Validation.md
```diff
@@ -570,6 +577,37 @@ fragment conflictingDifferingResponses on Pet {
...
 
+Fragment spread arguments can also cause fields to fail to merge.
+
+```graphql example
+fragment commandFragment($command: DogCommand!) on Dog {
+  doesKnowCommand(dogCommand: $command)
+}
+
+fragment potentiallyConflictingArguments(
+  $commandOne: DogCommand!
```
JoviDeCroock, line 588:
```suggestion
  $commandOne: DogCommand!
  $commandTwo: DogCommand!
```

JoviDeCroock:
@benjie I did not, the commits it points at are code-suggestions I applied so maybe it's not good at the co-authored ones? Not sure 😅

benjie:
@mjmahone Do you feel this replaces #1010? If so, I propose we close 1010 and move the RFC1 label to this PR.

@JoviDeCroock excellent work; to progress this further you should bring it to an upcoming Spec WG, and state the intent to move it from RFC1 to RFC2.

spec/Section 5 -- Validation.md
```diff
@@ -570,6 +577,38 @@ fragment conflictingDifferingResponses on Pet {
 }
 ```
 
+Fragment spread arguments can also cause fields to fail to merge.
+
+```graphql example
```
benjie, line 582: This example is explicitly not valid, right?

````suggestion
```graphql counter-example
````

File: spec/Section 5 -- Validation.md [hunk 25]
```diff
@@ -570,6 +577,38 @@ fragment conflictingDifferingResponses on Pet {
...
+
+If two fragment spreads with the same name, and hence the same selection, supply
+different argument values, their fields will not be able to merge. In this case,
+validation fails because the fragment spread `...commandFragment(command: SIT)`
+and `...commandFragment(command: DOWN)` are part of the visited selections that
+will be merged.
+
+If both of these spreads would have `$commandOne` or `$commandTwo` as the
+argument-value, it would be allowed as we can be sure that we'd resolve
+identical fields.
```
benjie, line 610: `or` here would not be interpreted as an exclusive or. Let's be more explicit:

```suggestion
If both of these spreads had used `$commandOne` as the argument value, it would
be allowed as we can be sure that we'd resolve identical fields.
```

Though of course it would not be valid since then `potentiallyConflictingArguments` would define an argument that wasn't used. Instead, maybe add an example that is valid?

spec/Section 5 -- Validation.md
```diff
@@ -418,8 +418,15 @@ fragment directFieldSelectionOnUnion on CatOrDog {
...
-  {set} including visiting fragments and inline fragments.
+- Let {visitedSelections} be the selections in {set} including visiting fields,
+  fragment-spreads and inline fragments.
+- Let {spreadsForName} be the set of fragment spreads with a given name in
+  {visitedSelections}.
+- For each {spreadsForName} as {name} and {spreads}:
+  - Each entry in {spreads} must have identical sets of arguments to each other
+    entry in {spreads}.
+- Let {fieldsForName} be the set of field selections with a given response name
+  in {visitedSelections}.
```
benjie, line 429: The `{fieldA} and {fieldB} must have identical sets of arguments.` line a few lines below this may be problematic. For example:

```graphql
query Q ($dogCommand: DogCommand!) {
  pet {
    doesKnowCommand(dogCommand: $dogCommand)
    ...frag(command: $dogCommand)
  }
}

fragment frag($command: DogCommand!) on Dog {
  doesKnowCommand(dogCommand: $command)
}
```

Here you have two visited fields:

- `doesKnowCommand(dogCommand: $dogCommand)`
- `doesKnowCommand(dogCommand: $command)`

The first uses the operation variable `$dogCommand`. The second uses the local fragment variable `$command` which technically resolves to the operation variable `$dogCommand` in this usage. I think this query should be seen as valid, so I think that fragment arguments should be resolved away during validation; either to a literal value, or to the underlying operation variable. This does *not* require that you know the variable _values_ during validation (we must not require this!), only that you know the name of the operation variable that would be used.

spec/Section 5 -- Validation.md
benjie: Another way to potentially make this valid is to call it as `...potentiallyConflictingArguments(commandOne: SIT, commandTwo: SIT)` or even `...potentiallyConflictingArguments(commandOne: $dogCommand, commandTwo: $dogCommand)` where `$dogCommand` is an operation variable.

benjie (changes requested):
I think there's more subtlety that needs to be addressed during validation.

spec/Section 5 -- Validation.md
benjie: Essentially, I think what I'm saying is that fragment arguments could be seen entirely as syntactic sugar, and could be entirely resolved via AST transform on the client side (or a server middleware) with no difference in observed behavior on the server. Not that they _should_ be, but that they _could_ be.

JoviDeCroock:
Update here, Implemented validation and everything, can be linked through from [here](https://github.com/graphql/graphql-js/pull/4015)

spec/Section 5 -- Validation.md
JoviDeCroock: Made a proof of concept here https://github.com/JoviDeCroock/graphql-js/pull/5/commits/755dd5af6317feb4b05eba4fd589ee5504907694

benjie: > if I am hearing you correctly there would be a step before validation and after parse where we effectively stub in the argument-value or variable reference?

To be clear, in spec terms, this probably wouldn't happen. Just that if someone were to do this, it would be equivalent. If we use that as a mental model, it might be clarifying around all the edge cases, is what I'm thinking - and would enforce a kind of consistency.

Urigo:
For those who has a production use case for this, wants to use it and provide feedback, feel free to reach out to @the-guild-org and test it with our support using [Yoga Server](https://the-guild.dev/graphql/yoga-server)

JoviDeCroock:
Changes in https://github.com/JoviDeCroock/graphql-spec/pull/1 are blocked by https://github.com/graphql/graphql-spec/pull/1039

leebyron:
@JoviDeCroock some significant changes have landed (including #1039) -- do you want to do a main merge to bring this PR up to date?

I'm also curious what remaining changes are necessary within GraphQL.js (beyond just making the flag to enable it default-on) to align to the spec?

JoviDeCroock:
@leebyron This might take me a while to get rebased 😅 I'll try my best when I find time.

GraphQL.JS is up to date with the spec in v17

spec/Section 6 -- Execution.md
```diff
@@ -712,6 +720,134 @@ A correct executor must generate the following result for that _selection set_:
 }
 ```
 
+### Field Collection
```
JoviDeCroock, line 723: I need to reconcile this with the new field collection algorithm, this is currently duplicated due to the merge commit

gramslawllc18 (approved):

JoviDeCroock:
Ah, damn, I just resolved a conflict this morning but another one :(

spec/Section 5 -- Validation.md
```diff
@@ -556,8 +556,15 @@ fragment directFieldSelectionOnUnion on CatOrDog {
 
 FieldsInSetCanMerge(set):
 
-- Let {fieldsForName} be the set of selections with a given _response name_ in
-  {set} including visiting fragments and inline fragments.
+- Let {visitedSelections} be the selections in {set} including visiting fields,
+  fragment-spreads and inline fragments.
```
benjie: If we visit fields, we'll get the entire tree. We only want to visit the fragment spreads/inline fragments.

benjie (changes requested):
This is an excellent start, I've gone through and done a full review of sections 2 and 5 and applied a number of editorial changes and clarifications to the text to align it more with the conventions of the spec and ensure that any ambiguities or edge cases are clarified; since there was a lot of edits I've raised them as a separate pull request:

- https://github.com/JoviDeCroock/graphql-spec/pull/2

I've not yet fully reviewed section 6, but I see some of the edits there may overlap with the incremental delivery work (that needed similar changes to the collect fields algorithm) so it would be worth making sure that these align. @robrichard and I can probably work through that and make it consistent.

Thanks for your work on this!

spec/Section 5 -- Validation.md
```diff
@@ -713,6 +720,58 @@ fragment conflictingDifferingResponses on Pet {
...
+  ...commandFragment(command: $operationCommand)
+  ...commandFragment(command: $fragmentCommand)
+}
+
+query($operationCommand: DogCommand!) {
+  pet {
+    ...noConflictWhenPassedOperationCommand(fragmentCommand: $operationCommand)
+  }
+}
+```
```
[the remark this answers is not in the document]
benjie: Fixed in my PR below

spec/Section 5 -- Validation.md
benjie: Fixed in my PR below

benjie:
Thanks for your work on this, @JoviDeCroock! I've re-opened it in #1224 to ensure the work isn't lost since it's RFC2; I've marked it as "needs champion". It already feels like it's fairly close to the finish line thanks to your hard work!

PR #1089, opened 2024-04-02
3.13 Directive validation edits

jbellenger:
The Validation section for Directive types is a little out of step with the validation sections for the other types. In particular, while the grammar definition of Directive requires one or more directive locations, this requirement isn't mentioned in the validation section.

Compare this to the validation sections for other types with "one or more" requirements:
- `1. An Object type must define one or more fields.`
- `1. A Union type must include one or more member types.`
- `1. An Enum type must define one or more unique enum values.`
- `1. An Input Object type must define one or more input fields.`
- `1. An Interface type must define one or more fields.`

This PR updates Directive validation to be more consistent with the validation sections of other types:
1. Add a "one or more" requirement as the first validation rule
1. Use "Directive" instead of "directive"
1. Rename the section from "Validation" to "Type Validation"

While this PR just clarifies the existing spec, I suspect it might be common for implementations to not check for non-empty directive locations. 
I've [updated graphql-java](https://github.com/graphql-java/graphql-java/pull/3552) and can do the same for graphql-js if I get a tentative OK on this PR.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -2040,7 +2040,7 @@ repeatable directives.
    references itself directly.
 3. A Directive definition must not contain the use of a Directive which
    references itself indirectly by referencing a Type or Directive which
-   transitively includes a reference to this directive.
+   transitively includes a reference to this Directive.
 4. The Directive must not have a name which begins with the characters {"\_\_"}
    (two underscores).
 5. For each argument of the Directive:


spec/Section 3 -- Type System.md
```diff
@@ -2033,19 +2033,20 @@ directive @invalidExample(arg: String @invalidExample) on ARGUMENT_DEFINITION
 Note: The order in which directives appear may be significant, including
 repeatable directives.
 
-**Validation**
+**Type Validation**
 
-1. A directive definition must not contain the use of a directive which
+1. A Directive definition must include at least one DirectiveLocation.
```
jbellenger, line 2038: **_Why is this worded differently from the other "1 or more" constraints?_**
If this were worded in the same style as Objects, it would read "A Directive must define one or more DirectiveLocations"

To be ultra pedantic, while this sentence attempts to use a plural form of the DirectiveLocation token, it can also be interpreted as a plural form of DirectiveLocation**s** tokens, which has a distinct definition in Appendix B.4.

I thought that this phrasing was less ambiguous, though recognize there's also value in using consistent language.  I'm open to feedback.

benjie: We're also generally okay with `must include one or more DirectiveLocation`.

benjie:
> I've https://github.com/graphql-java/graphql-java/pull/3552 and can do the same for graphql-js if I get a tentative OK on this PR.

@jbellenger Please go ahead and make the changes to GraphQL.js :raised_hands:

dondonz:
Nice! Great to see the spec clarified

michaelstaib (approved):

benjie:
GraphQL.js backport into v16 open here: https://github.com/graphql/graphql-js/pull/4137

benjie:
Scheduled for discussion in Thursday's WG: https://github.com/graphql/graphql-wg/blob/main/agendas/2024/07-Jul/18-wg-primary.md

benjie:
cc @graphql/tsc: TL;DR: merging this in one week unless anyone raises concerns.

This was discussed at last nights WG ([replay](https://youtu.be/1nl88fdbcVk); [agenda](https://github.com/graphql/graphql-wg/blob/main/agendas/2024/07-Jul/18-wg-primary.md); [notes](https://github.com/graphql/graphql-wg/blob/main/notes/2024/2024-07.md)) and it was agreed this is an editorial fix and good to merge. It already has two TSC approvals but I'll leave it open another week or so and then go ahead and merge it. Essentially, we see this as a bugfix.

benjie:
Oops :flushed: forgot to press the merge button. Thanks Lee!

PR #1091, opened 2024-04-05
Define Data Collections used in the spec

benjie:
The spec uses the terms "set", "list" and "map" in many places. This PR defines these terms and their conformance requirements.

This is a change extracted from #1063 as discussed at [last night's WG meeting](https://github.com/graphql/graphql-wg/blob/main/notes/2024/2024-04.md#ordering-of-schema-elements-10m-benjie)

The next PR in this stack is:
- https://github.com/graphql/graphql-spec/pull/1092

cc @graphql/tsc

PR #1092, opened 2024-04-05
Recommend that order of unordered collections is maintained where possible

benjie:
For legibility and consistency reasons, it's helpful if the order of unordered collections is maintained. For example, an enum describes a set (unordered) of possible values, but it would be irritating if every time you introspected the API or built an SDL for it the enum values were in a different (random) order. This PR recommends that you maintain the initial order that such unordered collections were created with. Theoretically, this means that if you generate SDL from a schema (e.g. via introspection) and then build a new cloned schema from this SDL, generating SDL from this cloned schema (e.g. via introspection) should match the original SDL (i.e. `printSchema(schema) == printSchema(buildClientSchema(introspectionFromSchema(schema)))`)

This PR builds on #1091 and is a change extracted from #1063 as discussed at [last night's WG meeting](https://github.com/graphql/graphql-wg/blob/main/notes/2024/2024-04.md#ordering-of-schema-elements-10m-benjie). It does need to be stacked since it also adds the ordering details to Appendix A (originally we thought it might be able to be separate).

This is PR 2 of 3 in the "data collections" trilogy stack

The previous PR in this stack is:

1. #1102

The next PR in this stack is:

3. #1063

### DIFF
--- a/spec/Appendix B -- Notation Conventions.md
+++ b/spec/Appendix B -- Notation Conventions.md
@@ -249,7 +249,7 @@ ordered map, which does not have an entry with that key, is ordered after
 existing entries.
 
 Note: This specification defines ordered data collection only when strictly
-necessary. When an order is observable, implementations should preserve it to
+required. When an order is observable, implementations should preserve it to
 improve output legibility and stability. For example if applying a grammar to an
 input string yields a _set_ of elements, serialization should emit those
 elements in the same source order.


benjie:
This was split from RFC2 PR #1063, so I have marked it as RFC 2 itself.

JoviDeCroock (approved):

mjmahone (approved):

michaelstaib (approved):

PR #1098, opened 2024-05-27
Add 5.2.4 Operation Type Exists

### LINKED ISSUE #1097: No validation rule exists to assert mutation/subscription operations exist in the schema
When validating a document against a schema, there exists various validation rules for operations, but none to verify that if a mutation or subscription operation type is requested, that the schema is configured for such an operation.

For example, the below document would current pass all defined validation checks even with no mutation operation defined in the schema:

```gql
mutation {
  __typename
}
```

While per 6.2.2 the execution phase would fail the assertion that "mutationType is an Object type" and would fail to execute, this problem should be caught during validation, not execution.

Note that in the above sample, only 5.3.1 Field Selections could possibly fail validation.  For GraphQL.NET, each validation rule assumes that prior validation steps have passed.  For example, we do not raise errors about invalid fields when the parent field does not exist.  So no error is raised due to 5.3.1.  Further, `__typename` is a meta-field and would be not be compared to the schema in any case.  We perform the assertion at the execution step pursuant to the GraphQL spec.

I suggest adding validation rule 5.2.4 indicating that for mutation or subscription operation types, the schema must be configured for the requested operation.

Shane32:
Closes #1097

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -332,6 +332,37 @@ contain any number of operations, each of which may contain different root
 fields. When executed, a document containing multiple subscription operations
 must provide the operation name as described in {GetOperation()}.
 
+### 5.2.4 Operation Type Configuration
+
+#### Formal Specification
+
+For each operation definition in the document:
+- Let operationType be the type of the operation (query, mutation, or subscription).
+- The corresponding type for operationType must be defined in the schema.
+
+#### Explanatory Text
+
+The schema must be properly configured to handle mutation and subscription operation
+types. This ensures that when a mutation or subscription operation is defined in the
+document, the schema includes the appropriate type to handle these operations.
+
+While query operations are required for all schemas, mutation and subscription operations
+are optional. If the schema does not include the necessary type for a mutation or subscription
+operation defined in the document, it will be considered invalid.
+
+For example, the following document is valid if the schema includes a Mutation type, but
+invalid if it does not:
+
+```graphql
+mutation {
+  likeStory(storyID: 12345) {
+    story {
+      likeCount
+    }
+  }
+}
+```
+
 ## Fields
 
 ### Field Selections
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -332,20 +332,20 @@ contain any number of operations, each of which may contain different root
 fields. When executed, a document containing multiple subscription operations
 must provide the operation name as described in {GetOperation()}.
 
-### Operation Type Configuration
+### Operation Type Exists
 
 #### Formal Specification
 
 For each operation definition in the document:
 
-- Let {operationType} be the type of the operation (query, mutation, or subscription).
-- The corresponding type for {operationType} must be defined in the schema.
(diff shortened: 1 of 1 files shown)


spec/Section 5 -- Validation.md
```diff
@@ -332,6 +332,37 @@ contain any number of operations, each of which may contain different root
 fields. When executed, a document containing multiple subscription operations
 must provide the operation name as described in {GetOperation()}.
 
+### 5.2.4 Operation Type Configuration
```
Shane32, line 335:
```suggestion
### Operation Type Configuration
```

spec/Section 5 -- Validation.md
```diff
@@ -332,6 +332,37 @@ contain any number of operations, each of which may contain different root
 fields. When executed, a document containing multiple subscription operations
 must provide the operation name as described in {GetOperation()}.
 
+### Operation Type Configuration
+
+#### Formal Specification
+
+For each operation definition in the document:
+- Let operationType be the type of the operation (query, mutation, or subscription).
+- The corresponding type for operationType must be defined in the schema.
```
Shane32, line 341:
```suggestion
- Let {operationType} be the type of the operation (query, mutation, or subscription).
- The corresponding type for {operationType} must be defined in the schema.
```

spec/Section 5 -- Validation.md
```diff
@@ -332,6 +332,37 @@ contain any number of operations, each of which may contain different root
 fields. When executed, a document containing multiple subscription operations
 must provide the operation name as described in {GetOperation()}.
 
+### Operation Type Configuration
+
+#### Formal Specification
+
+For each operation definition in the document:
```
Shane32, line 339:
```suggestion
For each operation definition in the document:

```

benjie (dismissed):
Great find!

benjie (approved):
Looks good to me, though some slight editorial on the wording to match the style of the spec as a whole might be warranted.

JoviDeCroock (approved):
This is a great validation rule 🙌

benjie:
The lack of this validation rule can be demonstrated in GraphQL.js with this:

```ts
import { GraphQLSchema, GraphQLInt, graphql, GraphQLObjectType } from 'graphql';

  name: 'Query',
  fields: {
    a: {
      type: GraphQLInt,
      resolve() {
        return 42;
      },
    },
  },
});

  query: Query,
});

const result = await graphql({ schema, source: `mutation { __typename }` });
console.dir(result);
```

which produces:

```
{
  data: null,
  errors: [
    GraphQLError: Schema is not configured to execute mutation operation.
        at executeOperation [...]
      path: undefined,
      locations: [Array],
    }
  ]
}
```

We should really mark this as an invalid request, and thus it should not have a `data` key in the response.

yaacovCR:
This is a refresh of #955 which has a pending PR at the implementation https://github.com/graphql/graphql-js/pull/3601

benjie:
Good spot @yaacovCR; it seems both @Shane32 and I had commented on that previous PR!

I have merged this PR into that PR and applied my own editorial edits on top of both:

- https://github.com/graphql/graphql-spec/pull/955

PR #1099, opened 2024-06-06
Editorial changes for Event Streams

leebyron:
This makes a few changes to the Subscriptions section where we're talking about event streams in an attempt to make it more clear about what's going on.

- Revised variable names from "fieldStream" to "sourceStream" to make it easier to trace variables through algorithms.

- Rewrote the "Event Streams" definition to be more clear about "emit" keyword and have clear paragraphs on completion and cancellation.

- Rewrote the `MapSourceToResponseEvent` algorithm to be a correctly formatted algorithm with a return statement at the end. Introduced a new "When" keyword to describe event subscriptions. Added explicit sections on passing back cancellation (discussed in WG) as well as completion with error (not discussed, but I realized was also left ambiguous)

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -217,14 +217,18 @@ chat room ID is the "topic" and each "publish" contains the sender and text.
 
 **Event Streams**
 
-An event stream represents a sequence of discrete events over time which can be
-observed. As an example, a "Pub-Sub" system may produce an event stream when
-"subscribing to a topic", with an event occurring on that event stream for each
-"publish" to that topic. Event streams may produce an infinite sequence of
-events or may complete at any point. Event streams may complete in response to
-an error or simply because no more events will occur. An observer may at any
-point decide to stop observing an event stream by cancelling it, after which it
-must receive no more events from that event stream.
+An event stream represents a sequence of discrete emitted events over time which
+can be observed. As an example, a "Pub-Sub" system may produce an event stream
+when "subscribing to a topic", with an event emitted for each "publish" to that
+topic.
+
+Event streams may complete at any point, often because no further events will
+occur. Event streams may emit an infinite sequence of events, in which they may
+never complete. If an event stream encounters an error, it must complete with
+that error.
+
+An observer may at any point decide to stop observing an event stream by
+cancelling it. When an event stream is cancelled, it must complete.
 
 **Supporting Subscriptions at Scale**
 
@@ -268,10 +272,10 @@ CreateSourceEventStream(subscription, schema, variableValues, initialValue):
 - Let {field} be the first entry in {fields}.
 - Let {argumentValues} be the result of {CoerceArgumentValues(subscriptionType,
   field, variableValues)}.
-- Let {fieldStream} be the result of running
+- Let {sourceStream} be the result of running
   {ResolveFieldEventStream(subscriptionType, initialValue, fieldName,
   argumentValues)}.
-- Return {fieldStream}.
+- Return {sourceStream}.
 
 ResolveFieldEventStream(subscriptionType, rootValue, fieldName, argumentValues):
 
@@ -292,12 +296,18 @@ subscription _selection set_ using that event as a root value.
 
 MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
 
-- Return a new event stream {responseStream} which yields events as follows:
-  - For each {event} on {sourceStream}:
-    - Let {response} be the result of running
-      {ExecuteSubscriptionEvent(subscription, schema, variableValues, event)}.
-    - Yield an event containing {response}.
-  - When {sourceStream} completes: complete {responseStream}.
+- Let {responseStream} be a new event stream.
+- When {sourceStream} emits {event}:
+  - Let {response} be the result of running
+    {ExecuteSubscriptionEvent(subscription, schema, variableValues, event)}.
+  - Emit {response} on {responseStream}.
+- When {sourceStream} completes normally:
+  - Complete {responseStream} normally.
+- When {sourceStream} completes with {error}:
+  - Complete {responseStream} with {error}.
+- When {responseStream} is cancelled:
+  - Cancel {sourceStream}.
+- Return {responseStream}.
 
 ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
 
@@ -317,9 +327,9 @@ Note: The {ExecuteSubscriptionEvent()} algorithm is intentionally similar to
 #### Unsubscribe
 
 Unsubscribe cancels the Response Stream when a client no longer wishes to
-receive payloads for a subscription. This may in turn also cancel the Source
-Stream. This is also a good opportunity to clean up any other resources used by
-the subscription.
+receive payloads for a subscription. This in turn also cancels the Source
+Stream, which is a good opportunity to clean up any other resources used by the
+subscription.
 
 Unsubscribe(responseStream):
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -164,7 +164,7 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
 
 ### Subscription
 
-If the operation is a subscription, the result is an event stream called the
+If the operation is a subscription, the result is an _event stream_ called the
 "Response Stream" where each event in the event stream is the result of
 executing the operation for each new event on an underlying "Source Stream".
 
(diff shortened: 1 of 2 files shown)


benjie (commented):
I think this is a massively clarifying PR!

I got hung up for a bit on the three verbs ("complete normally", "complete with error" and "cancel") and in particular trying to figure out the practical difference between "complete normally" and "cancel", but I think they ultimately make sense. My main concern is ensuring we've handled all the situations:

| stream | event | handler |
|---|---|---|
|sourceStream|completes normally|complete responseStream normally|
|sourceStream|completes with error|complete responseStream with error|
|sourceStream|is cancelled|unhandled, since can only occur from responseStream being cancelled :grey_question: |
|sourceStream|encounters error|forces complete with error -> handled above|
|responseStream|completes normally|unhandled, since can only occur from sourceStream completing|
|responseStream|completes with error|unhandled, occurs from sourceStream completing with error|
|responseStream|is cancelled|cancel sourceStream|
|responseStream|encounters error|forces complete with error -> unhandled above|

Is there a situation in which responseStream might encounter an error (other than from sourceStream encountering an error)? If so, this is currently unhandled, I think?

Are there any other ways for sourceStream to be cancelled? I assume user code cancelling it would be seen by us as "completes normally"? (If so, this might warrant a non-normative note?)

benjie:
Oh I merged in the latest `main` by the way.

JoviDeCroock (approved):

spec/Section 6 -- Execution.md
```diff
@@ -317,9 +327,9 @@ Note: The {ExecuteSubscriptionEvent()} algorithm is intentionally similar to
 #### Unsubscribe
 
 Unsubscribe cancels the Response Stream when a client no longer wishes to
-receive payloads for a subscription. This may in turn also cancel the Source
-Stream. This is also a good opportunity to clean up any other resources used by
-the subscription.
+receive payloads for a subscription. This in turn also cancels the Source
+Stream, which is a good opportunity to clean up any other resources used by the
+subscription.
```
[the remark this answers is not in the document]
leebyron: Exactly, the previous wording was too loose. Ultimately this is not observable behavior from an API consumer's point of view, so also not spec enforceable, but I think this language is much more clear. If a resulting stream consumer cancels, then it's certainly the case that the source stream must be made aware of that.

Extra unnecessary detail in case if ever comes up: if an implementer doesn't *want* to cancel the source stream, they should be using an intermediate like [tee](https://en.wikipedia.org/wiki/Tee_(command))

spec/Section 6 -- Execution.md
```diff
@@ -287,17 +294,33 @@ operation type.
...
-    - Yield an event containing {response}.
-  - When {sourceStream} completes: complete {responseStream}.
+- Let {responseStream} be a new _event stream_.
+- When {sourceStream} emits {sourceValue}:
+  - Let {response} be the result of running
+    {ExecuteSubscriptionEvent(subscription, schema, variableValues,
+    sourceValue)}.
+  - If internal {error} was raised:
+    - Cancel {sourceStream}.
+    - Complete {responseStream} with {error}.
```
leebyron, line 309: See the note below, but I checked with graphql-js and this aligns. Request errors occur before field execution begins, which for subscriptions would be before source stream is created. Then each event through the subscription could produce field errors, but those are handled normally as `{ data, errors }` and shouldn't complete the subscription. The only remaining scenario is some internal exceptional event.

I thought about continuing to leave this omitted, since it makes sense why it was omitted before, but I do agree that this is more clear about what to do in this non-normative scenario, and will make it more clear in the future if we ever decide to formally add "fatal error" as an error type

benjie (approved):
This is much more descriptive and precise; excellent work! :raised_hands:

n1ru4l (approved):

spec/Section 6 -- Execution.md
yaacovCR: I'm not 100% clear on what kind of internal error is being envisioned here. It seems like it's clearly not a request or field error, as explicitly stated above.

Is this analogous to the case of the `graphql-js` call to `asyncIterator.next()` resulting in an exception? I would think that's covered below when discussing "completing in error" but maybe that's not the case?

Of course, anything could error => maybe it just means "something else went wrong?" And we are making how to handle that more explicit in the case of a subscription than we would with regard to our other operation, because it is helpful to clients to understand how this case is handled in the context of the response stream?

leebyron: @yaacovCR that last bit you said is exactly right. The vast amount of our spec assumes algorithms are pretty typical functions which you call with arguments and return a value, and the most significant complication is the introduction of async functions (which the spec really leaves mostly unstated), so the novelty here is the fact that this algorithm is not returning the result but returning a stream which will contain results. That's important because for a typical function we can just say if there is some exceptional error in any given algorithm that it should work its way up until something explicitly handles it or it fatals the whole operation - that's how error handling works (in some way or another) in just about every language and environment. 

However it's a lot less obvious what should happen for a stream which encounters an exceptional error after that particular creation algorithm has completed - what is the right behavior? This tries to be explicit to avoid that ambiguity - we should make sure the resulting stream also ends with an error.

> Is this analogous to the case of the graphql-js call to asyncIterator.next() resulting in an exception?

Yes - but specifically inside where it is being mapped. graphql-js handles field errors during execution of a single subscription payload, resulting in a `{data: null, errors}` normal payload, but other exceptions could trigger [this codepath](https://github.com/graphql/graphql-js/blob/75dca3dd18e5be3031416965049fbde48a5adb20/src/execution/mapAsyncIterable.ts#L27). That's really what's being described here

spec/Section 6 -- Execution.md
```diff
@@ -287,17 +294,33 @@ operation type.
...
+  - Otherwise emit {response} on {responseStream}.
+- When {sourceStream} completes normally:
+  - Complete {responseStream} normally.
+- When {sourceStream} completes with {error}:
+  - Complete {responseStream} with {error}.
```
yaacovCR, line 314: Does this wording imply that when the {responseStream} completes with error, it must do so in the same manner as when the {sourceStream} does? We [have a request at graphql-js](https://github.com/graphql/graphql-js/issues/4001) for emitting well formated `GraphQLError`s instead of throwing. I would read the specification permissively, as in the manner in which the {responseStream} completes with said error is up to the implementation, allowing that change.

Once could argue instead/additionally that it belongs within the transport protocol specification.

Tagging @leebyron @enisdenjo @aseemk from the linked issue.

enisdenjo: From the SSE perspective, and in the [GraphQL over SSE RFC](https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverSSE.md), errors are emitted over the stream as an event.

This is because browser native EventSource will not report back to the user non-200 responses in detail - it'll just say "error" - leaving the user in dark. Furthermore, if the error occurs after a while _during_ the stream, the connection is already established and the only way to report an issue is in form of an event.

Same thing applies for WebSockets.

benjie, line 314: I agree that for normal errors in `sourceStream` we should capture them and emit an `{ errors: [ ... ] }` payload rather than terminating with an error.

```suggestion
  - Let {errors} be a list containing {error}.
  - Let {response} be an unordered map containing {errors}.
  - Emit {response} on {responseStream}.
  - Complete {responseStream} normally.
```

However, this is a change in behavior (not an editorial change like the rest of this PR), so perhaps this should be raised _after_ this editorial only change is merged - see https://github.com/graphql/graphql-spec/pull/1127 for a suitable follow-up PR.

benjie: I agree that for normal errors in `sourceStream` we should capture them and emit an `{ errors: [ ... ] }` payload rather than terminating with an error.

```suggestion
  - Let {errors} be a list containing {error}.
  - Let {response} be an unordered map containing {errors}.
  - Emit {response} on {responseStream}.
  - Complete {responseStream} normally.
```

However, this is a change in behavior (not an editorial change like the rest of this PR), so perhaps this should be raised _after_ this editorial only change is merged - see https://github.com/graphql/graphql-spec/pull/1127 for a suitable follow-up PR.

martinbonnin: > I agree that for normal errors in sourceStream we should capture them and emit an { errors: [ ... ] } payload rather than terminating with an error.

Generally agree with using the GraphQL response structure to convey errors but that makes me question why we would need anything else. Is there any specific reason to keep the `Complete {responseStream} with {error}` [just above](https://github.com/graphql/graphql-spec/pull/1099/files#diff-0f02d73330245629f776bb875e5ca2b30978a716732abca136afdd028d5cd33cR314)?

benjie: I'm not certain, but I imagine that an internal error implies that something fundamental has gone wrong and thus we're no longer sure that we can even complete the request correctly - a bit like returning a "500 Internal Server Error" rather than a 4xx error in HTTP: "something really really went wrong and we're just giving up right here and exiting immediately". Perhaps we can't even be trusted to form a GraphQL response payload at this point - perhaps it was attempting to form such a payload that caused the internal error in the first place?

yaacovCR: https://github.com/graphql/graphql-spec/pull/1099#discussion_r1643169652

See also above where @leebyron explains….

martinbonnin: @yaacovCR I've seen that comment but not being super familiar with graphql-js, I'm not sure I understand all the subtleties of it. Once we have https://github.com/graphql/graphql-js/pull/4302, are there still places where graphql-js can throw? I thought this was the last one.

benjie: Should we move this discussion to #1127 as it does not seem to be editorial but a change in behavior?

martinbonnin: This is also very related to https://github.com/graphql/graphql-js/issues/4001#issuecomment-2506524425. Happy to consolidate everything to #1127

enisdenjo: _*Not really making a point, just giving context from the transport perspective._

When dealing with HTTP request-response format, we have 2 places conveying information (as per GQL over HTTP): the body (_data_) and the headers (_metadata_). As @benjie said, 4XX and 5XX would indicate the severity of the errors in _data_.

However, in streaming transports this is different. When dealing with both WS and SSE, browser native implementations (WebSocket and EventSource respectively) give out less information about termination of requests. WebSocket's close event is limited in size, so you cannot send a big error message; and EventSource will tell you absolutely nothing if the server responds with a non-200 status.

Multipart/incremental-delivery/custom-SSE are different because the user does the heavy-lifting (implementation of the fetcher and parser) and there you certainly can use headers for metadata.

leebyron:
Note from live meeting:

Merging this _editorial_ change since it accurately describes todays behavior. This merge doesn't close discussion on managing stream error conditions, but will allow us to move the focus of that to focused discussions (of which some are already open, example https://github.com/graphql/graphql-spec/pull/1127)

PR #1100, opened 2024-06-10
chore: Update cspell.yml

Jason3S:
The current cspell.yml file doesn't parse.

!!! IMPORTANT !!!

Please Read https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md
before creating a Pull Request.

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -23,5 +23,5 @@ words:
   - zuckerberg
 # Forbid Alternative spellings
 flagWords:
-  - implementor->implementer
-  - implementors->implementers  
+  - implementor
+  - implementors
--- a/cspell.yml
+++ b/cspell.yml
@@ -21,6 +21,7 @@ words:
   - tatooine
   - zuck
   - zuckerberg
-  # Alternative spellings
-  - !implementor
-  - !implementors
+# Forbid Alternative spellings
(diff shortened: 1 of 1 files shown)


Jason3S:
Hello @benjie,

Two things:
1. I noticed that the latest changes to `cspell.yml` do not parse as correct Yaml. This PR is to address that. This fix includes one of the newer CSpell features to make suggestions on the desired spelling. An alternative fix is to just surround `- !implementor` and `- !implementors` with quotes: `- '!implementor'`.
2. The version of CSpell is quite old. Is there a reason to stay on v5?

cspell.yml
```diff
@@ -21,6 +21,7 @@ words:
   - tatooine
   - zuck
   - zuckerberg
-  # Alternative spellings
-  - !implementor
-  - !implementors
+# Forbid Alternative spellings
+flagWords:
+  - implementor->implementer
+  - implementors->implementers
```
benjie, line 27: I couldn't find this syntax in the documentation and confirmed it didn't work locally; but this simpler form did:

```suggestion
  - implementor
  - implementors
```

benjie (approved):
Great catch; thanks!

PR #1102, opened 2024-07-01
Define Data Collections used in the spec

benjie:
The spec uses the terms "set", "list" and "map" in many places. This PR defines these terms and their conformance requirements.

_This is a change extracted from #1063 as discussed at [last night's WG meeting](https://github.com/graphql/graphql-wg/blob/main/notes/2024/2024-04.md#ordering-of-schema-elements-10m-benjie)_

This is PR 1 of 3 in the "data collections" trilogy stack; next are:

2. https://github.com/graphql/graphql-spec/pull/1092
3. https://github.com/graphql/graphql-spec/pull/1063

cc @graphql/tsc

_(Replacement of #1091; to do PR stack with #1092 and #1063)_


rahulkr-22 (approved):

spec/Appendix A -- Notation Conventions.md
```diff
@@ -216,3 +216,34 @@ Fibonacci(number):
...
+types like "list", "set" and "map". These describe observable data collections
+such as the result of applying a grammar and the inputs and outputs of
+algorithms. They also describe unobservable data collections such as temporary
+data internal to an algorithm. Each data collection type defines the operations
+available, and whether values are unique or ordered.
+
+**List**
+
+:: The term _list_ describes a sequence of values which may not be unique. A
+list is ordered unless explicitly stated otherwise (as an "unordered list"). For
```
andimarek, line 232: I am not sure "unordered list" makes sense. I mean the ability to iterate over something in well defined way, meaning it can be viewed as a sequence of elements, implies an order in my view.

You could maybe argue that unordered list means the order could change with each iteration, but not sure that is helpful in our context.

Maybe we can get rid of "unordered list"?

andimarek (commented):
I think one logical consistent way to describe these data structures is with a minimal amount of predefined terms would be:

1. Everything is a collection. A collection is a finite group of elements that can be iterated over. The encounter order during iteration is not necessarily stable.

2. A List is a specific collection where each element can be accessed via a stable Integer >= 0, called index and the encounter order is stable. A List allows for duplicate elements.

3. A set is a specific collection where the elements are guaranteed to be unique. The encounter is not necessary stable ("unordered set"), but can be ("ordered set") 
(If we need to define unordered set vs ordered)

4. A Map is a specific collection of elements, where each element is a pair (called "entry") made of a key and a value. All keys are a set, while all values is just a collection.
The encounter order of a Map can be stable ("ordered Map") or not ("unordered Map")
(If we need to define unordered Map vs ordered)

spec/Appendix A -- Notation Conventions.md
benjie: We don't have any examples of unordered lists being referenced the spec currently, as far as I know. Though [the List wrapper type](https://spec.graphql.org/draft/#sec-List) only dictates that the _serialized_ result is an ordered list, it doesn't specifically, as far as I know, require that the underlying list is ordered.

A common example of an unordered list that might be exposed through GraphQL would be the result of "select * from my_table;" in SQL - this SQL statement has no implicit ordering, and thus may be returned in any order, each time you issue the query you may get a different result even without the underlying data having changed.

I think we should reserve space to reference this concept.

martinbonnin: Do we need to reserve space now? It's not part of the query language or anything where backward compatibility would be super important. We could always reserve space if/when we need the concept? As of today, I agree it's not really helpful if not used in the spec itself.

spec/Appendix A -- Notation Conventions.md
```diff
@@ -216,3 +216,34 @@ Fibonacci(number):
...
+clarity the term "ordered list" may be used when an order is semantically
+important.
+
+**Set**
+
+:: The term _set_ describes a unique collection of values, where each value is
+considered a "member" of that set. A set is unordered unless explicitly stated
```
martinbonnin, line 239: Is it the collection that is unique or its content? Having "unique collection" feels like the former to me (but maybe it's a thing?).


Tentative disambiguation:
```suggestion
:: The term _set_ describes a unique collection of values. Each value is
considered a "member" of that set and must be unique in the set. A set is unordered unless explicitly stated
```

benjie:
Hopefully my latest edit factors in @andimarek and @martinbonnin's feedback.

spec/Appendix A -- Notation Conventions.md
```diff
@@ -216,3 +216,38 @@ Fibonacci(number):
 Note: Algorithms described in this document are written to be easy to
 understand. Implementers are encouraged to include equivalent but optimized
 implementations.
+
+## Data Collections
+
+A data collection is a finite, iterable aggregation of elements whose iteration
```
benjie, line 222: Why 'aggregation'?

- collection: tautology - "a data collection is a collection"
- set: we define "set" below, so using set here would introduce ambiguity
- group: evokes group theory, best avoided
- assembly: technically correct, but feels unnatural

martinbonnin (approved):

BoD (approved):

mjmahone (approved):

spec/Appendix A -- Notation Conventions.md
martinbonnin: Resolving this conversation as there are no "unordered lists" in the current version of the pull request.

spec/GraphQL.md
```diff
@@ -92,10 +92,14 @@ Conformance requirements expressed as algorithms can be fulfilled by an
...
-implementations.
+implementations. Similarly, data collections such as _list_, _set_ and _map_
+also introduce conformance requirements. Implementers are free to use
+alternative data collections as long as the perceived result remains equivalent.
 
 See [Appendix A](#sec-Appendix-Notation-Conventions) for more details about the
-definition of algorithms and other notational conventions used in this document.
+definition of algorithms and other notational conventions used in this document,
+and [Appendix A: Data Collections](#sec-Data-Collections) for specifics of data
+collections and their ordering.
```
leebyron, line 102: note to lee: simplify this addition a bit before merge

spec/Appendix A -- Notation Conventions.md
leebyron: I removed this in editorial.

My thinking is that any reader of this spec does not need to learn what a data collection is in general (though fwiw, wikipedia does use the term "group"). 

What is important for us to clearly state is what aspects of that data collection is normative

PR #1110, opened 2024-09-18
Incremental Delivery spec draft

robrichard:
Spec draft of `@defer` based on the response format in https://github.com/graphql/defer-stream-wg/discussions/69, with no duplication of fields. `@stream` will be added soon.

Please keep discussion limited to specifics on how the spec text is written. There is a dedicated discussion forum for questions, feedback, and comments on defer & stream: https://github.com/graphql/defer-stream-wg/discussions

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -276,14 +276,16 @@ CreateSourceEventStream(subscription, schema, variableValues, initialValue):
 - Let {subscriptionType} be the root Subscription type in {schema}.
 - Assert: {subscriptionType} is an Object type.
 - Let {selectionSet} be the top level selection set in {subscription}.
-- Let {collectedFieldsMap} be the result of {CollectFields(subscriptionType,
-  selectionSet, variableValues)}.
+- Let {collectedFieldsMap} and {newDeferUsages} be the result of
+  {CollectFields(subscriptionType, selectionSet, variableValues)}.
+- Assert: {newDeferUsages} is empty.
 - If {collectedFieldsMap} does not have exactly one entry, raise a _request
   error_.
 - Let {fields} be the value of the first entry in {collectedFieldsMap}.
-- Let {fieldName} be the name of the first entry in {fields}. Note: This value
-  is unaffected if an alias is used.
-- Let {field} be the first entry in {fields}.
+- Let {fieldDetails} be the first entry in {fields}.
+- Let {field} be the corresponding entry on {fieldDetails}.
+- Let {fieldName} be the field name of {field}. Note: This value is unaffected
+  if an alias is used.
 - Let {argumentValues} be the result of {CoerceArgumentValues(subscriptionType,
   field, variableValues)}.
 - Let {sourceStream} be the result of running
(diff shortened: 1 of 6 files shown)


owen-zora:
Hi @robrichard , do you have any expected timeline for this feature? We are really excited to have this feature for our product!

spec/Section 6 -- Execution.md
```diff
@@ -513,14 +905,24 @@ CollectFields(objectType, selectionSet, variableValues, visitedFragments):
...
-    - If {fragmentSpreadName} is in {visitedFragments}, continue with the next
-      {selection} in {selectionSet}.
-    - Add {fragmentSpreadName} to {visitedFragments}.
+    - If {fragmentSpreadName} provides the directive `@defer` and its {if}
+      argument is not {false} and is not a variable in {variableValues} with the
+      value {false}:
+      - Let {deferDirective} be that directive.
+      - If this execution is for a subscription operation, raise a _field
+        error_.
+    - If {deferDirective} is not defined:
```
yaacovCR, line 922: I think this line can be removed. See https://github.com/graphql/graphql-js/pull/4320

spec/Section 6 -- Execution.md
```diff
@@ -325,39 +315,405 @@ Unsubscribe(responseStream):
...
+
+First, the selection set is turned into a grouped field set; then, we execute
+this grouped field set and return the resulting {data} and {errors}.
+
+ExecuteRootSelectionSet(variableValues, initialValue, objectType, selectionSet,
+serial):
+
+- If {serial} is not provided, initialize it to {false}.
+- Let {groupedFieldSet} and {newDeferUsages} be the result of
+  {CollectFields(objectType, selectionSet, variableValues)}.
```
michaelstaib: The new `CollectFields` requires a `deferUsage` argument ... this should be `null` for the initial call to `CollectFields`.

```suggestion
  {CollectFields(objectType, selectionSet, null, variableValues)}.
```

Also in this case its not clear while reading what null actually is for so maybe it should be declared first?

duckki:
👋 I noticed a few minor corrections to make:

1. In Section 6, [Executing Fields](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%206%20--%20Execution.md#executing-fields), this line appears to be a typo:

   > Append `{fieldName}` to `{path}`.

   It should append the response name—an alias when present, otherwise the field name—because response paths use response names.

2. In Section 7, [Incremental Stream Update Result](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%207%20--%20Response.md#incremental-stream-update-result), the final sentence appears to repeat `true`:

   > The value of `{"hasNext"}` must be `{true}` for all but the last response in the incremental stream. Otherwise, `{"hasNext"}` must be `{true}`.

   The second value should be `{false}`. The wording could also be simplified to: “The value of `{"hasNext"}` must be `{true}` for every response except the last; it must be `{false}` for the last response.”

3. In Section 6, [Yielding Incremental Stream Results](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%206%20--%20Execution.md#yielding-incremental-stream-results) calls `GetPendingEntry` before [Mapping Work Events to Response Events](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%206%20--%20Execution.md#mapping-work-events-to-response-events), yet the latter initializes `idMap` and `nextID`. Since `GetPendingEntry` already calls `EnsureID`, the ID state should instead be initialized in `YieldIncrementalResults`, before its initial `GetPendingEntry` call.

duckki:
For @defer(label: null) or @stream(label: null), should pending include "label": null, or omit label?

Both directives declare a nullable [label: String](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%203%20--%20Type%20System.md#L2337-L2343). The [validation rule](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%205%20--%20Validation.md#L1816-L1827) says:

> If {label} does not exist or is {null}:  
> - Continue to the next {directive}.

I read that as exempting null from the uniqueness check, while still allowing the argument.

In [GetPendingEntry](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%206%20--%20Execution.md#L1207-L1220), the algorithm takes the group or stream’s {label} and constructs a pending entry containing it. The [response section](https://github.com/graphql/graphql-spec/blob/045e19363c2b55f127960bd3b5e8072a15b29aec/spec/Section%207%20--%20Response.md#L451-L453) says that if the directive “contains a label argument,” the pending notice “must contain an entry {"label"} with the value of this argument.” Together, these suggest that label: null produces "label": null in the pending notice.

Currently, GraphQL.js instead [omits the entry and tests for that behavior](https://github.com/graphql/graphql-js/blob/v17.0.1/src/execution/incremental/__tests__/defer-test.ts#L303-L323).
Should this behavior be documented in the spec, or GraphQL.js needs to be changed?

PR #1111, opened 2024-09-20
move Field Collection section earlier

yaacovCR:
Suggestion for #1039, see https://github.com/graphql/graphql-spec/pull/1039#issuecomment-2360902791


benjie (approved):
This is a more consistent order :+1:

PR #1113, opened 2024-09-24
Add note above variables to Values of Correct Type rule

yaacovCR:
Thought to add an explicit note emerged from a discussion about `@oneOf`.

See https://github.com/benjie/graphql-spec/pull/1#issuecomment-2368002543


benjie:
I've taken a slightly different approach to this here: https://github.com/graphql/graphql-spec/pull/1118; what do you think?

yaacovCR:
> I've taken a slightly different approach to this here: #1118; what do you think?

Looks good to me. Closing in favor of #1118.

yaacovCR:
Actually, I think this is not equivalent, see my comments there.

Issue #144, opened 2016-02-15
Use ancestors as field arguments


lucaslorentz:
Considering the following schema:

```
Event {
  id: String
  name: String
  speakers: [Speaker]
  talks: [Talk]
}
Speaker {
  id: String
  name: String
  events: [Event]
  talks: [Talk]
}
Talk {
  name: String
  speaker: Speaker
  event: Event
}
```

I would like to create queries like:

```
query {
   events {
     name
     speakers {
        name,
        talks {
          name
        }
     }
   }
}
```

I checked the specs and didn't find any way to filter the speaker's talks by the ancestor event.

I'm using directives on the examples below to demonstrate a feature GraphQL could have:

```
query {
   events {
     id @into (variable: "$eventId")
     name
     speakers {
        name,
        talks(eventId: $eventId) {
          name
        }
     }
   }
}
```

That would allow us to completely re-shape the query using the same schema:

```
query {
   speakers {
     id @into (variable: "$speakerId")
     name
     events {
        name,
        talks(speakerId: $speakerId) {
          id,
          name
        }
     }
   }
}
```

Or even go further:

```
query {
   events {
     id @into (variable: "$eventId")
     name
     speakers {
        name,
        talksOnThisEvent: talks(eventId: $eventId) {
          name
        },
        talksOnOtherEvents: talks(excludeEventId: $eventId) {
          name,
          event: {
            name
          }
        }
     }
   }
}
```

I believe such feature is a nice to have and would decouple even more GraphQL consumers, giving them the ability to completely re-group the data using any entry point: Event, Speaker

Any suggestions/opinions about that?

Related to: https://github.com/facebook/graphql/issues/88

dylanahsmith:
It looks like your schema already supports this type of query, which you could write as

```graphql
query {
   events {
     name
     talks {
       speaker {
        name
       }
     }
   }
}
```

leebyron:
Interesting! We've been kicking around a similar idea at FB for a while for running two related queries, but doing so within a single query is new.

Typically, we try to design our type systems to avoid context-sensitive requests like this in favor of context-free requests.

For example, in your first example, you seem to want to know all of the talks at a given event and the speakers for them, excluding talks that are not relevant to the original event. So your query:

```
query {
   events {
     id @into (variable: "$eventId")
     name
     speakers {
        name,
        talks(eventId: $eventId) {
          name
        }
     }
   }
}
```

should be re-writable as:

```
query {
  events {
    id
    name
    talks {
[...]

PR #1114, opened 2024-09-25
add __directive meta field parallel to type

yaacovCR:
Use case: inquiring about specific directives such as `@defer` and `@stream`

### DIFF
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -85,13 +85,14 @@ operation.
 
 ## Schema Introspection
 
-The schema introspection system is accessible from the meta-fields `__schema`
-and `__type` which are accessible from the type of the root of a query
-operation.
+The schema introspection system is accessible from the meta-fields `__schema`,
+`__type`, and `_directive` which are accessible from the type of the root of a
+query operation.
 
 ```graphql
 __schema: __Schema!
 __type(name: String!): __Type
+__directive(name: String!): __Directive
 ```
 
 Like all meta-fields, these are implicit and do not appear in the fields list in
--- a/spec/Section 4 -- Introspection.md
+++ b/spec/Section 4 -- Introspection.md
@@ -56,6 +56,38 @@ would produce the result:
 }
 ```
 
+Similarly, a request containing the operation:
+
+```graphql example
+{
+  __directive(name: "skip") {
+    name
+    arguments {
+      name
+      type {
+        name
+      }
+    }
+  }
+}
+```
+
+would produce the result:
+
+```json example
+{
+  "__directive": {
+    "name": "skip",
+    "arguments": [
+      {
+        "name": "if",
+        "type": { "name": "Boolean" }
+      }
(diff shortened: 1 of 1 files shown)


spec/Section 4 -- Introspection.md
```diff
@@ -85,13 +117,14 @@ operation.
 
 ## Schema Introspection
 
-The schema introspection system is accessible from the meta-fields `__schema`
-and `__type` which are accessible from the type of the root of a query
-operation.
+The schema introspection system is accessible from the meta-fields `__schema`,
+`__type`, and `_directive` which are accessible from the type of the root of a
```
benjie, line 121:
```suggestion
`__type`, and `__directive` which are accessible from the type of the root of a
```

yaacovCR:
Implemented by https://github.com/graphql/graphql-js/pull/4203

benjie:
I've not reviewed if these spec changes are sufficient, but I'm supportive of the aim :+1:

yaacovCR:
This was discussed at the November 2024 WG => this small change prompted an interesting discussion about "The Future of Introspection."

To my understanding, @leebyron 's feedback:

Perhaps `__type` should have always been nested as a `type` argument under `__schema`, so that similarly `__directive` should be nested as `directive` under `__schema`.

This general change might unlock general `__schema` arguments like:

```graphql
  __schema(includeDeprecated: true) { ... }
```

that would apply to all nested fields.

As this change is not a "must-have" it will probably be paused to make sure it fits into whatever introspection future is planned. We do not have to actually implement that entire new future to move forward with this small feature change, but we have to make sure that this smaller change aligns with the plan.

martinbonnin:
Rereading the [wg notes](https://github.com/graphql/graphql-wg/blob/main/notes/2024/2024-11.md#directive-meta-field-proposal-yaacovcr-5-min):

```
Right now it’s possible to have an inconsistent schema by passing different includeDeprecated values
``` 

Trying to list the cases where this may happen, the only place I found is if an input value has a default value that is a deprecated enum value that was not fetched. Are there others? 

This has the potential to become much more of an issue with [deprecated object types](https://github.com/graphql/graphql-spec/pull/997) but as it is now it's okay-ish?

All in all, I agree with @benjie's point that this is setting a precedent about accessing ancestors that might not be worth the tradeoff. 

Personnal anecdote: I have been tempted to do something similar to https://github.com/graphql/graphql-spec/issues/144 in the [Confetti API](https://confetti-app.dev/). The idea was to have a top-level `conference(conferenceId)` field that would filter out the content per-conference and that would effectively break normalized caching as Benjie described [in his ancestors article](https://benjie.dev/graphql/ancestors). In retrospect, I'm glad I didn't do it. But were this pattern used in more places, I might have bought into it without really realizing all the consequences...

Another question (that deviates quite a bunch from the original issue): do we really need `includeDeprecated`? Is saving a few bytes really worth the complexity here? Put it otherwise: could we deprecated `includeDeprecated` 🤪 ?

PR #1118, opened 2024-10-17
Clarify 'Values of Correct Type' rule relates to literals

benjie:
An [Input Value](https://spec.graphql.org/draft/#sec-Input-Values) is defined as being either a variable (if not const) or one of the literal types (IntValue, FloatValue, StringValue, BooleanValue, NullValue, EnumValue, ListValue or ObjectValue).

The rule ["Values of Correct Type"](https://spec.graphql.org/draft/#sec-Values-of-Correct-Type) states:

> - For each input Value `value` in the document:
>   - Let `type` be the type expected in the position `value` is found.
>   - `value` must be coercible to `type`.

However, an input value can be a variable, and variable coercion is handled at runtime (by [CoerceVariableValues](https://spec.graphql.org/draft/#sec-Coercing-Variable-Values)). Further, we already have a rule that validates that variables are only used in the positions in which they are allowed: [All Variable Usages Are Allowed](https://spec.graphql.org/draft/#sec-All-Variable-Usages-Are-Allowed).

It seems to me that "Values of Correct Type" only meant to handle _literal_ input values, so I've added the word "literal" for clarity. I've also expanded the explanation to reference where to look for variable input value validation. Since input coercion for Input Object references "runtime value", and validation doesn't have access to runtime values, I've made explicit the assumption that values represented by variables will be of the requisite type.

This is an alternative solution to, and

- fixes https://github.com/graphql/graphql-spec/pull/1113

Thank you to @yaacovCR for pointing out this deficiency :raised_hands:

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1319,8 +1319,9 @@ which may be a variable usage. The
 [All Variable Usages Are Allowed](#sec-All-Variable-Usages-Are-Allowed)
 validation rule ensures that each {variableUsage} is of a type allowed in its
 position. The [Coercing Variable Values](#sec-Coercing-Variable-Values)
-algorithm ensures runtime values for variables coerce correctly. Therefore we
-can assume the runtime value of each {variableUsage} is valid for usage in its
+algorithm ensures runtime values for variables coerce correctly. Therefore, for
+the purposes of the coercibility assertion in this validation rule, we can
+assume the runtime value of each {variableUsage} is valid for usage in its
 position.
 
 The type expected in a position includes the type defined by the argument a
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1306,22 +1306,22 @@ fragment resourceFragment on Resource {
 - For each literal Input Value {value} in the document:
   - Let {type} be the type expected in the position {value} is found.
   - {value} must be coercible to {type} (with the assumption that any
-    {variableUsage} nested within {value} will represent a runtime value
-    suitable for usage in its position).
+    {variableUsage} nested within {value} will represent a runtime value valid
+    for usage in its position).
 
 **Explanatory Text**
 
 Literal values must be compatible with the type expected in the position they
 are found as per the coercion rules defined in the Type System chapter.
 
-A {ListValue} or {ObjectValue} may nest additional Input Values, some of which
-may be a {variableUsage}. The
+Note: A {ListValue} or {ObjectValue} may contain nested Input Values, some of
+which may be a variable usage. The
 [All Variable Usages Are Allowed](#sec-All-Variable-Usages-Are-Allowed)
-validation rule asserts that each {variableUsage} is of a type allowed in that
-position, and the [Coercing Variable Values](#sec-Coercing-Variable-Values)
-algorithm will ensure runtime values of these variables coerce correctly, thus
-we can assume the runtime value of each {variableUsage} will be suitable for
-usage in its position.
+validation rule ensures that each {variableUsage} is of a type allowed in its
+position. The [Coercing Variable Values](#sec-Coercing-Variable-Values)
+algorithm ensures runtime values for variables coerce correctly. Therefore we
+can assume the runtime value of each {variableUsage} is valid for usage in its
+position.
 
 The type expected in a position includes the type defined by the argument a
 value is provided for, the type defined by an input object field a value is
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1306,20 +1306,22 @@ fragment resourceFragment on Resource {
 - For each literal Input Value {value} in the document:
   - Let {type} be the type expected in the position {value} is found.
   - {value} must be coercible to {type} (with the assumption that any
-    {variableUsage} nested within {value} will represent a runtime value of the
-    referenced variable's type).
+    {variableUsage} nested within {value} will represent a runtime value
+    suitable for usage in its position).
 
 **Explanatory Text**
 
 Literal values must be compatible with the type expected in the position they
-are found as per the coercion rules defined in the Type System chapter. Variable
-values are handled by the rule
-[All Variable Usages Are Allowed](#sec-All-Variable-Usages-Are-Allowed).
-{ListValue} and {ObjectValue} may nest additional Input Values, some of which
-may be a {variableUsage}. Each nested {variableUsage} will be coerced during
-execution - see [Coercing Variable Values](#sec-Coercing-Variable-Values) - thus
-we assume their runtime value will coerce to the type of the referenced
-variable.
+are found as per the coercion rules defined in the Type System chapter.
+
+A {ListValue} or {ObjectValue} may nest additional Input Values, some of which
+may be a {variableUsage}. The
+[All Variable Usages Are Allowed](#sec-All-Variable-Usages-Are-Allowed)
+validation rule asserts that each {variableUsage} is of a type allowed in that
+position, and the [Coercing Variable Values](#sec-Coercing-Variable-Values)
+algorithm will ensure runtime values of these variables coerce correctly, thus
+we can assume the runtime value of each {variableUsage} will be suitable for
+usage in its position.
 
 The type expected in a position includes the type defined by the argument a
 value is provided for, the type defined by an input object field a value is
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -1303,9 +1303,11 @@ fragment resourceFragment on Resource {
 
 **Formal Specification**
 
-- For each literal input Value {value} in the document:
+- For each literal Input Value {value} in the document:
   - Let {type} be the type expected in the position {value} is found.
-  - {value} must be coercible to {type}.
+  - {value} must be coercible to {type} (with the assumption that any
+    {variableUsage} nested within {value} will represent a runtime value of the
+    referenced variable's type).
 
 **Explanatory Text**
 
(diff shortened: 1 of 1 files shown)


benjie:
@graphql/tsc I would love to get your input on this change.

spec/Section 5 -- Validation.md
```diff
@@ -1303,14 +1303,23 @@ fragment resourceFragment on Resource {
 
 **Formal Specification**
 
-- For each input Value {value} in the document:
+- For each literal Input Value {value} in the document:
   - Let {type} be the type expected in the position {value} is found.
-  - {value} must be coercible to {type}.
+  - {value} must be coercible to {type} (with the assumption that any
+    {variableUsage} nested within {value} will represent a runtime value of the
```
yaacovCR, line 1309: There are two conditions mentioned below, that the variable usage is allowed and that the variable value will coerce at runtime, but only one assumption is mentioned here. Is that intentional?

 ```suggestion
    {variableUsage} nested within {value} is allowed and will represent a runtime value of the
```

Comment (benjie) [reply spec/Section 5 -- Validation.md:1309]: It is deliberate: coercion happens at runtime where variables are already resolved; for us to run coercion during validation we need to run a _slightly modified_ coercion - the modification being that it assumes (singular assumption) each variable reference it sees (something coercion would not normally understand) is equivalent to an arbitrary runtime value of that variable's type. It does not need to assume that the variable is allowed, from its point of view the variable exists whether it's allowed or not.

The _foundation_ of that singular assumption is the behaviour of "All Variable Usages Are Allowed" combined with "Coercing Variable Values" - these aren't assumptions, these are specified behaviours.

Comment (yaacovCR) [reply spec/Section 5 -- Validation.md:1309]: Let me try to work through it with 2 examples according to the way I understand your language. That might help you point out where I don't understand or whether there is an actual problem.

EXAMPLE A:

1. `{value}` := `{$varA}`
2. Expected type at the position of `{value}` := `String`
3. Type of `$varA` := `String`
4. Let `$varA` represents a runtime value of `String` such as `"SomeString"`
5. `{value}` := `"SomeString"`
6. `"SomeString"` is coercible to `String`
7. VALIDATION SUCCESS!

EXAMPLE B:

1. `{value}` := `{$varA}`
2. Expected type at the position of `{value}` := `Number`
3. Type of `$varA` := `String`
4. Let `$varA` represents a runtime value of `String` such as `"SomeString"`
5. `{value}` := `"SomeString"`
6. `"SomeString"` is NOT coercible to `Number`
7. VALIDATION FILURE!

In the implementation and in #1113, the idea would be that we would get VALIDATION SUCCESS in BOTH cases because the other rules/assumptions take care of variables.

Comment (benjie) [reply spec/Section 5 -- Validation.md:1309]: We’re talking about literal values here, so your examples should factor that in. I don’t know what your curly braces represent, I’m assuming not objects since dollar keys are not allowed. Example B fails validation according to both your example and the text of this PR, so I’m not sure what you mean by it passes validation? Validation doesn’t have access to runtime values so steps 4-6 are confusing - are you meaning “an arbitrary value of type String” rather than the runtime value, perhaps?

Comment (yaacovCR) [reply spec/Section 5 -- Validation.md:1309]: Let me try again, sticking with just the problematic Example B:

```graphql
query SomeQuery($someVariable: String) {
  someField(someArgOfNumberTypeLetsSayFloat: $someVariable)
}
```

According to the proposed new algorithm for this rule by this PR

> {value} must be coercible to {type} (with the assumption that any
>    {variableUsage} nested within {value} will represent a runtime value of the
>    referenced variable's type).

So for example, `$someVariable` must be coercible to `Float` with the assumption that the variable usage of `$someVariable` nested within $someVariable (it's the only thing) will represent a runtime value of type `String`. But that's not going to work, as `String` does not coerce to `Float`, so this operation would fail validation by this rule, as edited.

> Example B fails validation according to both your example and the text of this PR, so I’m not sure what you mean by it passes validation?

What I mean is that Example B and my hopefully more clear version above -- in my opinion -- SHOULD pass validation according to this rule. This rule should assume that Variables are in their Correct Position (because we have a separate rule for that) and that the variables coerce without error (because that will be and must be taken care of runtime). That's what the implementation at graphql-js assumes and what I was going for in #1113.

I've reopened #1113 as I think there is a difference in intent for the two PRs as to what `Values of Correct Type` should cover.

benjie: We've two rules and two algorithms we particularly care about here:

## [Values of Correct Type](https://spec.graphql.org/draft/#sec-Values-of-Correct-Type)

Needs clarification that this relates to literal values IMO.

Importantly: note that literal input object values can be supplied in positions expecting custom scalars, such as the `JSON` scalar, and those literal input object values may well have variable values inside of them.

## [All Variable Usages Are Allowed](https://spec.graphql.org/draft/#sec-All-Variable-Usages-Are-Allowed)

I've not read the relevant GraphQL JS source code, but I suspect that when it gets to step 2 in [IsVariableUsageAllowed](https://spec.graphql.org/draft/#IsVariableUsageAllowed()) it cannot find the expected type for a variable usage within a custom scalar (because there isn't one) and gives up without error.

The rule only relates to explicit schema positions, and "inside a custom scalar" isn't one, thus we cannot rely on All Variable Usages Are Allowed within custom scalars.

## [CoerceVariableValues()](https://spec.graphql.org/draft/#CoerceVariableValues())

Ensures that the variables at runtime coerce to the values they say. Doesn't care where they're used.

## [CoerceArgumentValues()](https://spec.graphql.org/draft/#CoerceArgumentValues())

Coerces the argument values "according to the input coercion rules of `argumentType`" - which is allowed to include consuming variables nested inside of the arguments (critical for lists and input objects, for example, but also important for our JSON custom scalar).

---

For lists and input objects you're definitely right that the type of these variables doesn't need to be checked by this rule - we could say:

> (with the assumption that any {variableUsage} nested within {value} will represent a runtime value of _the type expected in that position_)."

However, as established above, there _is_ no "type expected in that position" for variable references inside of custom scalars.

My thought was that by saying it's of the expected variable's type we'd get around this - but you're right that (for values like `[1, $string, 3]` versus an `[Int]` position) this now causes the error to effectively be raised twice - once in All Variable Usages Are Allowed and once in Values of Correct Type. Good catch!

I think we can just not worry about validating variable references inside custom scalars, leave that for runtime coercion inside `CoerceArgumentValues()`. I'm not aware of any custom scalars that accept input object literals and yet place variable type constraints on specific fields within them? I'm not sure if GraphQL.js even tries to do value coercion for custom JSON scalars when performing the Values of Correct Type check?

Evaluating for oneOf; we can assume that any variables have the expected type in that position, we just need to be clear that expected type in a oneof field position is a non-nullable variant of the field type... So yeah I think that covers it too. I'll tweak the wording.

yaacovCR (approved):

spec/Section 5 -- Validation.md
```diff
@@ -1304,15 +1304,26 @@ fragment resourceFragment on Resource {
 
 **Formal Specification**
 
-- For each input Value {value} in the document:
+- For each literal Input Value {value} in the document:
```
martinbonnin, line 1307: Suggestion to makes the locations explicit:

```suggestion
- For each argument Value and variable DefaultValue {value} in the document:
```

I think this covers it all? 

It makes it easier for implementers to make sure they cover everything and removes ambiguity about list/object literals (they are a `Value` that may contain nested `Value`, we only need to validate the root value)

The algorithm could rule out variables as the first step:

```
  - If {value} is not a variable
    - ...
```

spec/Section 5 -- Validation.md
```diff
@@ -1304,15 +1304,26 @@ fragment resourceFragment on Resource {
...
   - Let {type} be the type expected in the position {value} is found.
-  - {value} must be coercible to {type}.
+  - {value} must be coercible to {type} (with the assumption that any
+    {variableUsage} nested within {value} will represent a runtime value valid
+    for usage in its position).
```
martinbonnin, line 1311: What about custom scalars? Should they also be skipped as their coercing is implementation dependant? 

It's not an issue for a server but GraphiQL for an example cannot really enforce this without knowledge about the implementation that is not available in introspection.

martinbonnin (approved):
Good clarification 👍 

Left a few additional questions that arose while doing the review but the PR in itself is an improvement from the current state and I support merging it.

spec/Section 5 -- Validation.md
benjie: If we're trying to cover everything should we indicate that the arguments don't just include field arguments, but also directive arguments? And in future, we'd also be including fragment arguments (which I can't remember if we decided to call something other than "arguments" in the end?).

```suggestion
For each literal field argument value, literal directive argument value, literal
fragment argument value and variable default value {value} in the document:
```

I think this is getting a bit of a mouthful though, and puts more burden on spec updates. Let's stick with "input value", but maybe define it in a follow-up PR?

spec/Section 5 -- Validation.md
benjie: To GraphiQL, all custom scalar values should be "valid" since it can include them in the request it sends to the server. The server should validate them.

martinbonnin: GraphiQL cannot really say whether `"InvalidDate"` is coercible to a `Date` in literal positions.

For an example, with this operation:

```
{
  events(after: "InvalidDate") {
    title
    startsAt
  }
}
```

If we read the spec strictly, this should trigger a validation error. If GraphiQL doesn't highlight it, then there's a discrepency. 

From an Apollo Kotlin point of view, this is also an issue because validating those literal positions would require the users to pass the coercing to the compiler, which is extra complexity. Doable, for sure, but not trivial.

spec/Section 5 -- Validation.md
martinbonnin: If we read `argument` in the "grammar" acceptation, the nice thing with `argument` is that it works for both fields and directive, they're using the same grammar rule. It'd actually be cool to have "AST coordinates" that we could refer to to easily identify those locations.

Agree that listing everything exhaustively is taking the risk this content gets up to date 👍

martinbonnin: In all cases, LGTM, resolving that conversation.

spec/Section 5 -- Validation.md
martinbonnin: Follow up PR: https://github.com/graphql/graphql-spec/pull/1157

BoD (approved):

mjmahone (approved):

michaelstaib (approved):

PR #1123, opened 2024-10-31
Change 'original' to 'previous' to clarify multiple extensions

benjie:
@benhead described this well in https://github.com/graphql/graphql-spec/discussions/1116

Essentially the issue boils down to:

```
1  union MyUnion = Foo
2  extend union MyUnion = Bar | Baz
3  extend union MyUnion = Bar | Qux
```

- On line 1 we define a union.
- On line 2 we extend the "original" union (defined on line 1).
  - The rule "All member types of a Union type extension must not already be a member of the original Union type." passes, since neither `Bar` nor `Baz` exist in the original union (`Foo`)
- On line 3 we extend [the "original" union](https://spec.graphql.org/draft/#sel-GAHdhBFABABvDt4V). AMBIGUITY: is this "original" the true original, defined on line 1, or is it the previous version of the union which is the combination of line 1 and line 2?
  - If we go with the "true original" interpretation, this type extension passes the rule since neither `Bar` nor `Qux` are referenced in the union on line 1.
  - If we go with the "previous" interpretation, this type extensions fails the rule (as we would expect?) because `Bar` is already included in the composite union `Foo | Bar | Baz`.

I believe we actually mean the previous (i.e. combined) version. This is certainly what the reference implementation seems to do:

```ts
import { buildSchema, printSchema, validateSchema } from 'graphql';

const schema = buildSchema(/* GraphQL */ `
  union MyUnion = Foo
  extend union MyUnion = Bar | Baz
  extend union MyUnion = Bar | Qux

  type Query { u: MyUnion }
  type Foo { foo: Int }
  type Bar { bar: Int }
  type Baz { baz: Int }
  type Qux { qux: Int }
`);

if (errors.length > 0) {
  console.dir(errors);
  process.exit(1);
}

console.dir(printSchema(schema));
```

yields:

```
GraphQLError: Union type MyUnion can only include type Bar once.
```

> _Aside: I actually think the original intention was to extend a type once in a derived schema as implied by [the Type System Extensions opening paragraph](https://spec.graphql.org/draft/#sel-FAHPJABAB3J_2T); however that ship has long sailed and multiple extensions are common - in fact extending types defined in the same schema (as shown in the document below) is common practice._

I attempted to add a non-normative note to clarify that "original here means the previous version, which may itself be an extended type..." or something but couldn't get happy with the wording. Then I realised that simply changing `original` to `previous` would convey the intent well, and by sheer coincidence they happen to have the same number of characters!

There is one caveat to this solution, and that is that original type definitions are "hoisted", so though the order of type extensions in the document is significant, the original type (_actual_ original!) can occur anywhere, so the use of "previous" is arguably dubious in this case (though I'd argue that named type definitions always precede type extensions in order of processing, so this is still fine):

```graphql
extend union MyUnion = Bar | Baz
extend union MyUnion = Qux

type Query { u: MyUnion }
type Foo { foo: Int }
type Bar { bar: Int }
type Baz { baz: Int }
type Qux { qux: Int }

union MyUnion = Foo
```

_Also, I can no longer read the word `union` without seeing it as `onion`, so that's a fun [malapropism](https://en.wikipedia.org/wiki/Malapropism) that's going to haunt me now_ :roll_eyes:

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -42,7 +42,7 @@ TypeSystemExtension :
 - TypeExtension
 
 Type system extensions are used to represent a GraphQL type system which has
-been extended from some original type system. For example, this might be used by
+been extended from some previous type system. For example, this might be used by
 a local service to represent data a GraphQL client only accesses locally, or by
 a GraphQL service which is itself an extension of another GraphQL service.
 
@@ -266,8 +266,8 @@ SchemaExtension :
 - extend schema Directives[Const]? { RootOperationTypeDefinition+ }
 - extend schema Directives[Const] [lookahead != `{`]
 
-Schema extensions are used to represent a schema which has been extended from an
-original schema. For example, this might be used by a GraphQL service which adds
+Schema extensions are used to represent a schema which has been extended from a
+previous schema. For example, this might be used by a GraphQL service which adds
 additional operation types, or additional directives to an existing schema.
 
 Note: Schema extensions without additional operation type definitions must not
@@ -279,7 +279,7 @@ The same limitation applies to the type definitions and extensions below.
 Schema extensions have the potential to be invalid if incorrectly defined.
 
 1. The Schema must already be defined.
-2. Any non-repeatable directives provided must not already apply to the original
+2. Any non-repeatable directives provided must not already apply to the previous
    Schema.
 
 ## Types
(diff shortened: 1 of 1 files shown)


mjmahone (approved):

robrichard (approved):

benjie:
I'll leave this open for a couple weeks for any further feedback, then merge it if there isn't any pushback.

benjie:
There's not been any pushback, so as agreed at the last WG I will now merge this :+1:

PR #1124, opened 2024-11-01
Spec edits for incremental delivery, Section 3 & 7 only

robrichard:


spec/Section 3 -- Type System.md
```diff
@@ -1942,6 +1943,11 @@ by a validator, executor, or client tool such as a code generator.
 
 GraphQL implementations should provide the `@skip` and `@include` directives.
 
+GraphQL implementations are not required to implement the `@defer` and `@stream`
+directives. If either or both of these directives are implemented, they must be
+implemented according to this specification. GraphQL implementations that do not
+support these directives must not make them available via introspection.
+
```
Keweiqu, line 1950: Nits: let's move this after the @deprecated section because these are not required. Also should we mention that implementations not implementing @defer and @stream when facing a request with @defer and @stream will ignore these directives? (this might be specified somewhere else already as a general rule for unknown directive)

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2168,99 @@ to the relevant IETF specification.
...
+    uri
+  }
+}
+```
+
+#### @defer Arguments
+
+- `if: Boolean! = true` - When `true`, fragment _should_ be deferred (see
+  related note below). When `false`, fragment will not be deferred and data will
+  be included in the initial response. Defaults to `true` when omitted.
```
Keweiqu, line 2208: nits: let's not mention "initial response" because for the case of @defer nested inside @stream. It's not always guaranteed to be in response #0.

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2168,99 @@ to the relevant IETF specification.
...
+  label: String
+  if: Boolean! = true
+  initialCount: Int = 0
+) on FIELD
+```
+
+The `@stream` directive may be provided for a field of `List` type so that the
+backend can leverage technology such as asynchronous iterators to provide a
+partial list in the initial response, and additional list items in subsequent
+responses. `@include` and `@skip` take precedence over `@stream`.
```
Keweiqu, line 2229: do you want to add a comment about validation error if this directive is applied on a non-list type?

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2168,99 @@ to the relevant IETF specification.
...
+  responses and associate it with the corresponding stream directive. If
+  provided, the GraphQL service must add it to the corresponding pending object
+  in the response. `label` must be unique label across all `@defer` and
+  `@stream` directives in a document. `label` must not be provided as a
+  variable.
+- `initialCount: Int` - The number of list items the service should return as
+  part of the initial response. If omitted, defaults to `0`. A field error will
+  be raised if the value of this argument is less than `0`.
+
+Note: The ability to defer and/or stream parts of a response can have a
```
Keweiqu, line 2256: Love this section!!!

spec/Section 7 -- Response.md
```diff
@@ -22,14 +27,31 @@ key `data`. The value of this entry is described in the "Data" section. If the
 request failed before execution, due to a syntax error, missing information, or
 validation error, this entry must not be present.
 
+When the response of the GraphQL operation contains incrementally delivered
+results, both the initial payload and all subsequent payloads must contain an
+entry with key `hasNext`. The value of this entry must be {true} for all but the
+last response in the stream. The value of this entry must be {false} for the
+last response of the stream. This entry must not be present for GraphQL
+operations that return a single response map.
```
Keweiqu, line 35: "This entry must not be present for GraphQL operations that return a single response map"

Why? Seems like an unnecessary restriction.

spec/Section 7 -- Response.md
```diff
@@ -244,6 +264,366 @@ discouraged.
...
+`id`.
+
+Every Pending Result must contain an entry with the key `path`. When the Pending
+Result is associated with a `@stream` directive, it indicates the response list
+field that is not known to be complete. Clients should expect the GraphQL
+Service to incrementally deliver the remainder of indicated list field. When the
+Pending Result is associated with a `@defer` directive, it indicates that the
+response fields contained in the deferred fragment are not known to be complete.
+Clients should expect the the GraphQL Service to incrementally deliver the
+remainder of the fields contained in the deferred fragment.
```
Keweiqu, line 309: there is some ambiguity with this statement. When I read this it seems like with a fragment X {a, b, c} which is deferred, the GraphQL service can incrementally deliver {a, b} and then deliver {c}

spec/Section 7 -- Response.md
```diff
@@ -244,6 +264,366 @@ discouraged.
...
+a prior Pending Result. This Pending Result must be associated with a `@defer`
+directive.
+
+The Incremental Object Result's `path` can be determined using the prior Pending
+Result with the same `id` as this Incremental Result. The Incremental Object
+Result may contain a `subPath` entry. If the `subPath` entry is present, The
+Incremental Object Record's path can be determined by concatenating the Pending
+Result's `path` with this `subPath`. If no `subPath` entry is present, the path
+is the same as the Pending Result's `path`.
+
```
Keweiqu, line 386: what about index? Do you want me to raise this as a separate PR?

spec/Section 7 -- Response.md
```diff
@@ -244,6 +264,366 @@ discouraged.
...
+Payload 3, contains the remaining data from the `NameAndHomeWorldFragment`.
+`lastName` is the only remaining field that has not been delivered in a previous
+payload.
+
+```json example
+{
+  "incremental": [
+    {
+      "id": "1",
+      "data": { "lastName": "Skywalker" }]
```
Keweiqu, line 619: typo^

Keweiqu (commented):
Thank you so much @robrichard  for this amazingly written edits! What a great progress!! See my inline comments. 

And let me know if you want me to raise the error handling as a separate PR or a separate topic in WG discussion.

spec/Section 7 -- Response.md
robrichard: We discussed this some time ago and landed on this: https://github.com/graphql/defer-stream-wg/discussions/57#discussioncomment-3824861

I think the main reason was a fear that clients may start to depend on `hasNext: false` being present in a single payload response, and this would break if they switch to a server that does not support defer/stream

spec/Section 7 -- Response.md
robrichard: I agree that it is ambiguous. I wrote this way because I was thinking that your example is technically possible. If `a` and `b` are in both a deferred and non-deferred fragment, a & b will be in the initial response and only `c` will be incrementally delivered. Or if `a` and `b` are in both deferred fragment X and deferred fragment Y, a & b might be delivered incrementally when fragment Y is complete, while `c` will be sent in a later payload.

Do you have a suggestion to clarify?

spec/Section 7 -- Response.md
robrichard: When we switched to the latest response format we decided not to put `index` in the incremental object. It's not strictly necessary since list items must be delivered in order. We were thinking it could be added in a future spec update if we want to support out of order or sparse responses. If this is an issue for you can you start a new discussion topic in https://github.com/graphql/defer-stream-wg/discussions?

spec/Section 3 -- Type System.md
robrichard: Moved and added a sentence with a link to the `Directives Are Defined` validation rule which will fail if `defer` or `stream` is used on a service that does not implement them.

spec/Section 3 -- Type System.md
robrichard: Updated!

spec/Section 3 -- Type System.md
robrichard: Added a sentence mentioning the "Stream Directives Are Used On List Fields" rule. It's a broken link for now, but we have this rule defined in the [full PR](https://github.com/graphql/graphql-spec/pull/1110)

robrichard:
@Keweiqu thanks for your review! I have addressed your comments and also moved the examples into a new appendix as discussed during the working group meeting.

spec/Section 7 -- Response.md
Keweiqu: Ohhhhh yeah now I remember!

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2163,109 @@ to the relevant IETF specification.
 ```graphql example
 scalar UUID @specifiedBy(url: "https://tools.ietf.org/html/rfc4122")
 ```
+
+### @defer
+
+GraphQL implementations are not required to implement the `@defer` and `@stream`
+directives. If either or both of these directives are implemented, they must be
+implemented according to this specification. GraphQL implementations that do not
+support these directives must not make them available via introspection. The
```
martinbonnin, line 2172: Can we clarify what is meant by implementation here? My current understanding is that an "implementation" is typically a library, such as `graphql-js` or `graphql-java` while a "service" is typically a server or a collection of servers, such as the GitHub or Shopify APIs.

If that's the case, the "not required" part could be confusing. If I'm writing the `graphql-cobol` library, can I decide to leave `@defer` out and still claim compatibility with the spec?

robrichard: @martinbonnin thanks for reviewing. I updated this to say `services` instead of `implementations` as I was referring to servers here

robrichard: I'm not sure this applies to `@defer` and `@stream`, because they are directives that apply to ["executable locations"](https://spec.graphql.org/October2021/#ExecutableDirectiveLocation), whereas `@specifiedBy` and `@oneOf` only apply to ["type system location"](https://spec.graphql.org/October2021/#TypeSystemDirectiveLocation).

I think a GraphQL server could not support the "type system definition language" but still support `@defer` and `@stream`. Clients could determine if `@defer` and `@stream` are supported by using the introspection query:

```graphql
query {
  __schema {
    directives {
      name
    }
  }
}
```

I think the inverse could also be true. A GraphQL Server that supports the type system definition language may have no desire to support `@defer`/`@stream` and that's ok since we want this feature to be strictly optional.

It's possible I'm not thinking about this correctly, let me know what you think.

martinbonnin: Apologies in advance, a bunch of nitpicking on the vocabulary/glossary is coming below. 

> a GraphQL server could not support the "type system definition language" but still support `@defer`

I don't think it makes sense for a GraphQL server to support the "type system definition language". A server reads request in and writes responses out. SDL is an implementation detail for the server. 

But maybe not for its implementation (==lib?). A lib takes SDL in and generates a compliant service out of it. I always thought of `implementation` in the spec as being the "lib" but re-reading it, I'm not sure any more.

Regardless, if we use this service vs implementation distinction, I would rephrase like so:

```
A GraphQL implementation that does not support the "type system definition language"  can
still support @defer and @stream

Clients can determine if @defer and @stream are supported by using the introspection 
query against a service.

A GraphQL service may have no desire to support @defer/@stream and that's ok since we 
want this feature to be strictly optional. 
```

I think this is the same for `@oneOf` and `@specifiedBy`:

```
A GraphQL implementation that does not support the "type system definition language"  can
still support @oneOf and @specifiedBy

Clients can determine if @oneOf and @specifiedBy are supported by using the introspection 
query against a service.

A GraphQL service may have no desire to support @oneOf and @specifiedBy and that's ok since we 
want this feature to be strictly optional. 
```

That's why I'm tempted to put them in the same bucket of features even if `@defer` is obviously quite important for clients to be aware of.

robrichard: @martinbonnin I am definitely open to changing this, just want to talk it through to make sure I fully understand.

Is this statement true for `@oneOf` and `@specifiedBy`?

> A GraphQL implementation that does not support the "type system definition language"  can
still support `@oneOf` and `@specifiedBy`

Since `@oneOf` and `@specifiedBy` are directives that can only be applied to grammar that is part of the "type system definition language", how can an implementation support this directive without supporting the grammar that is required to place the directive? I agree that `specifiedByURL` introspection field may be supported but this is not the same as the directive.

The same is not true for `@defer` and `@stream` because these directives are applied to the execution grammar.

> A GraphQL service may have no desire to support `@oneOf` and `@specifiedBy` and that's ok since we 
want this feature to be strictly optional. 

Again, I could be wrong about this but my understand for `@specifiedBy` is that ideally all GraphQL servers would support it, but that's not possible for servers written before they were added to the spec, so RFC 2119 `should` is used. Additionally, we further restrict it to "implementations that support the type system definition language" since `@specifiedBy` can only be placed in the type system grammar. 

For `@defer` and `@stream` we don't want to use RFC 2119 `should`/`recommended` language since it is optional and we expect some servers will choose not to implement it for reasons other than being behind on the latest spec changes.

martinbonnin: > my understand for @specifiedBy is that ideally all GraphQL servers would support it, but that's not possible for servers written before they were added to the spec, so RFC 2119 should is used.
> [...]
> For @defer and @stream we don't want to use RFC 2119 should/recommended language since it is optional and we expect some servers will choose not to implement it for reasons other than being behind on the latest spec changes.

Gotcha, thanks! So we've got:

1. RFC 2119 `must` for stuff that all services must support (`@include` for an example)
2. RFC 2119 `should` to allow older services to still be compatible but newer services should support (`@specifiedBy` for an example)
3. explicit language like here for stuff that is completely optional (`@defer` for an example)

Did I get that right? 

If yes, I was missing the 2. vs 3. distinction, thanks for pointing that out 🙏  

I'm still not 100% sold on it though. I kind of like that other directives all use the same sentence and feels a bit weird that `@defer` is not following the pattern (or maybe it's just my OCD...). 

What about using RFC 2119 `may` to denote that it's "more" optional than `@specifiedBy`?  Was that ever mentioned?

```
:: A _built-in directive_ is any directive defined within this specification.

GraphQL implementations should provide the `@skip` and `@include` directives.

GraphQL implementations that support the type system definition language must
provide the `@deprecated` directive if representing deprecated portions of the
schema.

GraphQL implementations that support the type system definition language should
provide the `@specifiedBy` directive if representing custom scalar definitions.

GraphQL implementations that support the type system definition language may
provide the `@defer` directive if they support returning response streams.

GraphQL implementations that support the type system definition language may
provide the `@stream` directive if they support incrementally delivered results.
```

I don't want to bikeshed this too much especially if it was discussed already. Is there any reading material and/or pointers I could catch up on? 

> A GraphQL implementation that does not support the "type system definition language" can
still support @oneOf and @specifiedBy
>> Since @oneOf and @specifiedBy are directives that can only be applied to grammar that is part of the "type system definition language", how can an implementation support this directive without supporting the grammar that is required to place the directive?

It can't but it's fine? 

My reading of `A GraphQL implementation that does not support the "type system definition language"` is a code-first GraphQL implementation like `graphql-kotlin` for an example. It has nothing to do with supporting the grammar or anything like this. A code-first implementation can implement `@specifiedBy`.

Supporting the "type system definition language" is a feature of some GraphQL libs that has nothing to do with how the grammar is internally implemented. 

I read all of these sentences as: "if a lib reads a SDL written by a user, then the lib must/should/may provide the built in directive definitions":

```
user SDL + built-in directives == final schema
```

> The same is not true for @defer and @stream because these directives are applied to the execution grammar.

The way I see it the directive location (executable or type system) is irrelevant. These sentences are about the directive definitions that must/should/may be added to the resulting schema.

But again this is my very personal understanding of it and maybe PTSD from the [full schemas](https://github.com/graphql/graphql-wg/blob/main/rfcs/FullSchemas.md) discussion. Thanks for diving into this with me!

martinbonnin: > I still don't think we should include that support the type system definition language for anything related to @defer or @stream because they are features that are unrelated to the SDL.

I see it now! Thanks for explaining 🙏

> My interpretation is that the spec is not providing any sort of guidance for the @specifiedBy directive for this type of implementation since it is excluded by the phrase GraphQL implementations that support the type system definition language ...

Agreed here as well 👍 But I also think this may be a problem? 

Do we have somewhere in the spec where we say what built-in directives are required in a valid full/final schema and which ones are not? I initially thought this was the place but since it excludes code-first implementations it can't reasonnably be it?

> But for @defer and @stream it is important that this type of implementation follow this guidance regardless of SDL support so we should not include the same phrase.

Right 👍 I actually like the sentence as it is now because it applies to services. I think using `service` here is correct since SDL support is irrelevant for a service.

Finally, I'd say there should be a similar sentence for other directives? `@include`/`@skip` come to mind for an example because they are used in executable locations. Can a service ignore them? But I would probably do it for all directives for symmetry and consistency reasons.

robrichard: > `@include`/`@skip` come to mind for an example because they are used in executable locations.

For `@include`/`@skip`, the spec does have the following sentence at the beginning of this section 


```
GraphQL implementations should provide the `@skip` and `@include` directives.
```

> Do we have somewhere in the spec where we say what built-in directives are required in a valid full/final schema and which ones are not? I initially thought this was the place but since it excludes code-first implementations it can't reasonnably be it?

I do think this is the correct place, and with the above sentence for `@skip` and `@include` we are covered for all SDL and non-SDL directives. But even `@skip` and `@include` are not required since the spec is using RFC 2119 `should`.

> My interpretation is that the spec is not providing any sort of guidance for the `@specifiedBy` directive for this type of implementation since it is excluded by the phrase `GraphQL implementations that support the type system definition language ...`

> Agreed here as well 👍 But I also think this may be a problem?

Perhaps! Maybe it's worth a WG discussion to clarify?

spec/Section 7 -- Response.md
benjie: I think the wording isn't quite right though. I think if the request uses `@defer`/`@stream` then it's okay to have `hasNext: false` but if the request did not use `@defer` or `@stream` then the result must not include `hasNext`. Does that differ from what we discussed before?

spec/Section 7 -- Response.md
```diff
@@ -244,6 +264,174 @@ discouraged.
...
+result. This field should be a list of path segments starting at the root of the
+response and ending with the field to be associated with. Path segments that
+represent fields should be strings, and path segments that represent list
+indices should be 0-indexed integers. If the path is associated to an aliased
+field, the path should use the aliased name, since it represents a path in the
+response, not in the request.
+
+When the `path` field is present on an "Error result", it indicates the response
+field which experienced the error.
+
```
benjie, line 279: Should we extract this and line 136/137 into a separate PR to merge sooner?

spec/Section 7 -- Response.md
```diff
@@ -244,6 +264,174 @@ discouraged.
...
+the response field at this path.
+
+If any field errors were raised during the execution of the results in `items`
+and these errors bubbled to a path higher than the Incremental List Result's
+path, The Incremental List Result is considered failed and should not be
+included in the response stream. The errors that caused this failure will be
+included in a Completed Result.
+
+If any field errors were raised during the execution of the results in `items`
+and these errors did not bubble to a path higher than the Incremental List
```
benjie, line 369: We don't use the word "bubble" in the spec, please use "propagate"

benjie:
One comment on this is that it quite often talks about multiple "responses", but I don't think that's really accurate. Subscriptions are the closest match, and they have a single response which is a "response stream" which has _yields "events". I don't think "events" works for this, so I've put "payloads" in my edits, but I think the terminology we use here should be explored.

---

I noticed in section 7 you've referred to "result" as well. I think a "response" consisting of multiple "payloads" of which each is a "result" makes sense, but I wonder if we should eliminate "payloads" and just talk about a "result" (itself being a payload) throughout?

spec/Section 3 -- Type System.md
```diff
@@ -794,8 +794,9 @@ And will yield the subset of each object type queried:
 When querying an Object, the resulting mapping of fields are conceptually
 ordered in the same order in which they were encountered during execution,
 excluding fragments for which the type does not apply and fields or fragments
-that are skipped via `@skip` or `@include` directives. This ordering is
-correctly produced when using the {CollectFields()} algorithm.
+that are skipped via `@skip` or `@include` directives or temporarily skipped via
+`@defer`. This ordering is correctly produced when using the {CollectFields()}
+algorithm.
```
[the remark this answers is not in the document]
yaacovCR: Small tweak: in this context, I think only defer is relevant as a cause for fields being postponed. (Stream may cause subfields to be postponed, but there would always be an empty list for the initial field.)

martinbonnin:
> One comment on this is that it quite often talks about multiple "responses", but I don't think that's really accurate. 

Great point about terminology. I'm taking this opportunity to promote the [Glossary](https://github.com/graphql/graphql-wg/blob/main/rfcs/glossary/Appendix%20C%20--%20Glossary.md). Whatever we decide we should add there.

> Subscriptions are the closest match, and they have a single response which is a "response stream" which has _yields "events"

The spec can allow both interpretations I think?

```
A GraphQL service generates a response from a request via execution.
```

But then later on, in the Subscribe algorithm:

```
Return responseStream.
```

But then in the response format:

```
A response to a GraphQL request must be a map.
```

Can a response stream be a map? 🤔 Oh well...

benjie:
> Can a response stream be a map? 🤔 Oh well...

Yeah I found this same issue with subscriptions too; I think the entire response section need to be adapted to work with subscriptions. Kind of weird this hasn't come up before TBH.

benjie:
@martinbonnin I've edited quite heavily [this comment above](https://github.com/graphql/graphql-spec/pull/1124#issuecomment-2506277125) - WDYT?

martinbonnin:
@benjie I agree that response is mostly associated with `{data, errors}` today and we should keep this as much as possible. Looks like we should extend it to account for `{data, errors, pending, hasNext}` with this proposal.

Re: `payload`, it's used pervasively in the Susbscription world where it's used either as generic data to be interpreted according to type (see [graphql-ws](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md)) or as execution result (==response) (see [graphql-sse](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md)) so agree we should keep `payload` as something holding data without mandating that this data has a specific shape.

I like `incremental result` but that introduces "result" in the mix. What's a non-incremental "result"? a response? If yes, then why not going with `incremental response`? Because an `incremental` response is not a response...etc.. naming is hard!

spec/Section 3 -- Type System.md
```diff
@@ -1946,6 +1947,14 @@ GraphQL implementations that support the type system definition language must
 provide the `@deprecated` directive if representing deprecated portions of the
 schema.
 
+GraphQL services are not required to implement the `@defer` and `@stream`
+directives. If either or both of these directives are implemented, they must be
+implemented according to this specification. GraphQL services that do not
+support these directives must not make them available via introspection. The
+[Directives Are Defined](#sec-Directives-Are-Defined) validation rule will
+prevent GraphQL Operations containing the `@defer` or `@stream` directive from
+being executed by a GraphQL service that does not implement these directives.
```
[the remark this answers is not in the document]
robrichard: Updated

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2171,101 @@ to the relevant IETF specification.
...
+  if: Boolean! = true
+) on FRAGMENT_SPREAD | INLINE_FRAGMENT
+```
+
+The `@defer` directive may be provided for fragment spreads and inline fragments
+to inform the executor to delay the execution of the current fragment to
+indicate deprioritization of the current fragment. A query with `@defer`
+directive will cause the request to potentially return multiple responses, where
+deferred data is delivered in subsequent responses. `@include` and `@skip` take
+precedence over `@defer`.
```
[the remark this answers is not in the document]
robrichard: Updated

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2171,101 @@ to the relevant IETF specification.
...
+
+- `if: Boolean! = true` - When `true`, fragment _should_ be deferred (see
+  related note below). When `false`, fragment will not be deferred and data will
+  be included in the initial response. Defaults to `true` when omitted.
+- `label: String` - May be used by GraphQL clients to identify the data from
+  responses and associate it with the corresponding defer directive. If
+  provided, the GraphQL service must add it to the corresponding pending object
+  in the response. `label` must be unique label across all `@defer` and
+  `@stream` directives in a document. `label` must not be provided as a
+  variable.
```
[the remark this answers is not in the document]
robrichard: I also updated the same section under `@stream`

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2171,101 @@ to the relevant IETF specification.
...
+) on FIELD
+```
+
+The `@stream` directive may be provided for a field of `List` type so that the
+backend can leverage technology such as asynchronous iterators to provide a
+partial list in the initial response, and additional list items in subsequent
+responses. `@include` and `@skip` take precedence over `@stream`. The
+[Stream Directives Are Used On List Fields](#sec-Stream-Directives-Are-Used-On-List-Fields)
+validation rule is used to prevent the `@stream` directive from being applied to
+a field that is not a `List` type.
```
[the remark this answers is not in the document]
robrichard: Updated with minor edits to remove references to the initial payload, see https://github.com/graphql/graphql-spec/pull/1124#discussion_r1833244721

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2171,101 @@ to the relevant IETF specification.
...
+
+```graphql example
+query myQuery($shouldStream: Boolean) {
+  user {
+    friends(first: 10) {
+      nodes @stream(label: "friendsStream", initialCount: 5, if: $shouldStream)
```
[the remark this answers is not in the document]
robrichard: I added a selection set. Most examples in this section don't have a schema. I can add one if you think it adds clarity

spec/Section 3 -- Type System.md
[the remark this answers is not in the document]
robrichard: I updated `temporarily skipped` to `postponed`, but left the rest as is, since as @yaacovCR mentioned only `@defer` will cause fields to be postponed.

spec/Section 7 -- Response.md
robrichard: Updated all references from "bubble" to "propagate"

spec/Section 7 -- Response.md
robrichard: Opened https://github.com/graphql/graphql-spec/pull/1129

spec/Section 7 -- Response.md
robrichard: My intention was that servers should never include `hasNext: false` in a single payload response, even if a request uses `@defer`/`@stream` but the server did not defer anything. I agree that in this case we don't have the same risk of allowing it in requests without `@defer`/`@stream`, but it seems simpler to me to not make this distinction.

robrichard:
@benjie @martinbonnin Regarding naming, I went back and forth on this many times and am still not totally confident the names I've chosen are best.

I've been using `Response` to refer to the entire stream, and `payload` to refer to each object that is sent in the stream. I think the distinction between subscriptions is important, so a query/mutation with defer/stream has one response with multiple payloads, while subscriptions have multiple responses.

I'm also using `result` to refer to the objects inside the `pending`, `incremental`, and `completed` arrays, i.e. `Pending Result`, `Incremental Result`, `Completed Result`. However, I noticed the spec seems to use `Response` and `Result` interchangeably: 
> The `data` entry in the response will be the result of the execution of the
requested operation .... 
> If an error was raised before execution begins, the `data` entry should not be
present in the result.

So maybe these should be renamed to `Pending Object`, `Incremental Object`, `Completed Object` or something else?

spec/Section 3 -- Type System.md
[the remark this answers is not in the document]
martinbonnin: TBH, I'm still not super happy about the "provide" wording. It seems it was introduced for SDL capable implementations but we're extending it to any kind of implementation here. The word "implementation" itself is still not 100% clear IMO.

But I don't want to bikeshed this to death. This is indeed much clearer can definitely be addressed outside the `@defer`/`@stream` context.

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2171,101 @@ to the relevant IETF specification.
...
+potentially significant impact on application performance. Developers generally
+need clear, predictable control over their application's performance. It is
+highly recommended that GraphQL services honor the `@defer` and `@stream`
+directives on each execution. However, the specification allows advanced use
+cases where the service can determine that it is more performant to not defer
+and/or stream. Therefore, GraphQL clients _must_ be able to process a response
+that ignores the `@defer` and/or `@stream` directives. This also applies to the
+`initialCount` argument on the `@stream` directive. Clients _must_ be able to
+process a streamed response that contains a different number of initial list
+items than what was specified in the `initialCount` argument.
```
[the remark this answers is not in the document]
robrichard: Happy to revisit this discussion. If I am remembering correctly, the arguments for this version were:
* We are still "highly recommending" servers send the correct number of list items earlier in this paragraph
* This allows the most flexibility, clients shouldn't make any assumptions about the length of the initial list
* From a type perspective clients should already not be making assumptions about the length of the list

benjie:
>  However, I noticed the spec seems to use Response and Result interchangeably:

I think the spec uses "result" very generically - it can be the result of anything. The GraphQL response is the result of a GraphQL request, but there's also results from fields, results from coercion, etc etc. I've raised #1130 to clarify the example you raised.

> I've been using Response to refer to the entire stream

Currently I think the spec fairly universally uses the term "response" to mean the "data and/or errors" object that you get either for a query/mutation, or as the result of each individual event in a subscription.

> I think the distinction between subscriptions is important, so a query/mutation with defer/stream has one response with multiple payloads, while subscriptions have multiple responses.

This is an important distinction and it helps to clarify that the intent of incremental/stream is effectively a "snapshot in time" rather than being another way of doing subscriptions.

Maybe instead of "response" being one thing, we should think of what we refer to as "response" currently actually being the standard response payload. Incremental delivery would then return an incremental response stream consisting of an initial response payload followed by one or more subsequent response payloads. We could trim the word "payload" from each of these. We would end up with:

- query / mutation (without incremental) - yields a single "response"
- subscription - yields a response stream, a stream of "response"s

A "response" is either:

- a "standard response"
  - data and/or errors
- or an "incremental stream"
  - a stream consisting of:
    - an "initial response"
      - (data and/or errors, pending, hasNext:true)
    - followed by one or more "subsequent response" (or maybe "continuation response"?)
      - (incremental and/or errors, pending?, hasNext)

This also factors in that subscriptions could, in future, also support stream/defer - their "response" is itself an "incremental stream".

> I'm also using result to refer to the objects inside the pending, incremental, and completed arrays, i.e. Pending Result, Incremental Result, Completed Result.

This seems fine for now.

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2170,108 @@ to the relevant IETF specification.
...
+  profile_picture {
+    uri
+  }
+}
+```
+
+#### @defer Arguments
+
+- `if: Boolean! = true` - When `true`, fragment _should_ be deferred (see
+  related note below). When `false`, fragment will not be deferred. Defaults to
```
BoD, line 2209: Should it be `must not` be deferred?

spec/Section 3 -- Type System.md
```diff
@@ -2162,3 +2170,108 @@ to the relevant IETF specification.
...
+
+```graphql
+directive @stream(
+  label: String
+  if: Boolean! = true
+  initialCount: Int = 0
+) on FIELD
+```
+
+The `@stream` directive may be provided for a field whose type incorporates a
```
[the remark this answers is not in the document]
robrichard: I removed it from here since we have the same note higher up now covering both defer and stream

robrichard:
@benjie I updated to reflect your suggestions with one change, I combined Initial Response and Standard Response since there is a lot of overlap.

A "response" is either:

- a "single initial response"
  - data and/or errors
- or an "incremental stream"
  - a stream consisting of:
    - an "initial response"
    - followed by one or more "subsequent response" 

I've now defined Initial Response and Subsequent response separately and removed references to the word "payload". Let me know what you think

robrichard:
I opened a new [pull request with only the changes from section 3](https://github.com/graphql/graphql-spec/pull/1132). It would be great to get some reviews there. Once those changes land into the integration branch, we can focus this PR on the response section.

spec/Section 7 -- Response.md
```diff
@@ -10,31 +10,74 @@ the case that any _field error_ was raised on a field and was replaced with
 
 ## Response Format
 
-A response to a GraphQL request must be a map.
+The result of a GraphQL request must be either a single initial response or an
```
robrichard, line 13: A GraphQL request returns either a _response_, a _response stream_, or an _incremental stream_.

benjie: I think I've mentioned this before, but returning an object like this feels dangerous because it makes it more probable that the stream will not be handled? IMO GraphQL should either return an object OR a stream, and never an object that has a stream inside it.

spec/Section 7 -- Response.md
```diff
@@ -253,6 +301,164 @@ response, not in the request.
...
+
+### Incremental
+
+The `incremental` entry in the response is a non-empty list of Incremental
+Results. If the response of the GraphQL operation is an incremental stream, this
+field may appear on both the initial and subsequent responses. If present, the
+`incremental` entry must contain at least one Incremental Result.
+
+The Incremental Result is used to deliver data that the GraphQL service has
+chosen to incrementally deliver. An Incremental Result may be ether an
```
pavelglac, line 356:
```suggestion
chosen to incrementally deliver. An Incremental Result may be either an
```
typo

robrichard:
Continued in https://github.com/graphql/graphql-spec/pull/1203

PR #1125, opened 2024-11-06
nit: add rate limit extension example

sbilello:
Add Rate Limit Extension Example

### DIFF
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -27,6 +27,41 @@ set, must have a map as its value. This entry is reserved for implementers to
 extend the protocol however they see fit, and hence there are no additional
 restrictions on its contents.
 
+GraphQL Response for an extension with rate limit might look like this:
+
+```json example
+{
+    "data": {
+        "hero": {
+            "name": "R2-D2",
+            "heroFriends": [
+                {
+                    "id": "1000",
+                    "name": "Luke Skywalker"
+                },
+                {
+                    "id": "1002",
+                    "name": null
+                },
+                {
+                    "id": "1003",
+                    "name": "Leia Organa"
+                }
+            ]
+        }
+    },
+    "extensions": {
+        "customNamespace": {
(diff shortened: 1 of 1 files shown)


benjie:
Generally it's good practice to namespace the top fields in extensions to avoid conflicts, since the meaning is effectively granted to them by the entire ecosystem as a whole with no real governing documents/registry. For example, this `rateLimit` might be confused by some tooling with a different `rateLimit` format and result in errors. Instead: `sb_rateLimit` would be safer, or even `extensions:{sb:{rateLimit:{...}}}`.

Out of interest, what motivated this addition in particular?

sbilello:
@benjie  I've revised the example to better demonstrate how developers can implement rate-limiting using the extensions. Could you take a look please?
Thanks

spec/Section 7 -- Response.md
```diff
@@ -27,6 +27,43 @@ set, must have a map as its value. This entry is reserved for implementers to
...
+  },
+  "extensions": {
+    {
+      "customNamespace" : {
+        "requestRate": 2,
+        "remaining": 98,
+        "retryAfterMs": 0,
+        "resetAfterMs": 1985
+      }
+    }
```
benjie, line 61: This wasn't valid JSON since it had `{{ ... }}`

```suggestion
    "customNamespace": {
      "requestRate": 2,
      "remaining": 98,
      "retryAfterMs": 0,
      "resetAfterMs": 1985
    }
```

benjie:
I'm not convinced that the GraphQL spec needs an example of using the `extensions` property. I'll leave this open in case others feel otherwise.

spec/Section 7 -- Response.md
sbilello: fixed

sbilello:
Thanks, @benjie! I fixed the json validation.
I wanted to add an example to make sure that developers are aware about the difference between `extensions` for errors and `extensions` for not errors

PR #1127, opened 2024-11-21
When `sourceStream` errors, raise an error payload instead

benjie:
See https://github.com/graphql/graphql-spec/pull/1099/files#r1799509253

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -306,12 +306,18 @@ MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
     sourceValue)}.
   - If internal {error} was raised:
     - Cancel {sourceStream}.
-    - Perform {EmitErrorAndComplete(responseStream, error)}.
+    - Let {errors} be a list containing {error}.
+    - Let {response} be an unordered map containing {errors}.
+    - Emit {response} on {responseStream}.
+    - Complete {responseStream} normally.
   - Otherwise emit {response} on {responseStream}.
 - When {sourceStream} completes normally:
   - Complete {responseStream} normally.
 - When {sourceStream} completes with {error}:
-  - Perform {EmitErrorAndComplete(responseStream, error)}.
+  - Let {errors} be a list containing {error}.
+  - Let {response} be an unordered map containing {errors}.
+  - Emit {response} on {responseStream}.
+  - Complete {responseStream} normally.
 - When {responseStream} is cancelled:
   - Cancel {sourceStream}.
   - Complete {responseStream} normally.
@@ -336,13 +342,6 @@ ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
 
 Note: The {ExecuteSubscriptionEvent()} algorithm is intentionally similar to
 {ExecuteQuery()} since this is how each event result is produced.
-
-EmitErrorAndComplete(responseStream, error):
-
-- Let {errors} be a list containing {error}.
-- Let {response} be an unordered map containing {errors}.
-- Emit {response} on {responseStream}.
-- Complete {responseStream} normally.
 
 #### Unsubscribe
 
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -306,10 +306,7 @@ MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
     sourceValue)}.
   - If internal {error} was raised:
     - Cancel {sourceStream}.
-    - Let {errors} be a list containing {error}.
-    - Let {response} be an unordered map containing {errors}.
-    - Emit {response} on {responseStream}.
-    - Complete {responseStream} normally.
+    - Complete {responseStream} with {error}.
   - Otherwise emit {response} on {responseStream}.
 - When {sourceStream} completes normally:
   - Complete {responseStream} normally.
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -306,12 +306,12 @@ MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
     sourceValue)}.
   - If internal {error} was raised:
     - Cancel {sourceStream}.
-    - Complete {responseStream} with {error}.
+    - Perform {EmitErrorAndComplete(responseStream, error)}.
   - Otherwise emit {response} on {responseStream}.
 - When {sourceStream} completes normally:
   - Complete {responseStream} normally.
 - When {sourceStream} completes with {error}:
-  - Complete {responseStream} with {error}.
+  - Perform {EmitErrorAndComplete(responseStream, error)}.
 - When {responseStream} is cancelled:
   - Cancel {sourceStream}.
   - Complete {responseStream} normally.
@@ -336,6 +336,13 @@ ExecuteSubscriptionEvent(subscription, schema, variableValues, initialValue):
 
 Note: The {ExecuteSubscriptionEvent()} algorithm is intentionally similar to
 {ExecuteQuery()} since this is how each event result is produced.
+
+EmitErrorAndComplete(responseStream, error):
+
+- Let {errors} be a list containing {error}.
+- Let {response} be an unordered map containing {errors}.
+- Emit {response} on {responseStream}.
+- Complete {responseStream} normally.
 
 #### Unsubscribe
 


benjie:
Closing in favour of an inline edit

yaacovCR:
tried a quick implementation at https://github.com/graphql/graphql-js/pull/4302

spec/Section 6 -- Execution.md
```diff
@@ -311,7 +311,10 @@ MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
 - When {sourceStream} completes normally:
   - Complete {responseStream} normally.
 - When {sourceStream} completes with {error}:
-  - Complete {responseStream} with {error}.
+  - Let {errors} be a list containing {error}.
+  - Let {response} be an unordered map containing {errors}.
+  - Emit {response} on {responseStream}.
+  - Complete {responseStream} normally.
```
martinbonnin, line 317: Follow up discussion about whether we need an `error` event in subscriptions. @benjie @yaacovCR @enisdenjo 

From https://github.com/graphql/graphql-spec/pull/1099#discussion_r1862506201 :

> what about our own code and it's unexpected exceptions?

Our own code is perfect, no exception, never 👌🙂

Jokes aside, I guess there's always the possibility of exceptions such as out of memory, etc...

But there's no language in the spec for this for regular queries (or is there?). Why would there be for subscriptions?

From https://github.com/graphql/graphql-js/issues/4001#issuecomment-2506524425

> for example the Nginx proxy reboots, terminating all connections cleanly with an error; or the wifi drops, terminating all connections uncleanly - and we should differentiate these as best we can. 

Isn't this out of the realm of the GraphQL spec? If TCP/PHY transport fails well, that's too bad but it's another layer. GraphQL is only concerned about its own applicative layer?

The fact that [graphql-sse has been working without an `error` event](https://github.com/graphql/graphql-js/issues/4001#issuecomment-2504789068) seems to be indication that we don't really need this?

benjie: > Isn't this out of the realm of the GraphQL spec?

Yes, 100%. We cannot mandate that intermediaries generate GraphQL-shaped responses. In fact, we would expect them _NOT_ to do that. That's why making sure that our own errors _are_ formatted like GraphQL errors is so important: so that you know it hasn't originated from an intermediary.

This is why we introduced `application/graphql-response+json` in the HTTP spec, so that we don't need to HTTP 200 on error. Without that, intermediaries could produce errors in JSON format without them being GraphQL errors, and clients may get confused/crash. The GraphQL response type meant that clients could be sure they originated from GraphQL.

yaacovCR: > But there's no language in the spec for this for regular queries (or is there?). Why would there be for subscriptions?

Redirecting to @leebyron comment from https://github.com/graphql/graphql-spec/pull/1099#discussion_r1643169652

where he addresses this:

> However it's a lot less obvious what should happen for a stream which encounters an exceptional error after that particular creation algorithm has completed - what is the right behavior? This tries to be explicit to avoid that ambiguity - we should make sure the resulting stream also ends with an error.

Why is it less obvious? Well, one could have made an argument that in the spirit of GraphQL partial data, GraphQL should send a well formatted errors object for that event, and continue processing the response stream. But no! We do the more conservative thing and just end the stream with an error…

martinbonnin: > one could have made an argument that in the spirit of GraphQL partial data, GraphQL should send a well formatted errors object for that event, and continue processing the response stream. 
> We do the more conservative thing and just end the stream with an error…

I'm going to advocate for the  3rd solution: send a well formatted errors object for that event, and complete the response stream normally. 

This avoids the cognitive load associated with "completes with error" and keeps transport simple (no need for `error` event, just like graphql-sse).

Users that want to convey additional error information (error code, quota, retry delay, etc...) can do so in `error.extension` but this is outside the GraphQL spec.

enisdenjo: I like [this](https://github.com/graphql/graphql-spec/pull/1127#discussion_r1862605107).

yaacovCR: My assumption was that an internal error is a spec bug or a bug/unexpected condition within the implementation of the spec (based on @leebyron's [comment](https://github.com/graphql/graphql-spec/pull/1099#discussion_r1643169652)).

It sounds like @martinbonnin  you are suggesting changing:

```
  - Let {response} be the result of running
    {ExecuteSubscriptionEvent(subscription, schema, variableValues,
    sourceValue)}.
  - If internal {error} was raised:
    - Cancel {sourceStream}.
    - Complete {responseStream} with {error}.
```

to

```
  - Let {response} be the result of running
    {ExecuteSubscriptionEvent(subscription, schema, variableValues,
    sourceValue)}.
  - If internal {error} was raised:
    - Cancel {sourceStream}.
    - Let {errors} be a list containing {error}.
    - Let {response} be an unordered map containing {errors}.
    - Emit {response} on {responseStream}.
    - Complete {responseStream} normally.
```

But what if an internal {error} is raised within these spec lines themselves:

```
    - Cancel {sourceStream}.
    - Let {errors} be a list containing {error}.
    - Let {response} be an unordered map containing {errors}.
    - Emit {response} on {responseStream}.
    - Complete {responseStream} normally.
```

In that case, presumably our response stream will complete with that {error}, at least in languages that allow for that construct. For example, in Javascript, the call to `next()` on the iterator for the response stream will throw . I believe what @leebyron is suggesting is just that this also applies to an internal error raised by {ExecuteSubscriptionEvent()}. I don't see a particular advantage in producing a well-formed `{errors}` GraphQL response for an internal error raised by {ExecuteSubscriptionEvent()} when we don't do the same for an internal error within other algorithms (e.g. among several, ExecuteQuery()}, although I would say this is not particularly harmful, as @leebyron points out in that comment, internal errors raised within deeper logic could end up being wrapped as field errors.

Basically, because internal errors could happen anywhere, we may have to allow for the fact that our response stream might complete with an {error}. But that's not the expected case, and I am not actually sure for example if that maps to the transport level. So I am not actually sure whether this language needs to be in the spec, or maybe an accompanying note.

martinbonnin: Thanks @yaacovCR for the detailed comment 🙏 

Wrapping the internal/unexpected errors from `ExecuteSubscriptions()` in a valid GraphQL response  is indeed what I had in mind. I agree the asymmetry with `ExecuteQuery()` isn't great and feels off and might need revisiting.

All in all, I think I'd be happy not mentioning "internal errors" at all. If something unexpected comes up then behaviour is unspecified. 

**A query might**:
* send HTTP 500
* send an {errors} GraphQL response
* terminate the TCP connection
* leave the TCP connection hanging (not great)
* ...

**A subscription might**:
* send a specific `error` event
* send a WebSocket `close` frame (single connection)
* send an {errors} GraphQL response
* terminate the TCP connection
* leave the TCP connection hanging (not great)
* ...

It's not great but it's the cold reality of software. If things do not go according to plan well... we can't plan for them. 

As of today, we have only [one instance of "internal error"](https://spec.graphql.org/draft/#sel-HAPHRNDCBgEBuEvtR) in the spec. It's very specific about errors in user resolvers and I wouldn't personally give the term more importance (see also [Glossary PR](https://github.com/graphql/graphql-wg/pull/1597)).
 
Mentioning "internal errors" for the GraphQL code could erode the trust in the algorithms (why would there be an internal error there in the first place?) but also, what if there is an internal error while processing the internal error? And how should that error be represented? 

Those are interesting questions but maybe better left out for the GraphQL over HTTP/WebSocket specs?

benjie: If you leave that out, then does it mean that if the iterator ever throws an error that implementation is not spec compliant?

martinbonnin: @benjie depends what iterators we're talking about?

My current view is that errors in user code must always be wrapped. Those are errors in resolvers for queries/mutations/subscriptions as well as errors raised by the stream for subscriptions. 

Other errors are unhandled. Unhandled errors would be spec compliant but the spec doesn't say anything about how they are represented. Similar to HTTP where unhandled errors can be HTTP 500 or something else. 

GraphQL has errors/raising/handling. JS has exception/throwing/catching. Other languages may have other terminology.

enisdenjo: > send a WebSocket close frame (single connection)

Note that the close message in the frame is limited in size.

spec/Section 6 -- Execution.md
```diff
@@ -287,17 +294,36 @@ operation type.
...
-    - Let {response} be the result of running
-      {ExecuteSubscriptionEvent(subscription, schema, variableValues, event)}.
-    - Yield an event containing {response}.
-  - When {sourceStream} completes: complete {responseStream}.
+- Let {responseStream} be a new _event stream_.
+- When {sourceStream} emits {sourceValue}:
+  - Let {response} be the result of running
+    {ExecuteSubscriptionEvent(subscription, schema, variableValues,
+    sourceValue)}.
+  - If internal {error} was raised:
```
fotoetienne: I read this as an internal error while iterating through source stream. Should implementations be allowed to recover, returning the error within the response stream and continuing?

benjie:
Timestamped link to this being discussed at the GraphQL WG last week:

https://youtu.be/6bW3Xcks5I4?t=3268

I'm going to leave this open to gather more feedback. I think we should bump it to RFC1 though; so I'm going to add it to the Jan WG. Forgot to ask at the last one whilst rushing through the schedule!

enisdenjo:
As a part of graphql-ws next major release, on internal errors I've done the: "Complete {responseStream} with {error}" approach. Which is the not breaking change of the spec (but is breaking for graphql-ws).

Example:

1. Subscribe
2. Iterable throws
3. `ErrorMessage` with the caught error emitted as GraphQLFormattedError ("Complete {responseStream} with {error}")

---

Alternative approach would be to "Emit {response (with errors)} on {responseStream}" and then "Complete {responseStream} normally" - however, I have decided against this for two reasons:
1. It's not aligned with GraphQL-JS and the current spec
2. If https://github.com/graphql/graphql-js/pull/4302 lands, graphql-ws will inherit this behaviour

Example:

1. Subscribe
2. Iterable throws
3. `NextMessage` with the caught error emitted as GraphQLFormattedError (https://github.com/graphql/graphql-js/pull/4302)
4. `CompleteMessage` emitted ("Complete {responseStream} normally")

PR #1129, opened 2024-12-03
Editorial: move "Path" to it's own section

robrichard:
Extracted from the `@defer`/`@stream` spec edits (https://github.com/graphql/graphql-spec/pull/1110) to minimize the diff there.

Since the incremental delivery response format also uses `Path`, I removed it from the Errors section and into it's own section.

### DIFF
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -108,7 +108,7 @@ If an error can be associated to a particular field in the GraphQL result, it
 must contain an entry with the key `path` that details the path of the response
 field which experienced the error. This allows clients to identify whether a
 `null` result is intentional or caused by a runtime error. The value of this
-field is described in the [Path](#sec-Path) section.
+entry is described in the [Path](#sec-Path) section.
 
 For example, if fetching one of the friends' names fails in the following
 operation:
@@ -240,16 +240,16 @@ discouraged.
 
 ### Path
 
-:: A _path field_ allows for the association to a particular field in a GraphQL
-result. This field must be a list of path segments starting at the root of the
-response and ending with the field to be associated with. Path segments that
-represent fields must be strings, and path segments that represent list indices
-must be 0-indexed integers. If the path is associated to an aliased field, the
-path must use the aliased name, since it represents a path in the response, not
-in the request.
+:: A _path entry_ allows for the association with a particular field reached
+during GraphQL execution. The value for this entry must be a list of path
+segments starting at the root of the response and ending with the field to be
+associated with. Path segments that represent fields must be strings, and path
+segments that represent list indices must be 0-indexed integers. If a path
+segment is associated with an aliased field it must use the aliased name, since
+it represents a path in the response, not in the request.
 
-When the _path field_ is present on an "Error result", it indicates the response
-field which experienced the error.
+When the _path entry_ is present on an "Error result", it identifies the
+response field which experienced the error.
 
 ## Serialization Format
 
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -108,7 +108,7 @@ If an error can be associated to a particular field in the GraphQL result, it
 must contain an entry with the key `path` that details the path of the response
 field which experienced the error. This allows clients to identify whether a
 `null` result is intentional or caused by a runtime error. The value of this
-entry is described in the [Path](#sec-Path) section.
+_path entry_ is described in the [Path](#sec-Path) section.
 
 For example, if fetching one of the friends' names fails in the following
 operation:
@@ -240,15 +240,17 @@ discouraged.
 
 ### Path
 
-:: A _path entry_ allows for the association with a particular field reached
-during GraphQL execution. The value for this entry must be a list of path
-segments starting at the root of the response and ending with the field to be
-associated with. Path segments that represent fields must be strings, and path
-segments that represent list indices must be 0-indexed integers. If a path
-segment is associated with an aliased field it must use the aliased name, since
-it represents a path in the response, not in the request.
+:: A _path entry_ is an entry within an _error result_ that allows for
+association with a particular field reached during GraphQL execution.
 
-When the _path entry_ is present on an "Error result", it identifies the
+The value for a _path entry_ must be a list of path segments starting at the
+root of the response and ending with the field to be associated with. Path
+segments that represent fields must be strings, and path segments that represent
+list indices must be 0-indexed integers. If a path segment is associated with an
+aliased field it must use the aliased name, since it represents a path in the
+response, not in the request.
+
+When the _path entry_ is present on an _error result_, it identifies the
 response field which experienced the error.
 
 ## Serialization Format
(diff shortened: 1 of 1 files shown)


spec/Section 7 -- Response.md
```diff
@@ -244,6 +238,19 @@ discouraged.
 }
 ```
 
+### Path
+
+A `path` field allows for the association to a particular field in a GraphQL
```
benjie, line 243: Should we define that?

```suggestion
:: A _path field_ allows for the association to a particular field in a GraphQL
```

spec/Section 7 -- Response.md
```diff
@@ -107,14 +107,8 @@ syntax element.
...
-`null` result is intentional or caused by a runtime error.
-
-If present, this field must be a list of path segments starting at the root of
-the response and ending with the field associated with the error. Path segments
-that represent fields must be strings, and path segments that represent list
-indices must be 0-indexed integers. If the error happens in an aliased field,
-the path to the error must use the aliased name, since it represents a path in
-the response, not in the request.
+`null` result is intentional or caused by a runtime error. The value of this
+field is described in the [Path](#sec-Path) section.
```
benjie, line 111: We seem to be using "entry" and "field" interchangeably in the old text, even though we also refer to the GraphQL "field". Let's take this opportunity to tighten it up by sticking to "entry" consistently when talking about the property on an error, and "field" when talking about a GraphQL object field.

```suggestion
entry is described in the [Path](#sec-Path) section.
```

spec/Section 7 -- Response.md
```diff
@@ -244,6 +238,19 @@ discouraged.
...
 
+### Path
+
+:: A _path field_ allows for the association to a particular field in a GraphQL
+result. This field must be a list of path segments starting at the root of the
+response and ending with the field to be associated with. Path segments that
+represent fields must be strings, and path segments that represent list indices
+must be 0-indexed integers. If the path is associated to an aliased field, the
+path must use the aliased name, since it represents a path in the response, not
+in the request.
```
benjie, line 249: "a particular field in a GraphQL result" isn't strictly true for errors, because errors may cause the selection sets to explode, and thus the path may point to a field that _does not exist_ in the GraphQL result.

```suggestion
:: A _path entry_ allows for the association with a particular field reached
during GraphQL execution. The value for this entry must be a list of path
segments starting at the root of the response and ending with the field to be
associated with. Path segments that represent fields must be strings, and path
segments that represent list indices must be 0-indexed integers. If a path
segment is associated with an aliased field it must use the aliased name, since
it represents a path in the response, not in the request.
```

spec/Section 7 -- Response.md
```diff
@@ -244,6 +238,19 @@ discouraged.
...
+
+When the _path field_ is present on an "Error result", it indicates the response
+field which experienced the error.
```
benjie, line 252: I personally would use the word "identifies" here, it feels like it has more precision... But this opinion is very weakly held.

"path field" -> "path entry" should be made if the above suggestions are accepted though.

```suggestion
When the _path entry_ is present on an "Error result", it identifies the
response field which experienced the error.
```

benjie (changes requested):
Some suggested modifications:

robrichard:
@benjie thanks for the feedback! I have applied your suggestions.

spec/Section 7 -- Response.md
```diff
@@ -107,14 +107,8 @@ syntax element.
...
-`null` result is intentional or caused by a runtime error.
-
-If present, this field must be a list of path segments starting at the root of
-the response and ending with the field associated with the error. Path segments
-that represent fields must be strings, and path segments that represent list
-indices must be 0-indexed integers. If the error happens in an aliased field,
-the path to the error must use the aliased name, since it represents a path in
-the response, not in the request.
+`null` result is intentional or caused by a runtime error. The value of this
+entry is described in the [Path](#sec-Path) section.
```
benjie, line 111:
```suggestion
_path entry_ is described in the [Path](#sec-Path) section.
```

spec/Section 7 -- Response.md
```diff
@@ -244,6 +238,19 @@ discouraged.
...
 
+### Path
+
+:: A _path entry_ allows for the association with a particular field reached
+during GraphQL execution. The value for this entry must be a list of path
+segments starting at the root of the response and ending with the field to be
+associated with. Path segments that represent fields must be strings, and path
+segments that represent list indices must be 0-indexed integers. If a path
+segment is associated with an aliased field it must use the aliased name, since
+it represents a path in the response, not in the request.
```
benjie, line 249:
```suggestion
:: A _path entry_ is an entry within an _error result_ that allows for
association with a particular field reached during GraphQL execution.

The value for a _path entry_ must be a list of path segments starting at the
root of the response and ending with the field to be associated with. Path
segments that represent fields must be strings, and path segments that represent
list indices must be 0-indexed integers. If a path segment is associated with an
aliased field it must use the aliased name, since it represents a path in the
response, not in the request.
```

I tried to make this independent of error but I was just not happy with how it read. When you integrate this into the stream/defer work, you can change just the definition, e.g. something like:

```
:: A _path entry_ is an entry within a _request error_ or _incremental payload_
that allows for association with a particular field reached during GraphQL
execution.
```

spec/Section 7 -- Response.md
```diff
@@ -244,6 +238,19 @@ discouraged.
...
+
+When the _path entry_ is present on an "Error result", it identifies the
+response field which experienced the error.
```
benjie, line 252:
```suggestion
When the _path entry_ is present on an _error result_, it identifies the
response field which experienced the error.
```

This references a definition that doesn't yet exist, but that's okay... We can create it later.

benjie (changes requested):

spec/Section 7 -- Response.md
robrichard: Should this reference `_field error_`, which is already defined? Can a `_request error_` have a path?

benjie: I deliberately didn't do _field error_ because we produce an error result object from a field error or a request error, and the error result object is the thing that we discuss adding the "path" entry to. However, the spec talks about the underlying errors, the field/request errors, and the error results almost interchangeably, which is interesting. Partly the reason might be that the reference implementation captures errors and turns them into GraphQLError at source, which can then be passed throughout; I've always thought of a field error as being the error that is thrown that is captured and turned into a GraphQLError, rather than the GraphQLError itself... but the spec doesn't seem to make this distinction.

I think _error result_ is fine, since that's what we're really talking about adding a "path" entry to.

(I think a request error can have a path in very rare circumstances - e.g. if the `subscribe` method on a subscription field were to throw, I think that would be a request error, and it could be associated with the path to that field?)

robrichard: Makes sense, thanks!

benjie (approved):

mjmahone (approved):

fotoetienne (approved):

PR #1132, opened 2025-01-07
Spec edits for incremental delivery, Section 3 only

robrichard:

Helpful reference material:

Response format examples: https://github.com/graphql/defer-stream-wg/discussions/69
GraphQL Conf talk: https://www.youtube.com/watch?v=LEyDeNoobT0


spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+  initially. If omitted, defaults to `0`. A field error will be raised if the
+  value of this argument is less than `0`.
+
+Note: The ability to defer and/or stream parts of a response can have a
+potentially significant impact on application performance. Developers generally
+need clear, predictable control over their application's performance. It is
+highly recommended that GraphQL services honor the `@defer` and `@stream`
+directives on each execution. However, the specification allows advanced use
+cases where the service can determine that it is more performant to not defer
+and/or stream. Therefore, GraphQL clients _must_ be able to process a response
```
yaacovCR, line 2271: I think we discussed around this proposal that the `@defer` (and maybe `@stream`?) directives would have to be honored or not honored (with responses inlined) in the same fashion as applied to each item in a list. I think this would be important to mention, as this allows clients to cache their determinations as to whether the defer was inlined. This is relied on, for example, in https://github.com/graphql/graphql-js/pull/4319 which introduces a transformer from the new format to the legacy version.[^1]

[^1]: Note that the transformer functionality grew a bit unwieldy and will be probably be spun off if useful to a separate library outside `graphql-js`.

yaacovCR: Offline, @robrichard reminded me that although we may have discussed the above at some point, we definitely discussed servers potentially capping the number of defers, which could happen in the middle of processing a list, which would go against the above guarantee. Seeing the value in this option, I'm going to retract this comment and I've worked around this in https://github.com/graphql/graphql-js/pull/4319

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+        name
+      }
+    }
+  }
+}
+```
+
+#### @stream Arguments
+
+- `initialCount: Int` - The number of list items the service should return
```
benjie, line 2252:
```suggestion
- `initialCount: Int! = true` - The number of list items the service should return
```

spec/Section 3 -- Type System.md
```diff
@@ -1948,6 +1948,14 @@ GraphQL implementations that support the type system definition language must
 provide the `@deprecated` directive if representing deprecated portions of the
 schema.
 
+GraphQL implementations may provide the `@defer` and/or `@stream` directives. If
+either or both of these directives are provided, they must conform to the
+requirements defined in this specification.
+
+Note: The [Directives Are Defined](#sec-Directives-Are-Defined) validation rule
+ensures that GraphQL Operations containing the `@defer` or `@stream` directives
+cannot be executed by a GraphQL service that does not support them.
```
benjie, line 1957: This read as if that validation rule contains logic specific to `@stream`/`@defer`; perhaps this is clearer?

```suggestion
Note: The [Directives Are Defined](#sec-Directives-Are-Defined) validation rule
ensures that GraphQL operations can only include directives available on the
schema; thus operations including `@defer` or `@stream` directives can only be
executed by a GraphQL service that supports them.
```

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+  if: Boolean! = true
+  label: String
+) on FRAGMENT_SPREAD | INLINE_FRAGMENT
+```
+
+The `@defer` directive may be provided on a fragment spread or inline fragment
+to indicate that execution of the related selection set should be deferred. When
+a request includes the `@defer` directive, the result may consist of multiple
+responses: the initial response containing all non-deferred data, while
+subsequent responses include deferred data.
```
benjie, line 2189: I believe that we've refined the terminology around this: there's one _response_. The response for stream/defer is a stream consisting of an initial _result payload_ followed by a number of _incremental payloads_. I couldn't find where we discussed this, so please correct as appropriate.

```suggestion
payloads: the initial payload containing all non-deferred data, while subsequent
payloads include deferred data.
```

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+
+- `if: Boolean! = true` - When `true`, fragment _should_ be deferred (see
+  related note below). When `false`, fragment must not be deferred. Defaults to
+  `true` when omitted.
+- `label: String` - An optional string literal (variables are disallowed) used
+  by GraphQL clients to identify data from responses and associate it with the
+  corresponding defer directive. If provided, the GraphQL service must include
+  this label in the corresponding pending object within the response. The
+  `label` argument must be unique across all `@defer` and `@stream` directives
+  in the document.
```
benjie, line 2218: I think the use of "response" here is probably fine, but feel free to change it to be more crisp. Either way, we should explain _why_ variables are disallowed:

```suggestion
  in the document.
  
Note: Variables are disallowed in the `label` because...
```

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+  initialCount: Int! = 0
+  if: Boolean! = true
+  label: String
+) on FIELD
+```
+
+The `@stream` directive may be provided for a field whose type incorporates a
+`List` type modifier; the directive enables the backend to leverage technology
+such as asynchronous iterators to provide a partial list initially, and
+additional list items in subsequent responses.
```
benjie, line 2233:
```suggestion
The `@stream` directive may be provided for a field whose type incorporates a
`List` type modifier. The directive enables returning a partial list initially,
followed by additional items in subsequent payloads. Should the field type
incorporate multiple `List` type modifiers, only the outermost list is streamed.

Note: The mechanism through which items are streamed is implementation-defined
and may use technologies such as asynchronous iterators.
```

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+query myQuery($shouldStream: Boolean! = true) {
+  user {
+    friends(first: 10) {
+      nodes
+        @stream(label: "friendsStream", initialCount: 5, if: $shouldStream) {
+        name
+      }
+    }
+  }
+}
```
benjie, line 2247: Neither `nodes` nor `edges` appear in the GraphQL spec, so let's stick to simple lists rather than connections. I've also added `(id: 4)` to mirror [Example 18](https://spec.graphql.org/draft/#example-bcf38), and changed the name of the query to be more descriptive.

I don't like the way that prettier has chosen to format this, but this may not be the right forum for that battle.

```suggestion
query streamedFriends($shouldStream: Boolean! = true) {
  user(id: 4) {
    friends(first: 10)
      @stream(label: "friendsStream", initialCount: 5, if: $shouldStream) {
      name
    }
  }
}
```

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+```
+
+#### @stream Arguments
+
+- `initialCount: Int! = 0` - The number of list items the service should return
+  initially. If omitted, defaults to `0`. A field error will be raised if the
+  value of this argument is less than `0`.
```
benjie, line 2254:
```suggestion
- `initialCount: Int! = 0` - The number of list items to include in the initial
  result payload. If omitted, defaults to `0`. A field error will be raised if
  the value of this argument is less than `0`.
```

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+- `if: Boolean! = true` - When `true`, field _should_ be streamed (see related
+  note below). When `false`, the field must not be streamed and all list items
+  must be initially included. Defaults to `true` when omitted.
```
benjie, line 2257: The "and" here implies this is an additional behaviour

```suggestion
- `if: Boolean! = true` - When `true`, field _should_ be streamed (see related
  note below). When `false`, the field must behave as if the `@stream` directive
  is not present—it must not be streamed and all of the list items must be
  included. Defaults to `true` when omitted.
```

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+- `label: String` - An optional string literal (variables are disallowed) used
+  by GraphQL clients to identify data from responses and associate it with the
+  corresponding stream directive. If provided, the GraphQL service must include
+  this label in the corresponding pending object within the result. The `label`
+  argument must be unique across all `@defer` and `@stream` directives in the
+  document.
```
benjie, line 2263:
```suggestion
  this label in the corresponding pending object within the response. The
  `label` argument must be unique across all `@defer` and `@stream` directives
  in the document.
```

benjie: Are we forbidding multiple occurrences of the same field with a `@stream` directive (even if directive arguments match)? If not:

```graphql
{
  friends @stream(label: "A") { id }
  friends @stream(label: "A") { name }
}
```

fails this text, I think?

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2172,104 @@ to the relevant IETF specification.
...
+potentially significant impact on application performance. Developers generally
+need clear, predictable control over their application's performance. It is
+highly recommended that GraphQL services honor the `@defer` and `@stream`
+directives on each execution. However, the specification allows advanced use
+cases where the service can determine that it is more performant to not defer
+and/or stream. Therefore, GraphQL clients _must_ be able to process a response
+that ignores the `@defer` and/or `@stream` directives. This also applies to the
+`initialCount` argument on the `@stream` directive. Clients _must_ be able to
+process a streamed response that contains a different number of initial list
+items than what was specified in the `initialCount` argument.
```
benjie, line 2275: I want to make it clear that each `@stream` or `@defer` may be ignored independently of the others.

```suggestion
that ignores individual `@defer` and/or `@stream` directives. This also applies
to the `initialCount` argument on the `@stream` directive. Clients must be
able to process a streamed response that contains a different number of initial
list items than what was specified in the `initialCount` argument.
```

I've also removed the italics from `_must_` because this is explicitly a non-normative note, and shouldn't have conformance requirements in it. If this is a conformance requirement that isn't stated elsewhere, it needs to be moved outside of the `Note:`.

I also think that the final sentence is worded a bit weakly... Can the server return _any number_ of results initially, followed by _any number_ of results in follow up payloads? I personally don't think this is ideal. If a client says `@stream(initialCount: 2)` then I think the server should _either_ return _at least_ two results in the initial result payload, and stream others; _or_ not stream the result at all. If servers want to be able to always stream zero results they should add a lint rule to force `initialCount` to always be omitted or `= 0`.

benjie (changes requested):
Lots of nit-picky comments but I think we're pretty close! Do we have a glossary somewhere? I think we need to be really crisp on terms like "result", "response", "payload" and the like.

spec/Section 3 -- Type System.md
robrichard: @benjie after [the last discussion we had](https://github.com/graphql/graphql-spec/pull/1124#issuecomment-2569497795), I decided to drop `payload`. The response section now says:

> The result of a GraphQL request must be either a single initial response or an
incremental stream. The response will be an incremental stream when the GraphQL
service has deferred or streamed data as a result of the `@defer` or `@stream`
directives. When the result of the GraphQL operation is an incremental stream,
the first value will be an initial response, followed by one or more subsequent
responses.

spec/Section 3 -- Type System.md
robrichard: As of https://github.com/graphql/defer-stream-wg/discussions/100 this is no longer allowed so I think this is ok?

spec/Section 3 -- Type System.md
robrichard: I added this note below the `@stream` section and worded it so it applies to both `@defer` and `@stream`.

spec/Section 3 -- Type System.md
robrichard: I applied this change, only updating `subsequent payloads` to `subsequent responses`.

spec/Section 3 -- Type System.md
robrichard: I updated this to `The number of list items to include initially.`. I originally had something similar but there was feedback in one of the prior PRs that we shouldn't say `initial response` since `@stream` may be used inside of a `@defer`. The initial items in this case would not be returned in the initial response.

spec/Section 3 -- Type System.md
robrichard: Since we had consensus in the primary WG on https://github.com/graphql/defer-stream-wg/discussions/104, i have updated the `initialCount` section above to include: 
> When the size of the list is greater than or equal
  to the value of `initialCount`, the GraphQL service _must_ initially include
  at least as many list items as the value of `initialCount` (see related note
  below).

With this text, I think it is safe to remove italics from `must`. I believe this note is now just stating client consequences of server behavior that is described earlier.

Additionally I updated "contains a different number of initial list items" to "contains more initial list items"

robrichard:
@benjie I added a rough glossary here: https://github.com/graphql/defer-stream-wg/discussions/106, I'll keep refining it as we go

spec/Section 3 -- Type System.md
benjie: "response" is already a well defined term in the GraphQL spec: https://spec.graphql.org/draft/#sec-Response - essentially it's the `{data?: {...}, errors?: GraphQLError[]}` object. We should be careful not to confuse it with other objects such as the incremental payload objects.

spec/Section 3 -- Type System.md
benjie: That's a good subtle catch. "Initially" is a bit weakly defined though; perhaps "The number of list items to include initially when completing the parent selection set." or something along those lines? I'm not happy with the term "completing" nor "resolving" here, mind. :thinking:

spec/Section 3 -- Type System.md
benjie: Right, because if it has fewer list items it's not a "streamed response"? Good thinking.

We should clarify that "response" doesn't mean the response as a whole, but just that from the field. Perhaps "streamed field response"/"streamed field result"?

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2173,115 @@ to the relevant IETF specification.
...
+```graphql
+directive @defer(
+  if: Boolean! = true
+  label: String
+) on FRAGMENT_SPREAD | INLINE_FRAGMENT
+```
+
+The `@defer` directive may be provided on a fragment spread or inline fragment
+to indicate that execution of the related selection set should be deferred. When
+a request includes the `@defer` directive, the result may consist of multiple
```
robrichard, line 2188: When a request includes the `@defer` directive, it may return an _incremental stream_, consisting of an initial response containing all non-deferred data, followed by ... subsequent payloads including the deferred data.

mjmahone (approved):
These changes look correct to me, I don't think the definition of the `@stream/@defer` directives have changed in a very long time, so this feels good.

spec/Section 3 -- Type System.md
```diff
@@ -2164,3 +2173,115 @@ to the relevant IETF specification.
...
+  initialCount: Int! = 0
+  if: Boolean! = true
+  label: String
+) on FIELD
+```
+
+The `@stream` directive may be provided for a field whose type incorporates a
+`List` type modifier. The directive enables returning a partial list initially,
+followed by additional items in subsequent payloads. Should the field type
+incorporate multiple `List` type modifiers, only the outermost list is streamed.
```
BoD, line 2234:
```suggestion
followed by additional items in subsequent payloads. If the field type
incorporates multiple `List` type modifiers, only the outermost list is streamed.
```

Extremely minor rephrasing that I think may be more "spec sounding". Feel free to ignore if that doesn't make sense.

BoD (approved):

benjie (changes requested):
This is looking really good; all my suggestions are minor except the last one which I think warrants some extra work.

spec/Section 3 -- Type System.md
```diff
@@ -2190,3 +2199,117 @@ to the relevant IETF specification.
...
+impact on application performance. Developers generally need clear, predictable
+control over their application's performance. It is highly recommended that
+GraphQL services honor the `@defer` and `@stream` directives on each execution.
+However, the specification allows advanced use cases where the service can
+determine that it is more performant to not defer and/or stream. Therefore,
+GraphQL clients _must_ be able to process a _response_ that ignores individual
+`@defer` and/or `@stream` directives. This also applies to the `initialCount`
+argument on the `@stream` directive. Clients must be able to process a streamed
+field result that contains more initial list items than what was specified in
+the `initialCount` argument.
```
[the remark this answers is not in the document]
robrichard: Updated this to a new section and added further clarification that it's case-by-case

benjie (approved):

PR #1134, opened 2025-02-10
Type "must not be {null}" -> "must exist"

benjie:
Minor editorial fix, we use "must exist" throughout the spec, and in this case it's appropriate to use "must exist" rather than "must not be null".


martinbonnin (approved):
Better terminology 🤩 👍

robrichard (approved):

JoviDeCroock (approved):

PR #1135, opened 2025-02-24
Editorial: Add response stream to Response Section

robrichard:
I updated section 7 to formally define _Response_ as the map returned by graphql query/mutations. I am also adding a definition of "Response Stream" to section 7 as subscriptions are not mentioned in this section. The defer/stream spec edits will also add a definition for an Incremental Stream here.

I also Linked Response Stream in the execution section to this new definition, and made the existing definition of _source stream_ linkable.

### DIFF
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -14,8 +14,8 @@ A GraphQL request returns either a _response_ or a _response stream_.
 
 ### Response
 
-:: A GraphQL request will return a _response_ when the GraphQL operation is a
-query or mutation. A _response_ must be a map.
+:: A GraphQL request returns a _response_ when the GraphQL operation is a query
+or mutation. A _response_ must be a map.
 
 If the request raised any errors, the response map must contain an entry with
 key `errors`. The value of this entry is described in the "Errors" section. If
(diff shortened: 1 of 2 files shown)


spec/Section 7 -- Response.md
```diff
@@ -10,7 +10,12 @@ the case that any _field error_ was raised on a field and was replaced with
 
 ## Response Format
 
-A response to a GraphQL request must be a map.
+A GraphQL request returns either a _response_ or a _response stream_.
+
+### Response
+
+:: A GraphQL request will return a _response_ when the GraphQL operation is a
```
martinbonnin, line 17: Maybe more of a personal thing but I tend to avoid future tense in spec/docs. I like that it "sounds" more universal and is generally easier to read. But I just checked the spec and there's a fair amount of future tense there so take this with a pinch of salt.

```suggestion
:: A GraphQL request returns a _response_ when the GraphQL operation is a
```

robrichard: Agreed, updated

martinbonnin (approved):

BoD (approved):

leebyron (approved):
pure win. nice cleanup and makes this more accurate

PR #1136, opened 2025-02-24
Field merging validation: clarify pair members are distinct

andimarek:
Hey ... I hope that can be seen as editorial clarification:

In the overlapping field merging validation the term "pair" is used twice, without making it perfectly clear that the two pair members should be distinct. 

In theory they don't have to be different, therefore this PR clarifies it.

Andi

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -420,7 +420,7 @@ FieldsInSetCanMerge(set):
 
 - Let {fieldsForName} be the set of selections with a given response name in
   {set} including visiting fragments and inline fragments.
-- Given each pair of members {fieldA} and {fieldB} in {fieldsForName}:
+- Given each pair of distinct members {fieldA} and {fieldB} in {fieldsForName}:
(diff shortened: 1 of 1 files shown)


martinbonnin (approved):

benjie (approved):

andimarek:
What was our process for editorial changes again? How long do we wait after getting some approvals? cc @benjie ?

benjie:
I generally give it a couple weeks and then merge if it has approvals. (This is already in [my queue](https://github.com/users/benjie/projects/4).)

But this one is so trivial, we might as well go ahead and merge it.

benjie:
I was just reading it again and I don’t think it really matters whether or not they’re distinct, since they will have the same response shape if they’re the same… But more efficient this way 🤷

andimarek:
@benjie yes it is more performant this way, but maybe even a bit more important it makes it more clear why it should run on every SelectionSet instead of per Operation. If the fields don't need to be distinct you could run the whole validation once. 

So overall this version I believe make more sense. Thanks for merging!

PR #1143, opened 2025-03-06
Editorial: Define _response map_

robrichard:
Follow up from https://github.com/graphql/graphql-spec/pull/1135, Based on top of https://github.com/graphql/graphql-spec/pull/1148

This creates a definition for _Response Map_, defined as the map normally returned by GraphQL queries and mutations containing data and/or errors. The term `response map` is already being use so I formally defined it and used the defined term throughout the rest of the spec. I also have an alternative proposal, [Response Payload](https://github.com/graphql/graphql-spec/pull/1149), as I think that fits better in the direction Incremental Delivery will be going.

The definition of Response is updated to include both Response Map, and Response Stream (for subscriptions).

If we go with response map, I think the end result with Incremental Delivery will be something like:

```ts
type ResponseMap = {
  data?: object
  errors?: Error[]
  extensions?: object
}
type ResponseStream = ResponseMap[]
type IncrementalStream = [InitialResponseMap, ...SubsequentResponseMap];
type Response = ResponseMap | ResponseStream | IncrementalStream;
```

~I went through every mention of response and most of the remaining usages are around response key or response field, referring to the object property name (ie for aliases and error paths). This could also likely be standardized in a follow up.~ See https://github.com/graphql/graphql-spec/pull/1147/files

### DIFF
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -10,71 +10,75 @@ the case that any _field error_ was raised on a field and was replaced with
 
 ## Response Format
 
-A GraphQL request returns either a _response_ or a _response stream_.
+:: A GraphQL _response_ the returned result of a GraphQL _request_. A _response_
+is either a _response map_ or a _response stream_.
 
-### Response
+### Response Map
 
-:: A GraphQL request returns a _response_ when the GraphQL operation is a query
-or mutation. A _response_ must be a map.
+:: A GraphQL request returns a _response map_ when the GraphQL operation is a
+query or mutation. When the GraphQL operation is a subscription, the GraphQL
+request returns a _response map_ if a _request error_ is raised. A response map
+must be a map.
 
-If the request raised any errors, the response map must contain an entry with
+If the request raised any errors, the _response map_ must contain an entry with
 key `errors`. The value of this entry is described in the "Errors" section. If
 the request completed without raising any errors, this entry must not be
 present.
 
-If the request included execution, the response map must contain an entry with
+If the request included execution, the _response map_ must contain an entry with
 key `data`. The value of this entry is described in the "Data" section. If the
 request failed before execution, due to a syntax error, missing information, or
 validation error, this entry must not be present.
 
-The response map may also contain an entry with key `extensions`. This entry, if
-set, must have a map as its value. This entry is reserved for implementers to
+The _response map_ may also contain an entry with key `extensions`. This entry,
+if set, must have a map as its value. This entry is reserved for implementers to
 extend the protocol however they see fit, and hence there are no additional
 restrictions on its contents.
 
 To ensure future changes to the protocol do not break existing services and
-clients, the top level response map must not contain any entries other than the
-three described above.
+clients, the top level _response map_ must not contain any entries other than
+the three described above.
 
-Note: When `errors` is present in the response, it may be helpful for it to
-appear first when serialized to make it more clear when errors are present in a
-response during debugging.
+Note: When `errors` is present in the _response map_, it may be helpful for it
+to appear first when serialized to make it more clear when errors are present in
+a response map during debugging.
 
 ### Response Stream
 
 :: A GraphQL request returns a _response stream_ when the GraphQL operation is a
-subscription. A _response stream_ must be a stream of _response_.
+subscription and a _request error_ is not raised. A response stream must be a
+stream of _response map_.
 
 ### Data
 
-The `data` entry in the response will be the result of the execution of the
-requested operation. If the operation was a query, this output will be an object
-of the query root operation type; if the operation was a mutation, this output
-will be an object of the mutation root operation type.
+The `data` entry in the _response map_ will be the result of the execution of
+the requested operation. If the operation was a query, this output will be an
+object of the query root operation type; if the operation was a mutation, this
+output will be an object of the mutation root operation type.
 
 If an error was raised before execution begins, the `data` entry should not be
-present in the response.
+present in the _response map_.
 
-If an error was raised during the execution that prevented a valid response, the
-`data` entry in the response should be `null`.
+If an error was raised during execution that propagated to the operation root,
+the `data` entry in the _response map_ should be `null`.
 
 ### Errors
 
-The `errors` entry in the response is a non-empty list of errors raised during
-the _request_, where each error is a map of data described by the error result
-format below.
+The `errors` entry in the _response map_ is a non-empty list of errors raised
+during the _request_, where each error is a map of data described by the error
+result format below.
 
-If present, the `errors` entry in the response must contain at least one error.
-If no errors were raised during the request, the `errors` entry must not be
-present in the response.
+If present, the `errors` entry in the _response map_ must contain at least one
+error. If no errors were raised during the request, the `errors` entry must not
+be present in the response map.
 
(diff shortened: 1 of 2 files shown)


spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,71 @@ the case that any _field error_ was raised on a field and was replaced with
...
-response during debugging.
+Note: When `errors` is present in the _response map_, it may be helpful for it
+to appear first when serialized to make it more clear when errors are present in
+a _response map_ during debugging.
 
 ### Response Stream
 
 :: A GraphQL request returns a _response stream_ when the GraphQL operation is a
-subscription. A _response stream_ must be a stream of _response_.
+subscription. A _response stream_ must be a stream of _response map_.
```
benjie, line 46: This isn't actually true, a subscription can yield a _response map_ if a request error occurs. This should be reflected in the response map definition too.

spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,71 @@ the case that any _field error_ was raised on a field and was replaced with
...
-A GraphQL request returns either a _response_ or a _response stream_.
+:: A GraphQL _response_ is either a _response map_ or a _response stream_.
 
-### Response
+### Response Map
 
-:: A GraphQL request returns a _response_ when the GraphQL operation is a query
-or mutation. A _response_ must be a map.
+:: A GraphQL request returns a _response map_ when the GraphQL operation is a
+query or mutation. A _response map_ must be a map.
```
benjie, line 18: A response map can be used in many situations. We should try and make the description more accurate. I don't particularly like this edit, but it does seem more accurate.

```suggestion
:: A _response map_ is a map containing data and/or errors returned from GraphQL
execution.

Note: A _response map_ is typically the result type that you would see for a
query or mutation operation, however you may also see it in response to a
subscription operation should a request error occur.
```

benjie (changes requested):
I like this in general, but it needs to be made a little more accurate :+1:

spec/Section 7 -- Response.md
benjie, line 46:
```suggestion
subscription and the subscription is successfully established (no request error
was raised). A _response stream_ must be a stream of _response map_.
```

spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,74 @@ the case that any _field error_ was raised on a field and was replaced with
...
+output will be an object of the mutation root operation type.
 
 If an error was raised before execution begins, the `data` entry should not be
-present in the response.
+present in the _response map_.
 
-If an error was raised during the execution that prevented a valid response, the
-`data` entry in the response should be `null`.
+If an error was raised during the execution that prevented a valid result, the
+`data` entry in the _response map_ should be `null`.
```
benjie, line 62:
```suggestion
If an error was raised during execution that propagated to the operation root,
the `data` entry in the _response map_ should be `null`.
```

benjie (approved):
Looks good to me :+1: 

(I have _not_ done a full review of the spec to ensure we caught everything.)

benjie:
I prefer https://github.com/graphql/graphql-spec/pull/1149 - "response payload" is much clearer IMO.

martinbonnin:
I like `payload` better too 👍 

Alternatively, did we already discuss reusing the [graphql-js terminology](https://github.com/martinbonnin/graphql-js/blob/aec37aa057e950871821be917a854eb1cacab639/src/execution/types.ts#L19)?

```ts
/**
 * The result of GraphQL execution.
 *
 *   - `errors` is included when any errors occurred as a non-empty array.
 *   - `data` is the result of a successful execution of the query.
 *   - `hasNext` is true if a future payload is expected.
 *   - `extensions` is reserved for adding non-standard properties.
 *   - `incremental` is a list of the results from defer/stream directives.
 */
export interface ExecutionResult {
  errors?: ReadonlyArray<GraphQLError>;
  data?: TData | null;
  extensions?: TExtensions;
}
```

benjie:
I like that, but it would involve renaming a few other things too. In particular `ResponseStream` being a stream of `ExecutionResult` is less clear. Since we're introducing `IncrementalStream`, I wonder if just calling it `SubscriptionStream` would be better? Then we'd end up with:

```ts
type Response = ExecutionResult | SubscriptionStream | IncrementalStream;
type ExecutionResult = {
  data?: object
  errors?: Error[]
  extensions?: object
}
type SubscriptionStream = ExecutionResult[]
type IncrementalStream = [InitialExecutionResult, ...SubsequentExecutionResult];

type SubsequentExecutionResult = {
  incremental?: IncrementalResult[]
  pending?: Pending[]
  completed?: CompletedResult[],
  hasNext?: boolean;
}

type InitialExecutionResult =
  ExecutionResult &
  Partial<IncrementalExecutionResult> &
  { hasNext: true };

type IncrementalResult = IncrementalObjectResult | IncrementalListResult
```

This kind of works?

martinbonnin:
That'd work for me 👍 . As I was explaining in the [other pr](https://github.com/graphql/graphql-spec/pull/1149#discussion_r1995191798), I expect `FooStream` to be a stream of `Foo`. If we wanted to keep things a bit more lightweight, we could also remove the `Execution` from `ExecutionResult`:

```ts
type Response = Result | ResultStream | IncrementalResultStream;
type Result = {
  data?: object
  errors?: Error[]
  extensions?: object
}
type ResultStream = Result[]
type IncrementalResultStream = [InitialResult, ...SubsequentResult];

type SubsequentResult = {
  incremental?: IncrementalResult[]
  pending?: Pending[]
  completed?: CompletedResult[],
  hasNext?: boolean;
}

type InitialResult =
  Result &
  Partial<IncrementalResult> &
  { hasNext: true };

type IncrementalResult = IncrementalObjectResult | IncrementalListResult
```

benjie:
> If we wanted to keep things a bit more lightweight, we could also remove the Execution from ExecutionResult.

We use the term "result" throughout the GraphQL spec, so I think that calling it "result" would be dangerous. ("response" is not used so generically, so doesn't suffer the same issue.)

>  I expect `${Foo}Stream` to be a stream of `Foo`

That's how it's used currently, but I think it's also reasonable for it to be a name for the stream - e.g. "SubscriptionStream" is clearly to me "a stream for a subscription" rather than "a stream of Subscription". We often use prefixes to name things to disambiguate them, e.g. in `CollectFields()` we [use `fragmentGroupedFieldSet`](https://spec.graphql.org/draft/#sel-LANRPLCFFHFRCAACCA1hK) to mean "the groupedFieldSet related to this fragment".

leebyron:
I think we should keep "Response" as the general concept and qualify it with each variant. Response Map, Incremental Response Map, Response Stream

This will present longer names, but generally we should aim to avoid long term ambiguity if possible

martinbonnin:
@leebyron "Response Stream" is really ambiguous to me. Is it a response or is it a stream of responses? Or both?

martinbonnin:
Also interesting question is how we look at this in the context of GraphQL over HTTP. 

Do we consider that `application/graphql-response+json` may contain incremental/subscription responses (feels quite weird IMO). 

Incremental responses could potentially use jsonl format (`application/graphql-response+jsonl`) but not having the media type discriminate what type of response is expected feels a bit weird.

robrichard:
Superseded by https://github.com/graphql/graphql-spec/pull/1159

PR #1147, opened 2025-03-10
Consistently use 'response name' not 'response key'

### LINKED ISSUE #1144: Clarify "response key" vs "response name"
Both terms are used in the spec and it appears that they are the same thing.

* [Sec. 2.7](https://spec.graphql.org/draft/#sec-Field-Alias) defines "response key" as
  * "By default a field’s response key in the response object will use that field’s name. However, you can define a different response key by specifying an alias."
* [Sec. 5.3.2](https://spec.graphql.org/draft/#sec-Field-Selection-Merging) uses "response name" without a definition.
* [Sec. 6.3.2](https://spec.graphql.org/draft/#sec-Field-Collection) uses "response key" with a definition in parentheses as
  * "the alias if defined, otherwise the field name."

Consistency would be appreciated. Or, if they are different, please add a definition for response name.

I see some implementations use "response name" and others use "response key". In the graphql-js codebase, only `responseName` is used.
* [OverlappingFieldsCanBeMergedRule.ts#L86](https://github.com/graphql/graphql-js/blob/1437cda8d19bb76823ff056ecc50cdb1eaa1f4ea/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts#L86)
* [execute.ts#L409](https://github.com/graphql/graphql-js/blob/1437cda8d19bb76823ff056ecc50cdb1eaa1f4ea/src/execution/execute.ts#L409)

benjie:
Extracts the definition of "response key" from #1039, replaces occurrences of "response key" with "response name" for consistency, and links in relevant places.

Fixes #1144.

Why "response name" and not "response key"? Previously we used a mixture to mean the same thing, and GraphQL.js generally uses `responseName` (`responseKey` is only used in the new code relating to incremental delivery).

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -466,8 +466,9 @@ These two operations are semantically identical:
 
 Alias : Name :
 
-By default a field's response key in the response object will use that field's
-name. However, you can define a different response key by specifying an alias.
+:: A _response key_ is the key in the response object which correlates with a
+field's result. By default the response key will use the field's name; however,
+you can define a different response key by specifying an alias.
 
 In this example, we can fetch two profile pictures of different sizes and ensure
 the resulting response object will not have duplicate keys:
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -466,9 +466,9 @@ These two operations are semantically identical:
 
 Alias : Name :
 
-:: A _response key_ is the key in the response object which correlates with a
-field's result. By default the response key will use the field's name; however,
-you can define a different response key by specifying an alias.
+:: A _response name_ is the key in the response object which correlates with a
+field's result. By default the response name will use the field's name; however,
+you can define a different response name by specifying an alias.
 
 In this example, we can fetch two profile pictures of different sizes and ensure
 the resulting response object will not have duplicate keys:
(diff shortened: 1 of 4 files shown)


spec/Section 2 -- Language.md
```diff
@@ -466,8 +466,9 @@ These two operations are semantically identical:
 
 Alias : Name :
 
-By default a field's response key in the response object will use that field's
-name. However, you can define a different response key by specifying an alias.
+:: A _response name_ is the key in the response object which correlates with a
```
robrichard, line 469: Thinking out loud -
Should `response object` be updated to `response map` in https://github.com/graphql/graphql-spec/pull/1039?

But then this will not be accurate with incremental delivery, as a `response name` can also appear in the subsequent payloads. Should there be a new term (Response Data Tree?) to refer to the entire response data?

benjie: I think this is referring to the object in the response representing the current Object Type's data, as opposed to the entire response which may use multiple names/aliases which may or may not overlap. It's a good point though, we should be clearer about this!

martinbonnin (approved):

BoD (approved):

michaelstaib (approved):

mjmahone (approved):
Personally I prefer `response key` but consistency is preferred to not, so not blocking.

PascalSenn (approved):

leebyron (approved):

PR #1149, opened 2025-03-13
Editorial: Define _response payload_

robrichard:
Follow up from https://github.com/graphql/graphql-spec/pull/1135, Based on top of https://github.com/graphql/graphql-spec/pull/1148

Alternative name `Response Map`: https://github.com/graphql/graphql-spec/pull/1143

This creates a definition for Response Payload, defined as the map normally returned by GraphQL queries and mutations containing data and/or errors.

The definition of Response is updated to include both Response Payload, and Response Stream (for subscriptions).

I like using `Payload` as i think it extends naturally for incremental delivery. If this change is approved I plan to add two additional types, `InitialResponsePayload` and `SubsequentResponsePayload` as well as `IncrementalStream`, the stream of those types.

Proposal for incremental delivery (roughly):

```ts
type ResponsePayload = {
  data?: object
  errors?: Error[]
  extensions?: object
}
type ResponseStream = ResponsePayload[]
type IncrementalStream = [InitialResponsePayload, ...SubsequentResponsePayload];
type Response = ResponsePayload | ResponseStream | IncrementalStream;
```

~I went through every mention of response and most of the remaining usages are around response key or response field, referring to the object property name (ie for aliases and error paths). This could also likely be standardized in a follow up.~ See https://github.com/graphql/graphql-spec/pull/1147/files

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -301,13 +301,13 @@ MapSourceToResponseEvent(sourceStream, subscription, schema, variableValues):
 
 - Let {responseStream} be a new _event stream_.
 - When {sourceStream} emits {sourceValue}:
-  - Let {response} be the result of running
+  - Let {responsePayload} be the result of running
     {ExecuteSubscriptionEvent(subscription, schema, variableValues,
     sourceValue)}.
   - If internal {error} was raised:
     - Cancel {sourceStream}.
     - Complete {responseStream} with {error}.
-  - Otherwise emit {response} on {responseStream}.
+  - Otherwise emit {responsePayload} on {responseStream}.
 - When {sourceStream} completes normally:
   - Complete {responseStream} normally.
 - When {sourceStream} completes with {error}:
@@ -804,22 +804,22 @@ A _field error_ is an error raised from a particular field during value
 resolution or coercion. While these errors should be reported in the response,
 they are "handled" by producing a partial response.
 
-Note: This is distinct from a _request error_ which results in a response with
-no data.
+Note: This is distinct from a _request error_ which results in a _response
+payload_ with no {"data"}.
 
 If a field error is raised while resolving a field, it is handled as though the
 field returned {null}, and the error must be added to the {"errors"} list in the
-response.
+_response payload_.
 
 If the result of resolving a field is {null} (either because the function to
 resolve the field returned {null} or because a field error was raised), and that
 field is of a `Non-Null` type, then a field error is raised. The error must be
-added to the {"errors"} list in the response.
+added to the {"errors"} list in the _response payload_.
 
 If the field returns {null} because of a field error which has already been
-added to the {"errors"} list in the response, the {"errors"} list must not be
-further affected. That is, only one error should be added to the errors list per
-field.
+added to the {"errors"} list in the _response payload_, the {"errors"} list must
+not be further affected. That is, only one error should be added to the errors
+list per field.
 
 Since `Non-Null` type fields cannot be {null}, field errors are propagated to be
 handled by the parent field. If the parent field may be {null} then it resolves
@@ -832,5 +832,5 @@ type is also wrapped in a `Non-Null`, the field error continues to propagate
 upwards.
 
 If all fields from the root of the request to the source of the field error
-return `Non-Null` types, then the {"data"} entry in the response should be
-{null}.
+return `Non-Null` types, then the {"data"} entry in the _response payload_
+should be {null}.
--- a/spec/Section 7 -- Response.md
+++ b/spec/Section 7 -- Response.md
@@ -10,71 +10,75 @@ the case that any _field error_ was raised on a field and was replaced with
 
 ## Response Format
 
-A GraphQL request returns either a _response_ or a _response stream_.
+:: A GraphQL _response_ is the returned result of a GraphQL _request_. A
+_response_ is either a _response payload_ or a _response stream_.
 
-### Response
+### Response Payload
 
-:: A GraphQL request returns a _response_ when the GraphQL operation is a query
-or mutation. A _response_ must be a map.
+:: A GraphQL request returns a _response payload_ when the GraphQL operation is
+a query or mutation. When the GraphQL operation is a subscription, the GraphQL
+request returns a _response payload_ if a _request error_ is raised. A response
+payload must be a map.
 
-If the request raised any errors, the response map must contain an entry with
-key `errors`. The value of this entry is described in the "Errors" section. If
-the request completed without raising any errors, this entry must not be
+If the request raised any errors, the _response payload_ must contain an entry
+with key `errors`. The value of this entry is described in the "Errors" section.
+If the request completed without raising any errors, this entry must not be
 present.
 
-If the request included execution, the response map must contain an entry with
-key `data`. The value of this entry is described in the "Data" section. If the
-request failed before execution, due to a syntax error, missing information, or
-validation error, this entry must not be present.
+If the request included execution, the _response payload_ must contain an entry
+with key `data`. The value of this entry is described in the "Data" section. If
+the request failed before execution, due to a syntax error, missing information,
+or validation error, this entry must not be present.
 
-The response map may also contain an entry with key `extensions`. This entry, if
-set, must have a map as its value. This entry is reserved for implementers to
-extend the protocol however they see fit, and hence there are no additional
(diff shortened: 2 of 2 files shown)


spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,74 @@ the case that any _field error_ was raised on a field and was replaced with
 
 ## Response Format
 
-A GraphQL request returns either a _response_ or a _response stream_.
+:: A GraphQL _response_ is either a _response payload_ or a _response stream_.
```
benjie, line 13:
```suggestion
:: A GraphQL _response_ is the returned result of a GraphQL _request_. A
_response_ is either a _response payload_ or a _response stream_.
```

spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,74 @@ the case that any _field error_ was raised on a field and was replaced with
...
+the `data` entry in the _response payload_ should be `null`.
 
 ### Errors
 
-The `errors` entry in the response is a non-empty list of errors raised during
-the _request_, where each error is a map of data described by the error result
-format below.
+The `errors` entry in the _response payload is a non-empty list of errors raised
+during the \_request_, where each error is a map of data described by the error
+result format below.
```
benjie, line 68:
```suggestion
The `errors` entry in the _response payload_ is a non-empty list of errors
raised during the _request_, where each error is a map of data described by the
error result format below.
```

benjie (approved):
I approve, I think this makes it a lot clearer.

spec/Section 6 -- Execution.md
```diff
@@ -805,22 +804,22 @@ A _field error_ is an error raised from a particular field during value
 resolution or coercion. While these errors should be reported in the response,
 they are "handled" by producing a partial response.
 
-Note: This is distinct from a _request error_ which results in a response with
-no data.
+Note: This is distinct from a _request error_ which results in a _response
+payload_ with no data.
```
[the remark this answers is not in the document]
benjie: Including this is entirely optional, perhaps it should be pulled into a separate PR.

spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,75 @@ the case that any _field error_ was raised on a field and was replaced with
 
 ## Response Format
 
-A GraphQL request returns either a _response_ or a _response stream_.
+:: A GraphQL _response_ is the returned result of a GraphQL _request_. A
+_response_ is either a _response payload_ or a _response stream_.
```
martinbonnin, line 14: This itches me a little bit I expect a `response stream` to be a stream of `responses` but here the definition is a bit circular. We can really define a `response` as being a `response stream` we just went into infinite recursion.

Maybe invert the logic?

```
:: A GraphQL _response_ is the returned result of a GraphQL _request_. A
_response_ is either a _payload response_ or a _streamed response_.
```

martinbonnin: Alternatively, something like so?

```
:: A GraphQL _response_ is the returned result of a GraphQL _request_. A
_response_ is either a _payload_ or a _payload stream_.
```

robrichard: @martinbonnin  I like `response payload` over `payload response` or `payload` because I think it extends naturally to the types we need to add for incremental delivery  `InitialResponsePayload`, ` SubsequentResponsePayload` (see top level description: https://github.com/graphql/graphql-spec/pull/1149#issue-2917891993).

I agree that `Response Stream` could perhaps be more clear. Personally I like `Subscription Stream`. The current proposal for incremental delivery does not allow `@defer` and `@stream` in subscriptions, but I imagine we'd like to add that eventually. So this stream may eventually be a stream of more than just `Response Payload`s.

But `Response Stream` is already the established name and this PR is not attempting to change that. Maybe we could continue that discussion as a follow on of this?

martinbonnin: IMO the naming should be discussed in globality if we want something ultimately consistent (HTTP media types are also a concern). 

Looking back at https://github.com/graphql/graphql-spec/pull/1135/files, it filled a gap in that subscription were not accounted to in the spec. This part was very welcome but I'm having second thoughts about "response stream". 

"response stream" has 2 important drawbacks:

* It can be thought as a "stream of responses" (which it is not).
* It breaks the request/response symmetry. I think it's quite natural that the response is the result of a request. That sentence: "A GraphQL request returns either a _response_ or a _response stream_." should really stop at "A GraphQL request returns a _response_."

I'm a bit wary that adding things incrementally (no pun intended) makes us lose the big picture here.

martinbonnin: [Example where the spec is currently inconsistent](https://github.com/graphql/graphql-spec/blob/b41339a0fd22fb3ca9a1a6110334a7094f1f7643/spec/Section%206%20--%20Execution.md#L20):

```
Given this information, the result of {ExecuteRequest(schema, document,
operationName, variableValues, initialValue)} produces the response, to be
formatted according to the Response section below.
```

Should we change this to "produces the response **or the response stream**? That feels really weird.

robrichard: @martinbonnin this PR is addressing your second bullet
> It breaks the request/response symmetry. I think it's quite natural that the response is the result of a request. That sentence: "A GraphQL request returns either a response or a response stream." should really stop at "A GraphQL request returns a response."

We are updating `Response` to mean the thing that a GraphQL request returns, so the sentence `"A GraphQL request returns a response."` is accurate after these changes.

I believe this also fixes the inconsistency you pointed out in the execution section

martinbonnin: > I believe this also fixes the inconsistency you pointed out in the execution section

Right! So that's taken care of. Thanks!

I`m still holding (for now) that `Response Stream` feels a bit weird but it's not like I have many other alternative propositions so I'll adhere to the majority here.

spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,75 @@ the case that any _field error_ was raised on a field and was replaced with
...
+output will be an object of the mutation root operation type.
 
 If an error was raised before execution begins, the `data` entry should not be
-present in the response.
+present in the _response payload_.
 
-If an error was raised during the execution that prevented a valid response, the
-`data` entry in the response should be `null`.
+If an error was raised during execution that propagated to the operation root,
+the `data` entry in the _response payload_ should be `null`.
```
benjie, line 63: Note to other reviewers: this line contains a subtle fix to the wording. :+1:

benjie (approved):
The term "payload" is used twice in the spec currently, each time to refer to the `{data,errors}` object that a subscription stream may yield on each event. This is consistent with our use of "response payload" here, so I approve :+1:

spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,75 @@ the case that any _field error_ was raised on a field and was replaced with
...
-three described above.
+clients, the top level _response payload_ must not contain any entries other
+than the three described above.
 
-Note: When `errors` is present in the response, it may be helpful for it to
-appear first when serialized to make it more clear when errors are present in a
-response during debugging.
+Note: When `errors` is present in the _response payload_, it may be helpful for
+it to appear first when serialized to make it more clear when errors are present
+in a response payload during debugging.
```
BoD, line 44:
```suggestion
Note: When `errors` is present in the _response payload_, it may be helpful for
it to appear first when serialized to make it more clear that errors are present.
```

nitpick, I thought the repetition was a bit odd, feel free to ignore

BoD (approved):

michaelstaib (approved):

spec/Section 7 -- Response.md
```diff
@@ -10,71 +10,75 @@ the case that any _field error_ was raised on a field and was replaced with
...
+:: A GraphQL _response_ is the returned result of a GraphQL _request_. A
+_response_ is either a _response payload_ or a _response stream_.
 
-### Response
+### Response Payload
 
-:: A GraphQL request returns a _response_ when the GraphQL operation is a query
-or mutation. A _response_ must be a map.
+:: A GraphQL request returns a _response payload_ when the GraphQL operation is
+a query or mutation. When the GraphQL operation is a subscription, the GraphQL
```
Keweiqu, line 19: This sentence is no longer accurate. A GraphQL query with incremental delivery no longer returns a response payload.

robrichard:
Superseded by https://github.com/graphql/graphql-spec/pull/1159

PR #1152, opened 2025-03-27
Rename field error to execution error; define response position

benjie:
This has bugged me for years, but it turns out to be a blocker around disabling error propagation... so I'm proposing we finally fix it.

The GraphQL spec is _wrong_ when it refers to field errors. Case in point:

> If all fields from the root of the request to the source of the field error return Non-Null types, then the "data" entry in the response should be null. 

This is not true for the schema `type Query { foo: [Foo]! } type Foo { throws: Int! }` with the query `{ foo { throws } }`. All fields from the root of the request to the source of the field error return Non-Null types, but the _list position_ inside the `foo` return type is nullable, and thus the error should (and does!) stop at that error boundary, returning `{ data: {foo: [null]}, errors: [...]}` and not `{ data: null, errors: [...] }` as the spec (incorrectly) states.

This seems to be an oversight of the list type throughout the spec; there are a lot of incidences where we refer to errors happening at a "field" when actually we mean they're happening at an errorable position - that is to say, a position that can be identified by the `path` in the response. Factoring in that this `path` will also want to be used by incremental delivery, rather than calling it an "error position" I've called it a "response position". I did moot calling it simply "position", which I really like, but we do have instances of referring to argument positions, which are distinct from response positions, so I opted for the longer but less ambiguous name.

To make this clearer I've renamed "field error" to "execution error". Technically all execution errors originate in fields, but not necessarily at the field itself - it could be nested within a list. There are three main sources:

- Argument coercion - happens at the field
- Errors thrown by the resolver and caught by GraphQL - happens at the field[^1]
- Errors occurring during result coercion - happens at the field, **and within any nested list positions**

[^1]: in spec terms, though in GraphQL.js it's possible to return e.g. `[1, new Error("two"), 3]` and have the second item be treated as an error thrown in that response position.

I've introduced `<a name="...">` tags in the spec text to maintain these key hyperlinks even after changing titles of sections.

I've also fixed a number of bugs, ambiguities, and omissions in the spec relating to error handling.

These all reflect the behavior of GraphQL.js (and, I believe, all other GraphQL implementations) and thus I've marked this as editorial.

@graphql/implementers Do any of your implementations handle errors in the way that the spec _currently_ states, as opposed to the way that the reference implementation works? If so, speak now and we can turn this into an RFC process rather than editorial!

cc @JoviDeCroock @yaacovCR Please confirm that my spec edits correctly represent the way in which GraphQL.js operates.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -329,8 +329,8 @@ A GraphQL schema may describe that a field represents a list of another type;
 the `List` type is provided for this reason, and wraps another type.
 
 Similarly, the `Non-Null` type wraps another type, and denotes that the
-resulting value will never be {null} (and that a _runtime error_ cannot result
-in a {null} value).
+resulting value will never be {null} (and that an _execution error_ cannot
+result in a {null} value).
 
 These two types are referred to as "wrapping types"; non-wrapping types are
 referred to as "named types". A wrapping type has an underlying named type,
@@ -461,14 +461,14 @@ more guidance.
 
 A GraphQL service, when preparing a field of a given scalar type, must uphold
 the contract the scalar type describes, either by coercing the value or
-producing a _runtime error_ if a value cannot be coerced or if coercion may
+producing an _execution error_ if a value cannot be coerced or if coercion may
 result in data loss.
 
 A GraphQL service may decide to allow coercing different internal types to the
 expected return type. For example when coercing a field of type {Int} a boolean
 {true} value may produce {1} or a string value {"123"} may be parsed as base-10
 {123}. However if internal type coercion cannot be reasonably performed without
-losing information, then it must raise a _runtime error_.
+losing information, then it must raise an _execution error_.
 
 Since this coercion behavior is not observable to clients of the GraphQL
 service, the precise rules of coercion are left to the implementation. The only
@@ -513,15 +513,15 @@ Fields returning the type {Int} expect to encounter 32-bit integer internal
 values.
 
 GraphQL services may coerce non-integer internal values to integers when
-reasonable without losing information, otherwise they must raise a _runtime
+reasonable without losing information, otherwise they must raise an _execution
 error_. Examples of this may include returning `1` for the floating-point number
 `1.0`, or returning `123` for the string `"123"`. In scenarios where coercion
-may lose data, raising a runtime error is more appropriate. For example, a
-floating-point number `1.2` should raise a runtime error instead of being
+may lose data, raising an execution error is more appropriate. For example, a
+floating-point number `1.2` should raise an execution error instead of being
 truncated to `1`.
 
 If the integer internal value represents a value less than -2<sup>31</sup> or
-greater than or equal to 2<sup>31</sup>, a _runtime error_ should be raised.
+greater than or equal to 2<sup>31</sup>, an _execution error_ should be raised.
 
 **Input Coercion**
 
@@ -548,12 +548,12 @@ Fields returning the type {Float} expect to encounter double-precision
 floating-point internal values.
 
 GraphQL services may coerce non-floating-point internal values to {Float} when
-reasonable without losing information, otherwise they must raise a _runtime
+reasonable without losing information, otherwise they must raise an _execution
 error_. Examples of this may include returning `1.0` for the integer number `1`,
 or `123.0` for the string `"123"`.
 
 Non-finite floating-point internal values ({NaN} and {Infinity}) cannot be
-coerced to {Float} and must raise a _runtime error_.
+coerced to {Float} and must raise an _execution error_.
 
 **Input Coercion**
 
@@ -579,7 +579,7 @@ that representation must be used to serialize this type.
 Fields returning the type {String} expect to encounter Unicode string values.
 
 GraphQL services may coerce non-string raw values to {String} when reasonable
-without losing information, otherwise they must raise a _runtime error_.
+without losing information, otherwise they must raise an _execution error_.
 Examples of this may include returning the string `"true"` for a boolean true
 value, or the string `"1"` for the integer `1`.
 
@@ -600,7 +600,7 @@ representation of the integers `1` and `0`.
 Fields returning the type {Boolean} expect to encounter boolean internal values.
 
 GraphQL services may coerce non-boolean raw values to {Boolean} when reasonable
-without losing information, otherwise they must raise a _runtime error_.
+without losing information, otherwise they must raise an _execution error_.
 Examples of this may include returning `true` for non-zero numbers.
 
 **Input Coercion**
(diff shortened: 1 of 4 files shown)


spec/Section 3 -- Type System.md
```diff
@@ -329,8 +329,8 @@ A GraphQL schema may describe that a field represents a list of another type;
 the `List` type is provided for this reason, and wraps another type.
 
 Similarly, the `Non-Null` type wraps another type, and denotes that the
-resulting value will never be {null} (and that a _field error_ cannot result in
-a {null} value).
+resulting value will never be {null} (and that a _runtime error_ cannot result
```
martinbonnin, line 332: YESSSSSSS!!! Quick feedback: what about "position error" or "execution error" instead of "runtime error"? To distinguish from a "request error".

Are there other types of error?

benjie: There are:
- parse error
- request error
- runtime error

We refer to "runtime" already in the spec, but you're right we use it for values provided for variables too which could result in a "request" error. Let's go with "execution error" - I think that's clear :raised_hands:

benjie: > If the `data` entry in the response is present (including if it is the value
> {null}), the `errors` entry must be present if and only if one or more
> _execution error_ was raised during execution.

Hmmm :thinking: 

# :shipit:

martinbonnin: I guess my only question regarding "execution error" is about subscriptions and https://github.com/graphql/graphql-spec/pull/894. 

Intuitively, I expect any error raised in `ExecuteRequest()` to be an "execution error". But looks like it might not always be the case?

martinbonnin: > Potentially we should rename ExecuteRequest to ProcessRequest or HandleRequest or something that doesn't use the word "execute". But that's out of scope for this PR.

+1 to that. Let me know if you want to do a follow up PR or else, I'm happy to look into this.

spec/Section 3 -- Type System.md
```diff
@@ -513,15 +513,15 @@ Fields returning the type {Int} expect to encounter 32-bit integer internal
...
 
 GraphQL services may coerce non-integer internal values to integers when
-reasonable without losing information, otherwise they must raise a _field
+reasonable without losing information, otherwise they must raise an _execution
 error_. Examples of this may include returning `1` for the floating-point number
 `1.0`, or returning `123` for the string `"123"`. In scenarios where coercion
-may lose data, raising a field error is more appropriate. For example, a
-floating-point number `1.2` should raise a field error instead of being
+may lose data, raising an execution error is more appropriate. For example, a
+floating-point number `1.2` should raise an execution error instead of being
```
spawnia, line 520:
```suggestion
may lose data, raising an _execution error_ is more appropriate. For example, a
floating-point number `1.2` should raise an _execution error_ instead of being
```

spec/Section 6 -- Execution.md
```diff
@@ -799,39 +808,45 @@ MergeSelectionSets(fields):
...
 resolution or coercion. While these errors should be reported in the response,
 they are "handled" by producing a partial response.
 
 Note: This is distinct from a _request error_ which results in a response with
 no data.
 
-If a field error is raised while resolving a field, it is handled as though the
-field returned {null}, and the error must be added to the {"errors"} list in the
-response.
+If an execution error is raised while resolving a field (either directly or
```
spawnia, line 824: Should we do underscores around `execution error` everywhere?

benjie: The general pattern is to do it at least once per headed section, but not to worry about it after that. I generally do it when I feel it adds clarity.

spec/Section 3 -- Type System.md
```diff
@@ -579,9 +579,9 @@ that representation must be used to serialize this type.
 Fields returning the type {String} expect to encounter Unicode string values.
 
 GraphQL services may coerce non-string raw values to {String} when reasonable
-without losing information, otherwise they must raise a _field error_. Examples
-of this may include returning the string `"true"` for a boolean true value, or
-the string `"1"` for the integer `1`.
+without losing information, otherwise they must raise an _execution error_.
+Examples of this may include returning the string `"true"` for a boolean true
+value, or the string `"1"` for the integer `1`.
```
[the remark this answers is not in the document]
benjie: That would be the subject for a different PR, the only change here (other than word wrapping) is changing `a _field` to `an _execution`.

I wish prettier would get back to me on [my semantic linebreak proposal](https://github.com/prettier/prettier/pull/16189).

robrichard (approved):
I like this, especially defining `response position` which improves clarity and will be helpful for Incremental Delivery

spec/Section 3 -- Type System.md
martinbonnin: https://github.com/graphql/graphql-spec/pull/1154

martinbonnin (approved):

BoD (approved):

mjmahone (approved):

spawnia (approved):

spec/Section 6 -- Execution.md
```diff
@@ -376,17 +376,23 @@ ExecuteSelectionSet(selectionSet, objectType, objectValue, variableValues):
 Note: {resultMap} is ordered by which fields appear first in the operation. This
 is explained in greater detail in the Field Collection section below.
 
-**Errors and Non-Null Fields**
+<a name="sec-Executing-Selection-Sets.Errors-and-Non-Null-Fields">
+  <!-- This link exists for legacy hyperlink support -->
+</a>
```
leebyron, line 381: nice touch

spec/Section 6 -- Execution.md
```diff
@@ -620,6 +626,9 @@ the type system to have a specific input type.
 At each argument position in an operation may be a literal {Value}, or a
 {Variable} to be provided at runtime.
 
+Any _request error_ raised during {CoerceArgumentValues()} should be treated
+instead as an _execution error_.
+
```
leebyron, line 631: this is fine for now. Ideally anywhere we raise a request error is definitely a request error.

Is the case this is covering against per-type input coercion?

spec/Section 6 -- Execution.md
```diff
@@ -798,39 +807,45 @@ MergeSelectionSets(fields):
...
 
-### Handling Field Errors
+<a name="sec-Handling-Field-Errors">
+  <!-- This link exists for legacy hyperlink support -->
+</a>
+
+### Handling Execution Errors
 
-A _field error_ is an error raised from a particular field during value
+An _execution error_ is an error raised from a particular field during value
```
leebyron, line 816: raised from a `response position`?

leebyron (commented):
I really like this change

leebyron (approved):

benjie:
Really nice edits, thanks Lee! I especially like the expansion and lifting of response position and think @robrichard will appreciate that too!

The one minor issue is moving the anchor tags below the title means the header will now not be visible when the (legacy) link is clicked. Not a huge deal, and it makes maintaining the spec easier. If we do this enough, might be worth adding a feature to spec-md, maybe something like:

```md
### Handling Execution Errors

<!-- !! Previously: Handling Field Errors -->
```

PR #1161, opened 2025-04-25
Change "grouped field set" to be a map of ordered sets

benjie:
Extension to #999 

Currently:

```
type GroupedFieldSet = Map<ResponseName, Array<FieldSelection{ResponseName}>>
```

I propose that we defined "field set" to be an ordered set (rather than list) of fields that all share the same response name. This makes a lot of the spec clearer, e.g. it's much clearer why `ExecuteField()` should receive a "field set" (since they all share the same response name) rather than simply receiving "fields".

Essentially I propose:

```
// Constraint: every FieldSelection in a FieldSet must have the same ResponseName
type FieldSet{ResponseName} = OrderedSet<FieldSelection{ResponseName}>

type GroupedFieldSet = Map<ResponseName, FieldSet{ResponseName}>
```

This also makes the term "grouped field set" a lot clearer.


yaacovCR:
> 2\. Note: `visitedFragments` doesn't prevent this because it doesn't work across sibling selections, only nested selections

I believe the current specification has `visitedFragments` working across sibling selections, but that there is a proposal to limit the use of `visitedFragments` to work only for nested selections, to prevent other problems:
 - https://github.com/graphql/graphql-spec/pull/1045

For what it's worth, I support #1045 as well as this change. :)

benjie:
This was merged into #1039 and since merged into the spec.

PR #1164, opened 2025-04-30
Allow `schema @behavior(onError: ...)` without explicit root operations

benjie:
This PR is rebased on #1163. It comprises of two commits. The first is from:

- #1166

The second includes the adjustments necessary to relate to:

- #1163 

Please see #1166 for details.

### DIFF
--- a/spec/Appendix B -- Grammar Summary.md
+++ b/spec/Appendix B -- Grammar Summary.md
@@ -258,8 +258,11 @@ TypeSystemExtension :
 - SchemaExtension
 - TypeExtension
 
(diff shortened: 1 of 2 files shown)


martinbonnin (approved):
+1 to this, would be quite helpful

benjie:
I've pulled the core of this out and raised it as a clean PR against `main`:

- #1166

benjie:
Lee was very hesitant to allow a "naked" schema, and feels that it would be ambiguous: does it mean that no operations are defined, or only the defaults? Since this is purely aesthetic and to try and match the behavior of the other types in the schema, I'm going to close it for now.

PR #1165, opened 2025-04-30
Add Transitional Non-Null appendix (`@noPropagate` directive)

benjie:
This is essentially solution 8 to the [Semantic Nullability RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/SemanticNullability.md):

- Enables **semantic nullability** to be reflected in schemas without breaking legacy behavior.
- Facilitates **incremental adoption** of modern error handling without requiring disruptive changes.
- Requires **minimal spec impact** and is fully optional for implementations.
- Reflects **transitional nature** of this change in behavior

I've based it on:

- #1163 

since, like all solutions to the semantic nullability problem[^1], it is designed to enable clients with error propagation disabled to leverage the true nullability of the underlying data without breaking legacy clients. The approach could be rebuilt atop an alternative method of toggling error propagation, for example a directive-based approach.

[^1]: Except solution 5

---

This PR introduces an appendix to the GraphQL specification defining an optional solution to the semantic nullability problem using the following key mechanisms:

- **`@noPropagate`[^2] directive** — allows schema authors to annotate `Non-Null` return types as *transitional*, suppressing propagation but preserving runtime error generation.
- **Transitional Non-Null semantics** — errors at these positions behave like nullable fields in terms of (no!) propagation but like non-nullable fields in value completion (error on null).
- **New `__Field.noPropagateLevels: [Int!]` field** — exposes transitional status to modern clients.
- **Transitional non-null hidden from legacy clients** — tooling using the legacy `PROPAGATE` _error behavior_ will get results from `__Field.type` that unwrap transitional non-null types.[^3]

[^2]: This is essentially the same as the `@semanticNonNull` directive, but more strictly defined and reflected through introspection.

This solution attempts to address all of the feedback on previous solutions to this problem, whilst being explicitly transitional. It:

- Is optional: explicitly only for schemas supporting legacy clients
- Requires no changes to the main spec text
- Introduces no new syntax
- True to its name: an error here will not propagate (`@noPropagate`), regardless of whether error propagation is enabled or disabled.
- Maintains introspection results for existing (deployed) clients and tooling
- Maintains error boundaries for existing (deployed) clients
- Allows all new schemas and new fields to use `!` (non-null) directly for semantically non-null positions
- Allows existing fields to use `!` (non-null) for error handling clients without breaking legacy clients by adding the `@noPropagate` directive[^4]
- Can be adopted gradually, field-by-field, or en masse by applying `@noPropagate` to all nullable positions.
- Can be removed from each field the moment no legacy clients query it

[^3]: This may be controversial, but I truly think it's the right decision. All new tooling (and all new clients!) should use `onError: ABORT` or `onError: NO_PROPAGATE`, and thus will see the true introspection. Existing tooling doesn't know about `onError` and so should not see these "transitional" non-null types.
[^4]: And if you forget to add it, adding it later is only a potentially breaking change for any new versions of legacy clients deployed since the change; error-handling clients (`NO_PROPAGATE` or `ABORT`) are unimpacted.

### DIFF
--- a/spec/Appendix C -- Transitional Non-Null.md
+++ b/spec/Appendix C -- Transitional Non-Null.md
@@ -8,28 +8,23 @@ gradual migration toward semantic nullability while preserving compatibility.
 
 ## Overview
 
+With the introduction of _error behavior_, clients can take responsibility for
+handling of _execution error_: correlating {"errors"} in the result with `null`
+values inside {"data"} and thereby removing the ambiguity that error propagation
+originally set out to solve. If all clients adopt this approach then schema
+designers can, and should, reflect true nullability in the schema, marking
+fields as `Non-Null` based on their data semantics without regard to whether or
+not they might error.
 
-The introduction of _error behavior_ to this specification allows clients to
-take responsibility for error handling, no longer having the schema perform
-error propagation and destroying potentially useful response data in the
-process. With this move towards clients handling errors, designers of new
-schemas (or new fields in existing schemas) no longer need to factor whether or
-not a field is likely to error into its nullability; designers can mark a
-_semantically_ non-nullable _response position_ (a place where {null} is not a
-semantically valid value for the data) as `Non-Null`, writing a {null} there on
-error in the knowledge that the client now takes responsibility for handling
-errors and preventing these placeholder {null} values from being read.
+However, legacy clients may not perform this correlation. Introducing `Non-Null`
+in such cases could cause errors to propagate further, potentially turning a
+previously handled error in a single field into a full-screen error in the
+application.
 
-However, for schema designers that need to support legacy clients that do not
-exhibit these error handling properties, marking semantically non-nullable
-response positions as `Non-Null` would mean that more of the response would be
-destroyed for these clients on error, potentially turning local widget errors
-into full screen errors.
-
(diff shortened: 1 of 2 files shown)


spec/Appendix C -- Transitional Non-Null.md
```diff
@@ -0,0 +1,129 @@
...
+
+To support a smooth transition, this appendix introduces the `@noPropagate`
+directive and the concept of _transitional_ Non-Null types. These wrappers raise
+errors like regular `Non-Null` types, but suppress propagation and appear
+nullable in introspection when using the legacy {"PROPAGATE"} _error behavior_.
+
+## The @noPropagate Directive
+
+```graphql
+directive @noPropagate(levels: [Int!]! = [0]) on FIELD_DEFINITION
```
fotoetienne, line 32: Some alternative names:

- @semantic - This field type is semantically non-null. It will never have a null value except in case of error, but we since errors happen it may have a null value
- @soft - This is a "soft" non-null in that it is not enforced by propagation

martinbonnin: * `@transitionalErrorBoundary`
* `@transitionallyNullable`
* `@localError`

??

PR #1167, opened 2025-05-19
Do not exclude schema keyword if schema has description

benjie:
This is a bugfix; previously we said that the `schema` keyword can/should be omitted if the root operation type names line up, but we neglected to think about the schema description. This fixes that.

Trivial fix @graphql/tsc

(Discussed as part of #1166, but since there's pushback to the naked schema keyword I'm separating this out into its own standalone fix.)

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -217,13 +217,14 @@ type MyMutationRootType {
 {"Subscription"} respectively.
 
 The type system definition language can omit the schema definition when each
-_root operation type_ uses its respective _default root type name_ and no other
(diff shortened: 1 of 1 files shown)


michaelstaib (approved):

mjmahone (approved):

enisdenjo (approved):

PascalSenn (approved):

Muaytie666 (approved):

benjie:
GraphQL.js test passes: https://github.com/graphql/graphql-js/pull/4441

leebyron (approved):

PR #1168, opened 2025-06-02
Add 'Assert' consistency to algorithm format check

benjie:
We have a mixture of `- Assert: blah` and `- Assert blah`. We should standardize on one pattern; this adds it to the spec format checks. (It also fixes some minor bugs in the script itself.)


martinbonnin (approved):
Nit: I like it better without the `:` since we don't have `Return: foo` but `Return foo` and more generally I find that the spec are usually closer to natural language but consistency > everything so I'm fine with everything and support enforcing this in the format check 👍

PascalSenn (approved):

robrichard (approved):

leebyron:
love this, thanks for adding the check

PR #1172, opened 2025-07-01
Pull copyright and license into repo-level LICENSE.md

leebyron:
Ensures the repo has a top level LICENSE.md file without concern of skew from the spec text by way of an import reference.

This also moves the License and conformance section into an appendix, and does some editorial to the spec introduction including adding links about contribution. Hopefully this makes it easier to read since the ToC is much more likely to be above the fold.

### DIFF
--- a/spec/GraphQL.md
+++ b/spec/GraphQL.md
@@ -11,7 +11,7 @@ client-server applications.
 A conforming implementation of GraphQL must fulfill all normative requirements
 described in this specification (see [Conformance](#sec-Appendix-Conformance)).
 The GraphQL specification is provided under the OWFa 1.0 license (see
-[Licensing](#sec-Appendix-Copyright-and-Licensing)).
+[Copyright and Licensing](#sec-Appendix-Copyright-and-Licensing)).
 
 GraphQL was originally created in 2012 and the development of this open standard
 started in 2015. It is a deliverable of the


spec/GraphQL.md
```diff
@@ -4,125 +4,37 @@ _Current Working Draft_
...
+client-server applications.
+
+GraphQL was originally created in 2012 and the development of this open standard
+started in 2015. This specification was licensed under OWFa 1.0 in 2017 (see
+[Licensing](#sec-Copyright-and-Licensing)). The
+[GraphQL Foundation](https://graphql.org/foundation/) was formed in 2019 as a
+neutral focal point for organizations who support the GraphQL ecosystem, and the
 [GraphQL Specification Project](https://graphql.org/community/) was established
 also in 2019 as the Joint Development Foundation Projects, LLC, GraphQL Series.
-
```
benjie, line 16: I would restore this paragraph break.

spec/GraphQL.md
```diff
@@ -4,125 +4,37 @@ _Current Working Draft_
...
 ecosystem.
 
-The GraphQL Specification Project has evolved and may continue to evolve in
-future editions of this specification. Previous editions of the GraphQL
-specification can be found at permalinks that match their
+This specification is developed on GitHub at
+[graphql/graphql-spec](https://github.com/graphql/graphql-spec/). Contributions
+are managed by the
+[GraphQL Working Group](https://github.com/graphql/graphql-wg), hosted by the
+[GraphQL TSC](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md).
```
benjie, line 27: We shouldn't use an abbreviation the first time the term is introduced:

```suggestion
[GraphQL Technical Steering Committee](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md).
```

(Feel free to append ` (TSC)` if we want to refer to it again, but I don't think we do.)

spec/Appendix D -- Conformance & Licensing.md
```diff
@@ -0,0 +1,64 @@
+# D. Appendix: Conformance & Licensing
```
benjie, line 1: Does it make sense to have these be in the same appendix? I feel like having an extra focussed appendix is not too costly, and "conformance" and "licensing" feel somewhat separate to me.

spec/GraphQL.md
```diff
@@ -4,125 +4,37 @@ _Current Working Draft_
...
+To learn more see the
+[contribution guide](https://github.com/graphql/graphql-spec/blob/main/CONTRIBUTING.md).
+
+GraphQL has evolved and may continue to evolve in future editions of this
+specification. Previous editions of the GraphQL specification can be found at
+permalinks that match their
 [release tag](https://github.com/graphql/graphql-spec/releases). The latest
 working draft release can be found at
 [https://spec.graphql.org/draft](https://spec.graphql.org/draft).
```
benjie, line 37: I think we should keep a couple of the main call-outs to the appendices here; especially where reading them up-front may be beneficial to some readers. Perhaps:

```suggestion

See [Appendix A](#sec-Appendix-Notation-Conventions) for more details about the
definition of algorithms, data collections, and other notational conventions
used in this document.

A conforming implementation of GraphQL must fulfil all normative requirements.
Conformance requirements expressed as algorithms and data collections can be
fulfilled by an implementation of this specification in any way as long as the
perceived result is equivalent. See [Appendix D](#sec-Appendix-Conformance) for
details.
```

(Aside: should that "algorithms **and** data collections" be "algorithms **or** data collections"?)

benjie (approved):
I like this reduction in the overhead of getting into the meat of the specification :+1: 

Here's a couple changes I'd like to see, but am not overly attached to.

PR #1174, opened 2025-07-02
Define 'execution' as in 'before execution begins'

leebyron:
Reintroduces #894 atop latest main

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -328,7 +328,7 @@ GraphQL supports two abstract types: interfaces and unions.
 An `Interface` defines a list of fields; `Object` types and other Interface
 types which implement this Interface are guaranteed to implement those fields.
 Whenever a field claims it will return an Interface type, it will return a valid
-implementing Object type during execution.
+implementing Object type during _execution_.
 
 A `Union` defines a list of possible types; similar to interfaces, whenever the
 type system claims a union will be returned, one of the possible types will be
@@ -505,7 +505,7 @@ information on the serialization of scalars in common JSON and other formats.
 If a GraphQL service expects a scalar type as input to an argument, coercion is
 observable and the rules must be well defined. If an input value does not match
 a coercion rule, a _request error_ must be raised (input values are validated
-before execution begins).
+before _execution_ begins).
 
 GraphQL has different constant literals to represent integer and floating-point
 input values, and coercion rules may apply differently depending on which type
@@ -810,7 +810,7 @@ And will yield the subset of each object type queried:
 **Field Ordering**
 
 When querying an Object, the resulting mapping of fields are conceptually
-ordered in the same order in which they were encountered during execution,
+ordered in the same order in which they were encountered during _execution_,
 excluding fragments for which the type does not apply and fields or fragments
 that are skipped via `@skip` or `@include` directives. This ordering is
 correctly produced when using the {CollectFields()} algorithm.
@@ -2058,8 +2058,8 @@ directive @example on
 
 Directives can also be used to annotate the type system definition language as
 well, which can be a useful tool for supplying additional metadata in order to
-generate GraphQL execution services, produce client generated runtime code, or
-many other useful extensions of the GraphQL semantics.
+generate GraphQL services, produce client generated runtime code, or many other
+useful extensions of the GraphQL semantics.
 
 In this example, the directive `@example` annotates field and argument
 definitions:
@@ -2122,7 +2122,7 @@ directive @skip(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
 ```
 
 The `@skip` _built-in directive_ may be provided for fields, fragment spreads,
-and inline fragments, and allows for conditional exclusion during execution as
+and inline fragments, and allows for conditional exclusion during _execution_ as
 described by the `if` argument.
 
 In this example `experimentalField` will only be queried if the variable
@@ -2142,7 +2142,7 @@ directive @include(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
 
 The `@include` _built-in directive_ may be provided for fields, fragment
 spreads, and inline fragments, and allows for conditional inclusion during
-execution as described by the `if` argument.
+_execution_ as described by the `if` argument.
 
 In this example `experimentalField` will only be queried if the variable
 `$someTest` has the value `true`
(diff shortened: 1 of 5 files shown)


spec/Section 6 -- Execution.md
```diff
@@ -39,27 +39,44 @@ and have no effect on the observable execution, validation, or response of a
 GraphQL document. Descriptions and comments on executable documents MAY be used
 for non-observable purposes, such as logging and other developer tools.
 
-## Executing Requests
+## Processing Requests
```
leebyron, line 42: I think the intermixing of "execute" and "process" is ultimately more confusing than helpful, despite it being technically more accurate. I think it's the result of the topline section being "Execution" and then within we define execution as a sub-step of "Processing". It's a much larger change to retitle the entire section.

I wonder if instead we could define phases of execution?

spec/Section 3 -- Type System.md
```diff
@@ -2058,8 +2058,8 @@ directive @example on
 
 Directives can also be used to annotate the type system definition language as
 well, which can be a useful tool for supplying additional metadata in order to
-generate GraphQL execution services, produce client generated runtime code, or
-many other useful extensions of the GraphQL semantics.
+generate GraphQL services, produce client generated runtime code, or many other
```
leebyron, line 2061: (I added this change as I grepped through for execution references, everywhere else we just refer to a "GraphQL service" not a "GraphQL execution service")

spec/Section 6 -- Execution.md
```diff
@@ -18,9 +18,9 @@ A GraphQL service generates a response from a request via execution.
 - {extensions} (optional): A map reserved for implementation-specific additional
   information.
 
-Given this information, the result of {ExecuteRequest(schema, document,
-operationName, variableValues, initialValue)} produces the response, to be
-formatted according to the Response section below.
+Given this information, the result of {Request(schema, document, operationName,
```
leebyron, line 21: Lets think of how we can keep the algorithm names with action verbs. Ideally this has an "execute" or "process" or something here.

Considering the prose above describes this as "request via execution" then "execute request" does seem like what a reader would expect

spec/Section 6 -- Execution.md
```diff
@@ -74,27 +91,28 @@ GetOperation(document, operationName):
 
 ### Validating Requests
 
-As explained in the Validation section, only requests which pass all validation
-rules should be executed. If validation errors are known, they should be
-reported in the list of "errors" in the response and the request must fail
-without execution.
+As explained in the Validation section, only operations from documents which
```
leebyron, line 94: nice cleanup here

spec/Section 6 -- Execution.md
```diff
@@ -176,12 +198,13 @@ ExecuteMutation(mutation, schema, variableValues, initialValue):
...
-executing the operation for each new event on an underlying _source stream_.
+executing the operation’s _root selection set_ for each new event on an
+underlying _source stream_.
 
-Executing a subscription operation creates a persistent function on the service
+Performing a subscription operation creates a persistent function on the service
 that maps an underlying _source stream_ to a returned _response stream_.
 
-Subscribe(subscription, schema, variableValues, initialValue):
+Subscription(subscription, schema, variableValues, initialValue):
```
leebyron, line 207: This one is admittedly the weirdest. Regardless of how we net-out on naming, we should rethink the terms used for the phases of subscription

spec/Section 6 -- Execution.md
```diff
@@ -39,27 +39,44 @@ and have no effect on the observable execution, validation, or response of a
...
+concerns such as determining the operation and coercing the inputs, before
+passing the request on to the relevant algorithm for the operation's type which
+then performs any other necessary preliminary steps (for example establishing
+the source event stream for subscription operations) and then initiates
+_execution_.
+
+Note: An error raised before _execution_ begins will typically be a _request
+error_, and once _execution_ begins will typically be an _execution error_.
+
+:: We define _execution_ as the process of executing the operation's _root
```
leebyron, line 64: I think here is where we could define "during execution" with something more specific to avoid the overloaded term.

These names might not be ideal, but it seems like we have two phases to execution. 

- First there is a pre-execution preparation step where we coerce variables and ensure we have a valid type and object to begin execution with

- Then there is the actual collection and execution of fields (though even here it's a bit arbitrary that "execution" starts as we begin collecting fields not when we start the first `ExecuteCollectedFields`. Perhaps we could call this phase "field execution" or "selection execution"?

I appreciate what we're trying to do here is more clearly define what scenarios lead to an error result with no data entry vs one with a partial data entry. We started with "field error" -- which while unclear for other reasons -- at least evaded this issue. They were clearly errors which occurred at a per-field level. Maybe we just need to do a half-step back to that?

leebyron (commented):
Hey @benjie - some thoughts on this PR for you. I'd love to do some brainstorming on terminology here to find a happy medium

benjie:
**TL;DR**: we already have a section "Executing Operations"; let's simply rename "execution" to "operation execution" where appropriate - WDYT?

---

@leebyron I've merged these changes back into #894 [^1] and have then[^5]:

1. wound back the changes to algorithm names[^2]
2. renamed "execution" to "operation execution" throughout
3. used the existing "Executing Operations" section as the point at which "operation execution begins" [^3]
4. moved establishing the source event stream into `ExecuteRequest()`[^4]



All in all I think this makes the spec text a lot clearer without having to rename any algorithms or introduce different verbs. The only other change that we might want to consider is renaming `Subscribe()` to `ExecuteSubscription()` for consistency; but honestly I'm fine how it is.

Please let me know if you're happy with this direction; if so I'll go over it more carefully to ensure it's consistent (in particular I think I might need to update the text of the subscriptions section).

[^1]: commit [`7bafeaf` (#894)](https://github.com/graphql/graphql-spec/pull/894/commits/7bafeaf25423a68a687bbba1c461d76bcbd33e26) has no diff against this PR
[^2]: also removing "perform" and "process" verbs
[^3]: which feels really natural, right?
[^4]: since it happens "before execution begins"
[^5]: you can see my changes since this PR here: https://github.com/graphql/graphql-spec/pull/894/files/7bafeaf25423a68a687bbba1c461d76bcbd33e26..94241a78d020fb0bdeea483a0f50d54beff890ba - but personally I'd just read the <a href="https://github.com/graphql/graphql-spec/pull/894/files">full diff</a> as it's simpler

benjie:
Haha you did all the work; I just did `git merge leesbranch` and then `git checkout leesbranch .` to resolve all the conflicts 🤣

benjie:
(And to reopen I had to restore the branch it was based on, then change the PR target to main.)

PR #1178, opened 2025-07-03
Remove CoerceArgumentValues() hasValue

leebyron:
An alternative to #1056 which fully removes the `hasValue` variable in favor of hopefully more clear conditions

The original intent of tracking `hasValue` was to avoid duplication of logic in places where once we've gotten the "value" from whatever source (value, variable, etc) then the rest of the steps are the same. However that's proven to be fragile and hard to read, since values from different sources are in fact handled differently.

This instead removes the concept of `hasValue` by instead just using more specific conditions like "If __ has an entry named ___, then let ___ be its value"

For future -- I'd love to overhaul the entire "input coercion" to be strict algorithms instead of prose, which seems particularly important for being crisp about what to do with variables.


leebyron:
Closing in favor of #1179

PR #1183, opened 2025-07-10
Fix "response position" definition; clarify sibling errors on propagation

benjie:
> If a _response position_ resolves to {null} because of an execution error which
has already been added to the {"errors"} list in the _execution result_, the
{"errors"} list must not be further affected. That is, only one error should be
added to the errors list per _response position_.

The _response position_ mentioned at the beginning of this paragraph is different to the response position from which the error originated; therefore "only one error should be added to the errors list per _response position_" is kind of moot. Actually what we mean here is that the _response path_ of the error (i.e. the "path" entry in the error object) should be unique.

This edit clarifies that it's the "path" of the error that matters (and should be unique), not the position that re-raised it.


benjie:
> I think this depends on if c is nullable or not (with b propagating the error to a).

That's right, yes - exactly as you describe.

> Which means then that not only is [a,c] (non-propagated) is not added to errors, it must also be removed from errors if it already exists?

That was how I understood it to work - my interpretation was that it was more of a functional approach where data/errors were "passed down" from nullable positions but non-null positions threw - thus sibling errors would be "overridden" by another field throwing. But when I checked this against the reference implementation I realised I was mistaken and the spec text is actually accurate to the behavior here:

https://github.com/graphql/graphql-js/blob/8dc295551f9a64276442ef850943794db6bcaa4a/src/execution/execute.ts#L597-L612

<details>
<summary>GraphQL.js reproduction</summary>
```ts
import {
  graphql,
  GraphQLInt,
  GraphQLNonNull,
  GraphQLObjectType,
  GraphQLSchema,
} from 'graphql';

const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));

const mkField = (counter: number) => ({
  async resolve() {
    console.log(`Triggered ${counter}`);
    await sleep(Math.random() * 10);
    console.log(`Resolved ${counter}`);
    throw new Error(`E${counter}`);
  },
});

const A = new GraphQLObjectType({
  name: 'A',
  fields: {
    b: {
      async resolve() {
        await sleep(10);
        throw new Error(`b`);
      },
    },
    c: {
      type: GraphQLInt,
      async resolve() {
        await sleep(0);
        throw new Error(`c`);
      },
    },
  },
});

  name: 'Query',
  fields: {
    a: {
      type: A,
      resolve() {
        return {};
      },
    },
  },
});
  query: Query,
});

  schema,
  source: `{ a { b c } }`,
});

console.dir(result);

```
</details>

If `c` is nullable and throws first then the error will be added to the errors list. Later when `b` fails it is _also_ added to the errors list.

However if `c` is non-nullable and throws first, then `b` will be "cancelled" and its error will not be added to the list. (It's not actually cancelled, we just ignore the result. In future we'll use AbortSignal to make it possible to cancel.)

And if `b` (non-null) throws before `c` then `c` will not be added to the list whether it's nullable or not.

Interestingly if you implement GraphQL according to my previous interpretation it's still spec compliant because you as the consumer do not know whether b would run before or after c - the only place where this interpretation would break the spec would be non-nullable mutation fields:

<details>
<summary>GraphQL.js example of non-nullable second mutation field persisting error from first mutation</summary>

```ts
import {
  graphql,
  GraphQLInt,
  GraphQLNonNull,
  GraphQLObjectType,
  GraphQLSchema,
} from './npmDist/index.mjs';

const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));

  name: 'Query',
  fields: {
    a: {
      type: GraphQLInt,
    },
  },
});
const Mutation = new GraphQLObjectType({
  name: 'Mutation',
  fields: {
    a: {
      type: GraphQLInt,
      async resolve() {
        await sleep(10);
        throw new Error(`a`);
      },
    },
    b: {
      async resolve() {
        await sleep(10);
        throw new Error(`b`);
      },
    },
    c: {
      type: GraphQLInt,
      async resolve() {
        await sleep(0);
        throw new Error(`c`);
      },
    },
  },
});
  query: Query,
  mutation: Mutation,
});

  schema,
  source: `mutation { a  b c  }`,
});

console.dir(result);
```

</details>


I'll update the edits to reflect this.

benjie:
Plan from today's WG:

1. open a new PR that clarifies "response position" without renaming to "execution position" (which could be ambiguous w.r.t. visiting fragments?)
2. rebase this change on that PR, so this becomes just the one sentence change
3. see how hard this would be to implement in GraphQL.js and a few other implementations
4. if hard, reject
5. in `graphql-toe`, throw the _last_ error that matches the path (just reverse the array?) or throw an aggregate error

yaacovCR:
https://github.com/graphql/graphql-js/pull/4458 not sure if this is the implementation you want to test, I thought it’s the implementation for: https://github.com/graphql/graphql-spec/pull/1184

benjie:
Editorial change for definition of response position extracted to here:

- https://github.com/graphql/graphql-spec/pull/1238

martinbonnin (approved):
```
If a _response position_ resolves to {null} because of an execution error which
has already been added to the {"errors"} list in the _execution result_, the
{"errors"} list must not be further affected. That is, only one error should be
added to the errors list per _response position_.
```

FWIW, I don't necessarily read this as "the response position at the beginning of the paragraph and at the end must be the same". So the curernt text was ok-ish to me.

Nevertheless, the new text is more explicit so all good 👍

PascalSenn (approved):

robrichard (approved):

leebyron (approved):
Nice. I like that this cleans up things since the response since the _response position_ might not exist in the response thanks to bubbling.

PR #1192, opened 2025-09-12
Fix `includeDeprecated` arg type in "Appendix D"

tyranron:
Follows #1142

In #1142 the `includeDeprecated` arg's type was changed to non-`Null`, however its [definition in the "Appendix D"](https://spec.graphql.org/September2025/#sec-Appendix-Type-System-Definitions) wasn't changed, leading to confusions and misinterpretations.
Screenshot 2025-09-12 at 11 19 36

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "build": "./build.sh",
     "test:build": "spec-md --metadata spec/metadata.json spec/GraphQL.md > /dev/null",
     "watch": "nodemon -e json,md --exec \"npm run build\"",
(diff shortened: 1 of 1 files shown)


benjie:
I believe Appendix D is produced by graphql.js, so this suggests that library has not caught up with this spec change. The PR should be directed there and the  reflected here.

martinbonnin:
At the time of writing the Appendix D. PR, [I didn't find a version of graphql-js that matched exactly the spec draft](https://github.com/graphql/graphql-spec/pull/1037#discussion_r2129266770) so I had to tweak this manually. This must have been lost in translation somewhere. Thanks for fixing it @tyranron 🙏

PR #1193, opened 2025-09-12
Update description of Fragments to emphasize evolving data needs

janettec:
## Why these changes?

**Fragments are not for reuse.**

The current language in [section 2.8 on Fragments](https://spec.graphql.org/October2021/#sec-Language.Fragments), namely the sentence...

> Fragments allow for the reuse of common repeated selections of fields, reducing duplicated text in the document.

...encourages using fragments to deduplicate common selections, even if those selections represent independent data needs. 

For example, let's consider these two functions

```swift
// This function formats a post for display in a feed
func formatPostForFeed(post: PostDisplayFragment) -> String {
 let authorName = post.author.name
 let contentText = post.content.text

 return "\(authorName) posted: \(contentText)"
}

// This function formats a post for notification purposes
func formatPostForNotification(post: PostDisplayFragment) -> String {
 let authorName = post.author.name
 let contentText = post.content.text

 return "New post from \(authorName): \(contentText)"
}
```

Given these two functions currently both use `authorName` and `contentText`, the current language in the spec encourages one to write a fragment

```graphql 
fragment PostDisplayFragment on Post {
 author
 content
}
```

If `formatPostForFeed` now needs `timestamp`, we will add naturally add `timestamp` to `PostDisplayFragment` 
```swift
// This function formats a post for display in a feed
func formatPostForFeed(post: PostDisplayFragment) -> String {
 let authorName = post.author.name
 let contentText = post.content.text
 let timestamp = post.createdAt.formatted // Added

 return "\(authorName) posted: \(contentText)\nPosted \(timestamp)"
}
```

```graphql 
fragment PostDisplayFragment on Post {
 author
 content
 timestamp # Added
}
```
If we have the following queries

```graphql
query NotificationQuery {
 ...PostDisplayFragment # Over-fetching timestamp
}

query FeedQuery {
 ...PostDisplayFragment
}
```
notice how `NotificationQuery` is now over-fetching timestamp! 

The key observation is that `formatPostForFeed` and `formatPostForNotification` are two independent functions, so by having them both rely on a single fragment to express their data needs, we are creating a dependency where one should not exist (because that dependency does not exist in the product logic itself).

## What are the proposed changes?

Updated:
1. The description for why one might use fragments
2. The first GraphQL example in section 2.8

The goal is to emphasize that fragments support evolving data needs (as opposed to recommending people identify common selections that are currently in an executable document).

Open to any and all feedback on the motivation for the change and how it's communicated via changes in the spec language!

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -587,13 +587,9 @@ FragmentName : Name but not `on`
 
 Fragments are the primary unit of composition in GraphQL.
 
-Fragments allow for the reuse of common repeated selections of fields, reducing
-duplicated text in the document. Inline Fragments can be used directly within a
-selection to condition upon a type condition when querying against an interface
-or union.
+Fragments allow for the definition of selection sets that are colocated with the logic that requires those selections, making it easier to add and remove selections as needed.
 
-For example, if we wanted to fetch some common information about mutual friends
-as well as friends of some user:
+For example, if we have some `friendProfile` logic that requires `id`, `name`, and `profilePic`, and we want to apply that logic to the mutual friends as well as friends of some user:
 
 ```graphql example
 query noFragments {
@@ -612,28 +608,32 @@ query noFragments {
 }
 ```
 
-The repeated fields could be extracted into a fragment and composed by a parent
-fragment or operation.
+The fields required of `friendProfile` can be extracted into a fragment and composed 
+by a parent fragment or operation.
 
 ```graphql example
 query withFragments {
   user(id: 4) {
     friends(first: 10) {
-      ...friendFields
+      ...friendProfileFragment
     }
     mutualFriends(first: 10) {
-      ...friendFields
+      ...friendProfileFragment
     }
   }
 }
-
-"Common fields for a user's friends."
-fragment friendFields on User {
+```
+```graphql example
+"Fields required to display a friend's profile"
+fragment friendProfileFragment on User {
   id
   name
   profilePic(size: 50)
 }
 ```
+If `friendProfile` no longer needs `name`, the `name` field can be removed from 
+`friendProfileFragment` and it will no longer be fetched in both locations the 
+fragment is consumed.
 
 Fragments are consumed by using the spread operator (`...`). All fields selected
 by the fragment will be added to the field selection at the same level as the
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -587,12 +587,11 @@ FragmentName : Name but not `on`
 
 Fragments are the primary unit of composition in GraphQL.
 
-Fragments allow for the definition of selection sets that are colocated with the 
-logic that requires those selections, making it easier to add and remove selections 
-as needed.
+Fragments allow for the definition of modular selection sets that make it easier
+to add and remove selections as needed.
 
-For example, if we have some `friendProfile` logic that requires `id`, `name`, 
-and `profilePic`, and we want to apply that logic to the mutual friends as well 
+For example, if we have some `friendProfile` logic that requires `id`, `name`,
+and `profilePic`, and we want to apply that logic to the mutual friends as well
 as friends of some user:
 
 ```graphql example
@@ -612,31 +611,31 @@ query noFragments {
 }
 ```
 
-The fields required by `friendProfile` can be extracted into a fragment and 
+The fields required by `friendProfile` can be extracted into a fragment and
 composed by a parent fragment or operation.
 
 ```graphql example
 query withFragments {
   user(id: 4) {
     friends(first: 10) {
-      ...friendProfileFragment
+      ...friendProfile
     }
     mutualFriends(first: 10) {
-      ...friendProfileFragment
+      ...friendProfile
     }
   }
 }
 ```
 ```graphql example
 "Fields required to display a friend's profile"
-fragment friendProfileFragment on User {
+fragment friendProfile on User {
   id
   name
   profilePic(size: 50)
 }
 ```
-If `friendProfile` no longer needs `name`, the `name` field can be removed from 
-`friendProfileFragment` and it will no longer be fetched in both locations the 
+If `friendProfile` no longer needs `name`, the `name` field can be removed from
+the `friendProfile` fragment and it will no longer be fetched in both locations the
 fragment is consumed.
 
 Fragments are consumed by using the spread operator (`...`). All fields selected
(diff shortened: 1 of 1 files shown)


spec/Section 2 -- Language.md
```diff
@@ -587,13 +587,13 @@ FragmentName : Name but not `on`
 
 Fragments are the primary unit of composition in GraphQL.
 
-Fragments allow for the reuse of common repeated selections of fields, reducing
-duplicated text in the document. Inline Fragments can be used directly within a
-selection to condition upon a type condition when querying against an interface
-or union.
+Fragments allow for the definition of selection sets that are colocated with the 
+logic that requires those selections, making it easier to add and remove selections 
+as needed.
```
martinbonnin, line 592: Tentative suggestion that keeps the concept of "colocation" out of the spec for the time being.

```suggestion
Fragments allow grouping selections that belong together, making it easier to scope 
data accesses, add and remove selections as needed.
```

spec/Section 2 -- Language.md
```diff
@@ -612,28 +612,32 @@ query noFragments {
...
-fragment or operation.
+The fields required by `friendProfile` can be extracted into a fragment and 
+composed by a parent fragment or operation.
 
 ```graphql example
 query withFragments {
   user(id: 4) {
     friends(first: 10) {
-      ...friendFields
+      ...friendProfileFragment
```
martinbonnin, line 622: I personally like `friendProfile`, `Fragment` is making me type more for no obvious reason as I know it's a fragment from the context.

martinbonnin (approved):
A couple comments but otherwise very supportive of updating that language!

spec/Section 2 -- Language.md
```diff
@@ -612,29 +611,35 @@ query noFragments {
 }
 ```
 
-The repeated fields could be extracted into a fragment and composed by a parent
-fragment or operation.
+The fields required by `friendProfile` can be extracted into a fragment and
```
benjie, line 614:
```suggestion
The fields required by the `friendProfile` logic can be extracted into a fragment and
```

benjie (changes requested):
Calling the logic `friendProfile` and the fragment `friendProfile` is confusing. Can we differentiate the logic - e.g. just call it "profile rendering logic" (prose) rather than using code? (Not sure if this is the best fix.)

martinbonnin (approved):

spec/Section 2 -- Language.md
```diff
@@ -587,13 +587,12 @@ FragmentName : Name but not `on`
 
 Fragments are the primary unit of composition in GraphQL.
 
-Fragments allow for the reuse of common repeated selections of fields, reducing
-duplicated text in the document. Inline Fragments can be used directly within a
-selection to condition upon a type condition when querying against an interface
-or union.
+Fragments allow for the definition of modular selection sets that make it easier
+to add and remove selections as needed.
```
benjie, line 591: Maybe something along the lines of (written whilst on Zoom, not final wording!):

```
Fragments allow client components and logic to describe their specific
data requirements locally. Fragments can then be composed together to
form a GraphQL operation to issue to the server.
```

martinbonnin (approved):

martinbonnin (approved):

martinbonnin:
[The video is just out](https://www.youtube.com/watch?v=gMCh8jRVMiQ). Can we merge this and iterate on the wording if needed. I'd love the message to be consistent.

michaelstaib (approved):
Looking good!

spec/Section 2 -- Language.md
```diff
@@ -587,13 +587,13 @@ FragmentName : Name but not `on`
 
 Fragments are the primary unit of composition in GraphQL.
 
-Fragments allow for the reuse of common repeated selections of fields, reducing
-duplicated text in the document. Inline Fragments can be used directly within a
-selection to condition upon a type condition when querying against an interface
-or union.
+Each client component should declare its data needs in a dedicated fragment.
+These fragments may then be composed in the same fashion as the components
+themselves to form a GraphQL operation to issue to the server.
```
benjie: I'm concerned that "component" here will be read too specifically (e.g. to mean a React component) rather than broadly (e.g. a "component" of client logic such as a function, class, or other...). Perhaps something along the lines of:

```suggestion
Each data-consuming component (function, class, UI element, and so on) of
a client application should declare its data needs in a dedicated fragment.
These fragments may then be composed, following the usage of the components
themselves, to form a GraphQL operation to issue to the server.
```

benjie (approved):
I don't know how Lee feels about using `_` in fragment names, but I'd be tempted to adopt Relay naming conventions here and be super explicit.

Perhaps we could lay out the entire scenario showing multi-level composition?

## Example

Each data-consuming component (function, class, UI element, and so on) of a client application should declare its data needs in a dedicated fragment. These fragments may then be composed, following the usage of the components themselves, to form a GraphQL operation to issue to the server.

Consider a social media app. The `userCard(user, layout)` function renders a tiny details card about a user for inclusion in other components. For this, it requires certain information about the `user` which we can describe with a fragment dedicated to this purpose:

```graphql
fragment userCard_user on User {
  id
  name
  profilePic(size: 50)
}
```

When rendering a user's profile with the `userProfile(user)` function, the `userCard` can be used to display each of the user's friends and mutual friends. The `userProfile()` function does not need to know about the specific requirements of the `userCard()` function since it can simply compose it's dedicated fragment in the relevant places:

```graphql
fragment userProfile_user on User {
  id
  name
  profilePic(size: 250)
  biography
  friends(first: 10) {
    ...userCard_user
  }
  mutualFriends(first: 10) {
    ...userCard_user
  }
}
```

This fragment can be used whenever a user profile is displayed, for example when hovering over a user, or on their dedicated user profile page (which may also need to compose fragments for additional concerns):

```graphql
query userProfilePage($id: Int!) {
  user(id: $id) {
    ...userProfile_user
  }
  ...sideBar_query
  ...titleBar_query
}
```

Tobbe:
> Each data-consuming component (function, class, UI element, and so on) ...

I like this! The spec should be as framework and implementation agnostic as possible

> Perhaps we could lay out the entire scenario showing multi-level composition?

Yes! This is great. I'm hearing from our users (RedwoodJS/CedarJS) that fragments are seen as a very advanced concept. A full example like this really helps. 
Maybe the text could be even clearer with the fact that it's entirely natural for a component having to include both the "code definition" (i.e. function signature) and the "data definition" (i.e. fragment) of any children it wants to render.

(And when this change lands, I'll definitely have to update the [Cedar docs on fragments](https://cedarjs.com/docs/graphql/fragments/#what-are-fragments))

leebyron:
> I don't know how Lee feels about using _ in fragment names,

Would like to keep examples camel case in the spec

mjmahone (approved):

Issue #1162, opened 2025-04-25
[RFC] Field extensions


martinbonnin:
Not being able to add directives to an existing field definition has been a growing pain point in Apollo Kotlin. This RFC solves that.

* The RFC PR is [there](https://github.com/graphql/graphql-wg/pull/1729).
* You can read the formatted RFC [there](https://github.com/graphql/graphql-wg/blob/45bc7cc80586f9a3156261506f0d7ccdf0dfbaa5/rfcs/FieldExtensions.md).

This issue is the tracking issue that can be labelled/updated/commented. Feel free to comment here for general feedback or in the [PR](https://github.com/graphql/graphql-wg/pull/1729) for more specific ones.

See https://github.com/graphql/graphql-spec/issues/952

egoodwinx:
I'll work on this!

PR #1196, opened 2025-10-08
Field extensions: spec edits

egoodwinx:
* Tracking issue: https://github.com/graphql/graphql-spec/issues/1162
* graphql-js implementation: https://github.com/graphql/graphql-js/pull/4522


martinbonnin (commented):
Thanks for looking into this! Few comments below

spec/Section 3 -- Type System.md
```diff
@@ -1105,6 +1105,32 @@ Object type extensions have the potential to be invalid if incorrectly defined.
 6. The resulting extended object type must be a super-set of all interfaces it
    implements.
 
+### Field Extensions
+
+FieldExtension :
+
+- extend field MemberCoordinate Directives[Const]
+- extend field Description MemberCoordinate
```
benjie:
```suggestion
```

spec/Section 3 -- Type System.md
```diff
@@ -1105,6 +1105,32 @@ Object type extensions have the potential to be invalid if incorrectly defined.
...
+
+Field extensions are used to represent a field which has been extended from some
+previously defined field. For example this may be a GraphQL service which is
```
benjie:
```suggestion
previously defined field. For example this may be used by a GraphQL service which is
```

spec/Section 3 -- Type System.md
```diff
@@ -1105,6 +1105,32 @@ Object type extensions have the potential to be invalid if incorrectly defined.
...
+itself an extension of another GraphQL service.
+
+In this example, we can deprecate the id field on the User type.
+
+```graphql example
+extend field User.name @deprecated(”Some reason”)
+```
+
+** Field Validation **
```
benjie: The "Type" here relates to "type" as in "type-system" rather than `type` as in "object type".

```suggestion
**Type Validation**
```

spec/Section 3 -- Type System.md
```diff
@@ -1105,6 +1105,32 @@ Object type extensions have the potential to be invalid if incorrectly defined.
...
+
+Field validation have the potential to be invalid if incorrectly defined.
```
benjie:
```suggestion
Fields have the potential to be invalid if incorrectly defined.
```

spec/Section 3 -- Type System.md
```diff
@@ -1105,6 +1105,32 @@ Object type extensions have the potential to be invalid if incorrectly defined.
...
+
+1. MemberCoordinate must be resolved to an existing field on a object or
+   interface type.
```
benjie:
```suggestion
1. MemberCoordinate must resolve to an existing field on a object or
   interface type.
```

benjie (changes requested):
A little editorial

spec/Section 3 -- Type System.md
```diff
@@ -1089,20 +1089,41 @@ In this example, a directive is added to a `User` type without adding fields:
 extend type User @addedDirective
 ```
 
+Object type extensions may choose to extend existing fields.
+
+In this example, we can deprecate the field `id` on type `Query` by adding a
+`@deprecated` directive through a type extension:
```
martinbonnin: Nit: simplify

```suggestion
In this example, we deprecate the field `id` on type `Query` by adding a
`@deprecated` directive:
```

spec/Section 3 -- Type System.md
```diff
@@ -1089,20 +1089,41 @@ In this example, a directive is added to a `User` type without adding fields:
...
+   1. The field must have a unique name within that Object type extension; no
+      two fields may share the same name.
+   2. If a field with the same name exists on the previous Object type:
+      1. The field type must match the previous definition exactly.
+      2. The argument definitions defaultValues must match.
+      3. The description or deprecation may be added if none exists, or must
+         match the existing description or deprecation exactly.
+      4. Directives may be added but cannot conflict with existing
+         non-repeatable directives.
+3. Any interfaces provided must not be already implemented by the previous
```
martinbonnin: A bit more verbose but I think we should be explicit:

```suggestion
2. For each field of an Object type extension:
   1. The field must have a unique name within that Object type extension; no
      two fields may share the same name.
   2. If a field with the same name exists on the previous Object type:
      1. The field type must match the previous definition exactly.
      2. The field description must match the previous definition exactly.
      3. Any non-repeatable directives provided must not already apply to the previous definition.
      4. For each argument of the field:
          1. If an argument with the same name exists on the previous field definition:
              1. The argument type must match the previous definition exactly.
              2. If the argument has a default value, it must match the previous definition exactly.
              3. The argument description must match the previous definition exactly.
              4. Any non-repeatable directives provided must not already apply to the previous definition. 3. Any non-repeatable directives provided must not already apply to the previous definition.
 4. Any interfaces provided must not be already implemented by the previous 
```

mjmahone (commented):
So the biggest thing I'd say is that we should be explicit that we're making `extend` mean `merge`: at each level it's OK for something to be *missing* in the `extend`, in which case that means we use the original value if any, but if it exists on both it must be exactly matching.

spec/Section 3 -- Type System.md
```diff
@@ -1089,20 +1089,49 @@ In this example, a directive is added to a `User` type without adding fields:
...
 
 Object type extensions have the potential to be invalid if incorrectly defined.
 
 1. The named type must already be defined and must be an Object type.
-2. The fields of an Object type extension must have unique names; no two fields
-   may share the same name.
-3. Any fields of an Object type extension must not be already defined on the
-   previous Object type.
-4. Any non-repeatable directives provided must not already apply to the previous
-   Object type.
```
glen-84, line 1102: Is removing these lines intentional? It's the only type-level guard against re-applying a non-repeatable directive through an object extension, and it isn't replaced — the new field (2.ii.3) and argument (2.ii.4.i.4) rules only cover "the previous *definition*", not the object type itself.

Two reasons it looks unintended:

1. The Interface Extensions section keeps its equivalent ("Any non-repeatable directives provided must not already apply to the previous Interface type"); only the object section loses it. Forbidding it on interfaces but permitting it on objects is inconsistent.
2. There is no general "directives are unique per location" rule in the type system — these per-section extension rules are the only enforcement. Without this line, `extend type Foo @nonRepeatable`, where `@nonRepeatable` already applies to `Foo`, is prohibited by nothing.

If it wasn't deliberate, it's worth keeping.

_Drafted by Claude (Anthropic AI assistant)._

spec/Section 3 -- Type System.md
```diff
@@ -1340,14 +1369,47 @@ fields:
...
+            definition:
+            1. The argument type must match the previous definition exactly.
+            2. If the argument has a default value, it must match the previous
+               definition exactly.
+            3. The argument description must match the previous definition
+               exactly.
+            4. Any non-repeatable directives provided must not already apply to
+               the previous definition.
 3. Any fields of an Interface type extension must not be already defined on the
    previous Interface type.
```
glen-84: This line now contradicts the block added just above it. Rule 2 ("For each field of an Interface type extension: … 2. If a field with the same name exists on the previous Interface type: …") permits redeclaring an existing field with a matching definition, but this rule still forbids any field "already defined on the previous Interface type."

The Object Extensions section removed its equivalent line as part of this change; the interface section looks like it should too, otherwise field extensions are simultaneously allowed (rule 2) and disallowed (this rule) on interfaces.

_Drafted by Claude (Anthropic AI assistant)._

glen-84 (changes requested):

spec/Section 3 -- Type System.md
martinbonnin: Good catch @glen-84, thanks! Fixed in https://github.com/graphql/graphql-spec/pull/1196/changes/ed203cf690b5ef5a812925608ac51d700d914e81

spec/Section 3 -- Type System.md
martinbonnin: Also good catch. Fixed in https://github.com/graphql/graphql-spec/pull/1196/changes/fe7e02073169b4571f1655784d3e40683a6a213b

martinbonnin (commented):
Thanks all for the thorough review. 

As it stands, it looks to me like this proposal needs to be adressed more globally as we refine type extension vs type merging and make consistent and predictable rules. 

With @egoodwinx approval, I'll close this one so we can focus on other areas of the spec. 
I'll keep [GAP-4](https://github.com/graphql/gaps/pull/4/) around as a reference for what's implemented in Apollo Kotlin in case other implementers want to try it out. 

Edit: got confirmation out of the band from @egoodwinx that it's ok to close this one.

PR #1203, opened 2025-10-31
Spec edits for incremental delivery, Response Section and Examples Appendix

robrichard:

The response section is especially important, as it defines all of the terms which will be used in the execution algorithm

Helpful reference material:

* Response format examples: https://github.com/graphql/defer-stream-wg/discussions/69
* Glossary: https://github.com/graphql/defer-stream-wg/discussions/106#discussion-7942514
* GraphQL Conf talk: https://www.youtube.com/watch?v=LEyDeNoobT0

### DIFF
--- a/spec/Appendix E -- Examples.md
+++ b/spec/Appendix E -- Examples.md
@@ -21,13 +21,21 @@ fragment HomeWorldFragment on Person {
 }
 ```
 
-The _incremental stream_ might look like:
+The response to this request will be an _incremental stream_ consisting of an
+_initial execution result_ followed by one or more _execution update result_.
 
-The _initial execution result_ does not contain any deferred or streamed results
-in the {"data"} entry. The initial execution result contains a {"hasNext"}
-entry, indicating that _execution update result_ will be delivered. There are
-two _pending result_ indicating that results for both the `@defer` and `@stream`
-in the query will be delivered in the execution update results.
+The _initial execution result_ has:
+
+- a {"data"} entry containing the results of the GraphQL operation except for
+  the `@defer` and `@stream` selections;
+- a {"pending"} entry containing two _pending result_, one for the `@defer`
+  selection and for the the `@stream` selection, indicating that these results
+  will be delivered in a later _execution update result_;
+- a {"hasNext"} entry with the value {true}, indicating that the response is not
+  yet complete.
+
+If an error were to occur, it would also have an {"error"} entry; but not in
+this example.
 
 ```json example
 {
@@ -45,9 +53,11 @@ in the query will be delivered in the execution update results.
 }
 ```
 
-_Execution update result_ 1 contains the deferred data and the first streamed
-list item. There is one _completed result_, indicating that the deferred data
-has been completely delivered.
+Depending on the behavior of the backend and the time at which the deferred and
+streamed resources resolve, the stream may produce results in different orders.
+In this example, our first _execution update result_ contains the deferred data
+and the first streamed list item. There is one _completed result_, indicating
+that the deferred data has been completely delivered.
 
 ```json example
 {
@@ -68,10 +78,10 @@ has been completely delivered.
 }
 ```
 
-_Execution update result_ 2 contains the final stream results. In this example,
-the underlying iterator does not close synchronously so {"hasNext"} is set to
-{true}. If this iterator did close synchronously, {"hasNext"} would be set to
-{false} and this would be the final execution update result.
+The second _execution update result_ contains the final stream results. In this
+example, the underlying iterator does not close synchronously so {"hasNext"} is
+set to {true}. If this iterator did close synchronously, {"hasNext"} would be
+set to {false} and this would be the final execution update result.
 
 ```json example
 {
@@ -85,9 +95,9 @@ the underlying iterator does not close synchronously so {"hasNext"} is set to
 }
 ```
 
-_Execution update result_ 3 contains no incremental data. {"hasNext"} set to
-{false} indicates the end of the _incremental stream_. This response is sent
-when the underlying iterator of the `films` field closes.
+The third _execution update result_ contains no incremental data. {"hasNext"}
+set to {false} indicates the end of the _incremental stream_. This execution
+update result is sent when the underlying iterator of the `films` field closes.
 
 ```json example
 {
@@ -121,14 +131,15 @@ fragment NameAndHomeWorldFragment on Person {
 }
 ```
 
-The _incremental stream_ might look like:
+In this example the response is an _incremental stream_ of the following
+results.
 
 The _initial execution result_ contains the results of the `firstName` field.
 Even though it is also present in the `HomeWorldFragment`, it must be returned
-in the initial response because it is also defined outside of any fragments with
-the `@defer` directive. Additionally, there are two _pending result_ indicating
-that results for both `@defer`s in the query will be delivered in the execution
-update results.
+in the initial execution result because it is also defined outside of any
+fragments with the `@defer` directive. Additionally, there are two _pending
+result_ indicating that results for both `@defer`s in the query will be
+delivered in later _execution update result_.
 
 ```json example
 {
@@ -145,15 +156,17 @@ update results.
 }
 ```
 
-_Execution update result_ 1 contains the deferred data from `HomeWorldFragment`.
-There is one Completed Result, indicating that `HomeWorldFragment` has been
-completely delivered. Because the `homeWorld` field is present in two separate
-`@defer`s, it is separated into its own _incremental result_.
+In this example, the first _execution update result_ contains the deferred data
+from `HomeWorldFragment`. There is one _completed result_, indicating that
+`HomeWorldFragment` has been completely delivered. Because the `homeWorld` field
+is present in two separate `@defer`s, it is separated into its own _incremental
+result_.
 
-The second _incremental result_ contains the data for the `terrain` field. This
-_incremental result_ contains a {"subPath"} entry to indicate to clients that
-the path of this result can be determined by concatenating the path from the
-_pending result_ with id `"0"` and this {"subPath"} entry.
+The second _incremental result_ in this _execution update result_ contains the
+data for the `terrain` field. This _incremental result_ contains a {"subPath"}
+entry to indicate to clients that the _response position_ of this result can be
+determined by concatenating the path from the _pending result_ with id `"0"` and
+the value of this {"subPath"} entry.
 
 ```json example
 {
@@ -173,9 +186,12 @@ _pending result_ with id `"0"` and this {"subPath"} entry.
 }
 ```
 
-_Execution update result_ 2 contains the remaining data from the
-`NameAndHomeWorldFragment`. `lastName` is the only remaining field that has not
-been delivered in a previous response.
+The second _execution update result_ contains the remaining data from the
+`NameAndHomeWorldFragment`. `lastName` is the only remaining field from this
+selection that has not been delivered in a previous result. With this field now
+delivered, clients are informed that the `NameAndHomeWorldFragment` has been
+completed by the presence of the associated _completed result_. Additionally,
+{"hasNext"} is set to {false} indicating the end of the _incremental stream_.
 
 ```json example
 {
--- a/spec/Appendix E -- Examples.md
+++ b/spec/Appendix E -- Examples.md
@@ -22,15 +22,16 @@ fragment HomeWorldFragment on Person {
 ```
 
 The response to this request will be an _incremental stream_ consisting of an
-_initial execution result_ followed by one or more _execution update result_.
+_initial incremental stream result_ followed by one or more _incremental stream
+update result_.
 
-The _initial execution result_ has:
+The _initial incremental stream result_ has:
 
 - a {"data"} entry containing the results of the GraphQL operation except for
   the `@defer` and `@stream` selections;
 - a {"pending"} entry containing two _pending result_, one for the `@defer`
   selection and for the the `@stream` selection, indicating that these results
-  will be delivered in a later _execution update result_;
+  will be delivered in a later _incremental stream update result_;
 - a {"hasNext"} entry with the value {true}, indicating that the response is not
   yet complete.
 
@@ -55,9 +56,9 @@ this example.
 
 Depending on the behavior of the backend and the time at which the deferred and
 streamed resources resolve, the stream may produce results in different orders.
-In this example, our first _execution update result_ contains the deferred data
-and the first streamed list item. There is one _completed result_, indicating
-that the deferred data has been completely delivered.
+In this example, our first _incremental stream update result_ contains the
+deferred data and the first streamed list item. There is one _completed result_,
+indicating that the deferred data has been completely delivered.
 
 ```json example
 {
@@ -78,10 +79,11 @@ that the deferred data has been completely delivered.
 }
 ```
 
-The second _execution update result_ contains the final stream results. In this
-example, the underlying iterator does not close synchronously so {"hasNext"} is
-set to {true}. If this iterator did close synchronously, {"hasNext"} would be
-set to {false} and this would be the final execution update result.
+The second _incremental stream update result_ contains the final stream results.
+In this example, the underlying iterator does not close synchronously so
+{"hasNext"} is set to {true}. If this iterator did close synchronously,
+{"hasNext"} would be set to {false} and this would be the final incremental
+stream update result.
 
 ```json example
 {
@@ -95,9 +97,10 @@ set to {false} and this would be the final execution update result.
 }
 ```
 
-The third _execution update result_ contains no incremental data. {"hasNext"}
-set to {false} indicates the end of the _incremental stream_. This execution
-update result is sent when the underlying iterator of the `films` field closes.
+The third _incremental stream update result_ contains no incremental data.
+{"hasNext"} set to {false} indicates the end of the _incremental stream_. This
+incremental stream update result is sent when the underlying iterator of the
+`films` field closes.
 
 ```json example
 {
@@ -134,12 +137,12 @@ fragment NameAndHomeWorldFragment on Person {
 In this example the response is an _incremental stream_ of the following
 results.
 
-The _initial execution result_ contains the results of the `firstName` field.
-Even though it is also present in the `HomeWorldFragment`, it must be returned
-in the initial execution result because it is also defined outside of any
-fragments with the `@defer` directive. Additionally, there are two _pending
-result_ indicating that results for both `@defer`s in the query will be
-delivered in later _execution update result_.
+The _initial incremental stream result_ contains the results of the `firstName`
+field. Even though it is also present in the `HomeWorldFragment`, it must be
+returned in the initial incremental stream result because it is also defined
+outside of any fragments with the `@defer` directive. Additionally, there are
+two _pending result_ indicating that results for both `@defer`s in the query
+will be delivered in later _incremental stream update result_.
 
 ```json example
 {
@@ -156,17 +159,17 @@ delivered in later _execution update result_.
 }
 ```
 
-In this example, the first _execution update result_ contains the deferred data
-from `HomeWorldFragment`. There is one _completed result_, indicating that
-`HomeWorldFragment` has been completely delivered. Because the `homeWorld` field
-is present in two separate `@defer`s, it is separated into its own _incremental
-result_.
+In this example, the first _incremental stream update result_ contains the
+deferred data from `HomeWorldFragment`. There is one _completed result_,
+indicating that `HomeWorldFragment` has been completely delivered. Because the
+`homeWorld` field is present in two separate `@defer`s, it is separated into its
+own _incremental result_.
 
-The second _incremental result_ in this _execution update result_ contains the
-data for the `terrain` field. This _incremental result_ contains a {"subPath"}
-entry to indicate to clients that the _response position_ of this result can be
-determined by concatenating the path from the _pending result_ with id `"0"` and
-the value of this {"subPath"} entry.
+The second _incremental result_ in this _incremental stream update result_
+contains the data for the `terrain` field. This _incremental result_ contains a
+{"subPath"} entry to indicate to clients that the _response position_ of this
+result can be determined by concatenating the path from the _pending result_
+with id `"0"` and the value of this {"subPath"} entry.
 
 ```json example
 {
@@ -186,8 +189,8 @@ the value of this {"subPath"} entry.
 }
 ```
 
-The second _execution update result_ contains the remaining data from the
-`NameAndHomeWorldFragment`. `lastName` is the only remaining field from this
+The second _incremental stream update result_ contains the remaining data from
+the `NameAndHomeWorldFragment`. `lastName` is the only remaining field from this
 selection that has not been delivered in a previous result. With this field now
 delivered, clients are informed that the `NameAndHomeWorldFragment` has been
 completed by the presence of the associated _completed result_. Additionally,
--- a/spec/Appendix E -- Examples.md
+++ b/spec/Appendix E -- Examples.md
@@ -126,7 +126,7 @@ The _incremental stream_ might look like:
 The _initial execution result_ contains the results of the `firstName` field.
 Even though it is also present in the `HomeWorldFragment`, it must be returned
 in the initial response because it is also defined outside of any fragments with
-the `@defer` directive. Additionally, There are two _pending result_ indicating
+the `@defer` directive. Additionally, there are two _pending result_ indicating
 that results for both `@defer`s in the query will be delivered in the execution
 update results.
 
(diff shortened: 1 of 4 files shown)


spec/Appendix E -- Examples.md
```diff
@@ -0,0 +1,191 @@
...
+      "subPath": ["homeWorld"],
+      "data": { "terrain": "desert" }
+    }
+  ],
+  "completed": [{ "id": "0" }],
+  "hasNext": true
+}
+```
+
+Subsequent response 2, contains the remaining data from the
```
BoD:
```suggestion
Subsequent response 2 contains the remaining data from the
```

spec/Section 7 -- Response.md
```diff
@@ -323,15 +388,171 @@ discouraged.
...
+An _incremental result_ must be a map.
+
+Every _incremental result_ must contain an entry with the key {"id"} with a
+string value. The definition of _associated pending result_ describes how this
+value is used to determine the associated pending result for a given
+_incremental result_.
+
+#### Incremental List Result
+
+:: An _incremental list result_ is a _incremental result_ used to deliver
```
BoD:
```suggestion
:: An _incremental list result_ is an _incremental result_ used to deliver
```

spec/Section 7 -- Response.md
```diff
@@ -323,15 +388,171 @@ discouraged.
...
+
+If any _execution error_ were raised during the execution of the results in
+{"items"} and these errors did not propagate to a path higher than the
+_incremental list result_'s path, the incremental list result must contain an
+entry with key {"errors"} containing these execution errors. The value of this
+entry is described in the "Errors" section.
+
+#### Incremental Object Result
+
+:: An _incremental object result_ is a _incremental result_ used to deliver
```
BoD:
```suggestion
:: An _incremental object result_ is an _incremental result_ used to deliver
```

BoD (approved):

benjie (dismissed):
I don't know if it's just me, but I feel like the examples could use more narrative and explanation that this is just one way the query could turn out. Here's how I'd change the beginning of the first example:

spec/Appendix E -- Examples.md
```diff
@@ -0,0 +1,210 @@
...
+
+- a {"data"} entry containing the results of the GraphQL operation except for
+  the `@defer` and `@stream` selections;
+- a {"pending"} entry containing two _pending result_, one for the `@defer`
+  selection and for the the `@stream` selection, indicating that these results
+  will be delivered in a later _incremental stream update result_;
+- a {"hasNext"} entry with the value {true}, indicating that the response is not
+  yet complete.
+
+If an error were to occur, it would also have an {"error"} entry; but not in
```
benjie:
```suggestion
If an error were to occur, it would also have an {"errors"} entry; but not in
```

spec/Appendix E -- Examples.md
```diff
@@ -0,0 +1,210 @@
...
+    { "id": "1", "path": ["person", "films"], "label": "filmsStream" }
+  ],
+  "hasNext": true
+}
+```
+
+Depending on the behavior of the backend and the time at which the deferred and
+streamed resources resolve, the stream may produce results in different orders.
+In this example, our first _incremental stream update result_ contains the
+deferred data and the first streamed list item. There is one _completed result_,
```
benjie: Maybe change "completed result" to "incremental completion result" to tie it in with the "incremental" naming? (Also to be more specific.)

```suggestion
deferred data and the first streamed list item. There is one _incremental completion result_,
```

spec/Appendix E -- Examples.md
```diff
@@ -0,0 +1,210 @@
...
+  ]
+  "hasNext": true
+}
+```
+
+The second _incremental stream update result_ contains the final stream results.
+In this example, the underlying iterator does not close synchronously so
+{"hasNext"} is set to {true}. If this iterator did close synchronously,
+{"hasNext"} would be set to {false} and this would be the final incremental
+stream update result.
```
benjie: I think "could" rather than "would" to allow vendor flexibility?

```suggestion
{"hasNext"} is set to {true}. If this iterator did close synchronously,
{"hasNext"} could be set to {false} and make this the final incremental
stream update result.
```

spec/Appendix E -- Examples.md
```diff
@@ -0,0 +1,210 @@
...
+    }
+  ],
+  "hasNext": true
+}
+```
+
+The third _incremental stream update result_ contains no incremental data.
+{"hasNext"} set to {false} indicates the end of the _incremental stream_. This
+incremental stream update result is sent when the underlying iterator of the
+`films` field closes.
```
benjie:
```suggestion
When the underlying iterator of the `films` field closes there is no more data
to deliver, so the third and final _incremental stream update result_ sets
{"hasNext"} to {false} to indicate the end of the _incremental stream_.
```

spec/Appendix E -- Examples.md
```diff
@@ -0,0 +1,210 @@
...
+deferred data from `HomeWorldFragment`. There is one _completed result_,
+indicating that `HomeWorldFragment` has been completely delivered. Because the
+`homeWorld` field is present in two separate `@defer`s, it is separated into its
+own _incremental result_.
+
+The second _incremental result_ in this _incremental stream update result_
+contains the data for the `terrain` field. This _incremental result_ contains a
+{"subPath"} entry to indicate to clients that the _response position_ of this
+result can be determined by concatenating the path from the _pending result_
+with id `"0"` and the value of this {"subPath"} entry.
```
benjie: This reads a little ambiguously: one misinterpretation would be "concatenating (the path from the pending result) with (id "0") and (the value of this subpath entry)".

I've tweaked the grammar slightly to avoid this:

```suggestion
result can be determined by concatenating: the path from the _pending result_
for id `"0"`, and the value of this {"subPath"} entry.
```

spec/Appendix E -- Examples.md
```diff
@@ -0,0 +1,210 @@
...
+
+```json example
+{
+  "incremental": [
+    {
+      "id": "0",
+      "data": { "homeWorld": { "name": "Tatooine" } }
+    },
```
benjie, line 183: I feel like the weirdness of this needs more calling out - name belongs to both id:0 and id:1, so we're kind of using id:0 as a shortcut (but id:1 would also work). We should make it clear to a client that it can't form `HomeWorldFragment` from solely the root data plus the `id:0` stuff and `NameAndHomeWorldFragment` from solely the root data and the `id:1` stuff - they must merge the trees before extracting the data.

spec/Section 7 -- Response.md
```diff
@@ -70,6 +79,70 @@ The _request error result_ map must not contain an entry with key {"data"}.
...
+The _initial incremental stream result_ must contain an entry with key {"data"},
+and may contain entries with keys {"errors"} and {"extensions"}. The value of
+these entries are defined in the same way as an _execution result_ as described
+in the "Data", "Errors", and "Extensions" sections below.
+
+The _initial incremental stream result_ must contain an entry with the key
+{"hasNext"}. The value of this entry must be {true} if there are any
+_incremental stream update results_ in the _incremental stream_. The value of
+this entry must be {false} if the initial incremental stream result is the last
+response of the incremental stream.
```
benjie: Splitting this across two paragraphs is a little weird, I feel like the list of keys up front should be exhaustive.


```suggestion
The _initial incremental stream result_ must contain an entries with keys {"data"} and {"hasNext"}, and may contain entries with keys
{"errors"}, {"pending"}, {"incremental"}, {"completed"}, and {"extensions"}.

The value of {"data"}, {"errors"} and {"extensions"} are defined in the
same way as an _execution result_ as described in the "Data", "Errors", and
"Extensions" sections below.

The value of the {"hasNext"} entry must be {false}
if the initial incremental stream result is the last response of the incremental
stream. Otherwise, {"hasNext"} must be {true}.
```

I'd like to see a non-normative note below this explaining that {"hasNext"} is required only for it to be an initial incremental stream result, and that omitting it is valid if it's the only result, just that it means it's an execution result instead (i.e. there's no "incremental" at all). In this case, the GraphQL response would not return a stream.

This is a terrible version of that note since I'm writing this already over-time on our call...

Note: Omitting {"hasNext"} would make the response a regular _execution result_, which is valid only when returned directly and not as part of a stream.

spec/Section 7 -- Response.md
```diff
@@ -70,6 +79,70 @@ The _request error result_ map must not contain an entry with key {"data"}.
...
+
+The _initial incremental stream result_ must contain an entry with the key
+{"pending"}. The value of this entry must be a non-empty list of _pending
+result_. Each _pending result_ must be a map as described in the "Pending
+Result" section below.
+
+The _initial incremental stream result_ may contain an entry with they key
+{"incremental"}. The value of this entry must be a non-empty list of
+_incremental result_. Each _incremental result_ must be a map as described in
+the "Incremental Result" section below.
```
benjie: I'd add a non-normative note detailing why this is allowed.

benjie: Same for "completed"

spec/Section 7 -- Response.md
```diff
@@ -70,6 +79,70 @@ The _request error result_ map must not contain an entry with key {"data"}.
...
+The _initial incremental stream result_ must contain an entry with the key
+{"hasNext"}. The value of this entry must be {true} if there are any
+_incremental stream update results_ in the _incremental stream_. The value of
+this entry must be {false} if the initial incremental stream result is the last
+response of the incremental stream.
+
+The _initial incremental stream result_ must contain an entry with the key
+{"pending"}. The value of this entry must be a non-empty list of _pending
+result_. Each _pending result_ must be a map as described in the "Pending
+Result" section below.
```
benjie: If "hasNext" is true, pending _must_ be present.
~~If "hasNext" is false, pending _must not_ be present.~~ < Not true, could be hasNext:false along with pending:[...],incremental:[...],completed:[...]

spec/Section 7 -- Response.md
```diff
@@ -70,6 +79,70 @@ The _request error result_ map must not contain an entry with key {"data"}.
...
+An _incremental stream update result_ must contain an entry with the key
+{"hasNext"}. The value of this entry must be {true} for all but the last
+response in the _incremental stream_. The value of this entry must be {false}
+for the last response of the incremental stream.
+
+The _incremental stream update result_ may contain entries with keys
+{"pending"}, {"incremental"}, and/or {"completed"}. The value of these entries
+are defined in the same way as an _initial incremental stream result_ as
+described in the "Pending Result", "Incremental Result", and "Completed Result"
+sections below.
```
benjie: Should we add that `{"hasNext": true}` alone is invalid?

I propose that we *do not* do so. Not doing so allows `{"hasNext: true}` to be used as a keepalive.

spec/Section 7 -- Response.md
```diff
@@ -94,6 +167,9 @@ represents a path in the response, not in the request.
 When a _response path_ is present on an _error result_, it identifies the
 _response position_ which raised the error.
 
+When a _response path_ is present on an _incremental result_, it identifies the
+_response position_ of the incremental data update.
+
```
benjie, line 182: ... _can_ a response path be present on an _incremental result_? Maybe you mean _pending result_?

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+a pending result indicates that clients should expect the associated data in
+either the current response, or one of the following responses.
+
+**Pending Result Format**
+
+A _pending result_ must be a map.
+
+Every _pending result_ must contain an entry with the key {"id"} with a string
+value. This {"id"} should be used by clients to correlate pending results with
+_incremental result_ and _completed result_. The {"id"} value must be unique for
```
benjie:
```suggestion
_incremental result_ and _completed result_. The {"id"} value must be unique across
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+the entire _incremental stream_ response. There must not be any other pending
+result in the _incremental stream_ that contains the same {"id"}.
```
benjie:
```suggestion
result in the _incremental stream_ with the same {"id"}.
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+update result_ in the _incremental stream_.
+
+:: The _associated pending result_ is a specific _pending result_ associated
+with any given _incremental result_ or _completed result_. The associated
+pending result can be determined by finding the pending result where the value
+of its {"id"} entry is the same value of the {"id"} entry of the given
+incremental result or completed result. The associated pending result must
+appear in the _incremental stream_, in the same or prior _initial incremental
+stream result_ or _execution update result_ as the given incremental result or
+completed result.
```
benjie:
```suggestion
:: The _associated pending result_ of an _incremental result_ or _completed
result_ is the _pending result_ whose {"id"} entry has the same value as the
{"id"} entry of the given incremental result or completed result.
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+:: The _incremental result_ is used to deliver data that the GraphQL service has
+chosen to incrementally deliver. An incremental result may be either an
+_incremental list result_ or an _incremental object result_.
+
+An _incremental result_ must be a map.
+
+Every _incremental result_ must contain an entry with the key {"id"} with a
+string value. The definition of _associated pending result_ describes how this
+value is used to determine the associated pending result for a given
+_incremental result_.
```
benjie:
```suggestion
Every _incremental result_ must contain an entry with the key {"id"}, the value
of which is a string referencing its _associated pending result_. The associated
pending result must appear either in the _initial incremental stream result_,
in a prior _incremental stream updated result_, or in the same _incremental
stream updated result_ as the _incremental result_ that references it.
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+Every _incremental list result_ must contain an {"items"} entry. The {"items"}
+entry must contain a list of additional list items for the list field in the
+incremental list result's _response position_. The value of this entry must be a
+list of the same type of the response field at this _response position_.
+
+If any _execution error_ were raised during the execution of the results in
+{"items"} and these errors propagate to a _response position_ higher than the
+_incremental list result_'s response position, the incremental list result is
+considered failed and should not be included in the _incremental stream_. The
+errors that caused this failure will be included in a _completed result_.
```
benjie: This doesn't feel like the right wording for error propagation here ("response position higher than ..." doesn't feel right) but it's too late at night right now for me to figure out the correct wording.

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+
+If any _execution error_ were raised during the execution of the results in
+{"items"} and these errors did not propagate to a path higher than the
+_incremental list result_'s path, the incremental list result must contain an
+entry with key {"errors"} containing these execution errors. The value of this
+entry is described in the "Errors" section.
```
benjie: Should we add a non-normative (or normative, maybe) note to say that the response path of these errors will always start with the response position followed by an index? Want to make sure that no-one attempts to attach the error directly to the list, for example.

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+result_ must be associated with a `@defer` directive.
+
+**Incremental Object Result Format**
+
+The _incremental object result_ may contain a {"subPath"} entry. If this entry
+is present, the incremental object result's _response position_ can be
+determined by concatenating the value of the _associated pending result_'s
+{"path"} entry with the value of this {"subPath"} entry. If no {"subPath"} entry
+is present, the _response position_ is the value of the associated pending
+result's {"path"} entry.
```
benjie:
```suggestion
The _incremental object result_ may contain a {"subPath"} entry. If such an entry
is present, the _response position_ of the incremental object result is the
result of appending the value of this {"subPath"} to the value of the {"path"}
entry of the _associated pending result_. If no {"subPath"} entry is present, the
_response position_ is the value of the associated pending result's {"path"}
entry.
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+
+An _incremental object result_ may be used to deliver data for response fields
+that were contained in more than one deferred fragments. In that case, the
```
benjie:
```suggestion
that were contained in more than one deferred fragment. In that case, the
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+_associated pending result_ of the incremental object result must be a _pending
+result_ with the longest {"path"}.
```
benjie: This is insufficiently precise; it must relate to the deferred fragment. Further, is there a way in which this can be broken when combined with adding incremental/completed to the initial incremental execution result through eager execution and maybe nested fragments? I feel like you might asynchronously complete the requirements and reference a short path, and then a long path comes about later (whilst still executing the non-deferred selection set)...

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+
+Every _incremental object result_ must contain a {"data"} entry. The {"data"}
+entry must contain a map of additional response fields. The {"data"} entry in an
+incremental object result will be of the type of the field at the incremental
+object result's _response position_.
```
benjie, line 544: "will be of the type of the field" feels wrong. The field may be `User` but if the selection set is `{ foo: username }` then the "type" is `{ foo: String }`. It's more about the execution of the selections than the type of the field. (Similar concern for lists.)

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+
+If any _execution error_ were raised during the execution of the results in
+{"data"} and these errors propagated to a _response position_ higher than the
+_incremental object result_'s response position, the incremental object result
+is considered failed and should not be included in the incremental stream. The
+errors that caused this failure will be included in a _completed result_.
```
benjie: Is "errors" correct? I think not: when response bubbling causes a selection set to be destroyed you only get one error independent of how many bubbled, and I think that should be the same here?

```suggestion
error that caused this failure will be included in a _completed result_.
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+
+If any _execution error_ were raised during the execution of the results in
+{"data"} and these errors did not propagate to a _response position_ higher than
```
benjie:
```suggestion
{"data"} and no such error propagated to a _response position_ higher than
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+_initial incremental stream result_ or _incremental stream update result_ in
+which this completed result appears.
+
+**Completed Result Format**
+
+A _completed result_ must be a map.
+
+Every _completed result_ must contain an entry with the key {"id"} with a string
+value. The definition of _associated pending result_ describes how this value is
+used to determine the associated pending result for a given _completed result_.
```
benjie:
```suggestion
Every _completed result_ must contain an entry with the key {"id"}, the
value of which is a string referencing its _associated pending result_. The
associated pending result must appear either in the _initial incremental stream
result_, in a prior _incremental stream updated result_, or in the same
_incremental stream updated result_ as the _completed result_ that references
it.
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+associated with a `@stream` directive.
+
+The _response position_ for an _incremental list result_ is the {"path"} entry
+from its _associated pending result_.
+
+**Incremental List Result Format**
+
+Every _incremental list result_ must contain an entry with the key {"id"}, used
+to determine the _associated pending result_ for this _incremental result_.
+
```
benjie: Already covered in _incremental result_?

```suggestion
```

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+Every _completed result_ must contain an entry with the key {"id"} with a string
+value. The definition of _associated pending result_ describes how this value is
+used to determine the associated pending result for a given _completed result_.
+
+A _completed result_ may contain an {"errors"} entry. When the {"errors"} entry
+is present, it informs clients that the delivery of the data from the
+_associated pending result_ has failed, due to an execution error propagating to
+a _response position_ higher than the _incremental result_'s response position.
+The {"errors"} entry must contain these execution errors. The value of this
+entry is described in the "Errors" section.
```
benjie: 🤔  Can errors ever contain more than one error?

Answer: yes - for defer there be many associated selection sets at different levels, and multiple of these may fail due to error propagation.

spec/Section 7 -- Response.md
```diff
@@ -323,17 +399,173 @@ discouraged.
...
+
 ### Additional Entries
 
 To ensure future changes to the protocol do not break existing services and
-clients, the _execution result_ and _request error result_ maps must not contain
-any entries other than those described above. Clients must ignore any entries
-other than those described above.
+clients, any of the maps described in the "Response" section (with the exception
+of {"extensions"}) must not contain any entries other than those described
+above. Clients must ignore any entries other than those described above.
```
benjie, line 591: We could probably upstream this change.

yaacovCR (approved):
This is amazing work, and others have given already and I’m sure we’ll continue to give wonderful feedback.

My two cents is that it may be beneficial to refer to the pending/completed and possibly even the incremental entries as notifications rather than results. The result is the deferred/streamed data at that path that has an id. It’s an abstract concept in terms of the response format, to be assembled, perhaps as needed, by the client.

The response can strictly speaking never send a pending result, in fact that’s in some sense an internal contradiction, or, at best, a shorthand for what we really mean, a notification or description of a result that is pending/completed. The incremental entries are actually results/updates, although calling them as such may confuse them with the overarching defer/stream result that they help complete.

I offer this comment just as a suggestion, I could live with this somewhat confusing shorthand if others feel that it is the best compromise in terms of clarity.

spec/Section 7 -- Response.md
robrichard: This suggested edit changes pending from required to optional on the initial incremental stream result, I don't think we have a case for an incremental stream without a pending? I will keep it as required when I incorporate this change.

Do you still think we need a note about omitting hasNext? it's a bit more complicated as you also need to omit pending?

I will however add a note explaining the reasoning for allowing hasNext: false in the initial incremental stream result

spec/Section 7 -- Response.md
robrichard: Going to leave this as is since `pending` must be present in all cases for the initial incremental stream result

spec/Section 7 -- Response.md
robrichard: Agreed, do you think we should specifically mention this, outside of every other entry being defined as optional?

spec/Section 7 -- Response.md
robrichard: Updating to 
> and these errors propagate to a parent _response position_ of the _incremental list result_'s response position

`parent` and `sibling` are used in the execution section

spec/Section 7 -- Response.md
robrichard: I added more detail to this section that I think makes this more clear.

spec/Section 7 -- Response.md
robrichard: Updated this section to be more precise.

I think since we know about all of the deferred fields after collect fields, the server should always be able to tell which is the longest pending path regardless of the execution order.

spec/Section 7 -- Response.md
robrichard: I was basing it off of this existing section, where we say the root data is either the query or mutation root operation type:

> If the operation was a query, this output will be an
object of the query root operation type; if the operation was a mutation, this
output will be an object of the mutation root operation type.

spec/Appendix E -- Examples.md
robrichard: Added a paragraph here: https://github.com/graphql/graphql-spec/pull/1203/changes#diff-0237c26b29c484df8d43b13dc3a3658867cfd015d8ccea8ed93454c466dbacb3R200-R203

spec/Section 7 -- Response.md
robrichard: Note added here: https://github.com/graphql/graphql-spec/pull/1203/changes#diff-98d0cd153b72b63c417ad4238e8cc0d3385691ccbde7f7674bc0d2a718b896ecR116-R126

spec/Appendix E -- Examples.md
robrichard: Updated now to _incremental completion notice_

robrichard:
@yaacovCR @benjie The following terms have now been updated:

Pending Result => Incremental Pending Notice
Completed Result => Incremental Completion Notice

PR #1206, opened 2025-12-18
Add support for directives on directive definitions

BoD:
Adds the ability to apply directives to directive definitions and to deprecate directives, with this syntax:

```graphql
directive @onDirective on DIRECTIVE_DEFINITION

directive @foo @onDirective on OBJECT

directive @baz @deprecated(reason: "...") on OBJECT

extend directive @quux @deprecated(reason: "...")
```

With @IvanGoncharov's blessing, this PR is a continuation and updated version of #907 which I think can be closed. (There's also prior art in #567).

## Motivation
Allow applying directive-exclusive features (e.g. `@specifyBy`, `@deprecated`) on directives.

For instance, on Apollo Kotlin we'd like to deprecate some client side directives.


BoD:
Related graphql-js PR: https://github.com/graphql/graphql-js/pull/4521

martinbonnin:
Another use case: [`@ignore`](https://github.com/apollographql/specs/pull/82)

spec/Section 3 -- Type System.md
```diff
@@ -2321,3 +2323,24 @@ input UserUniqueCondition @oneOf {
...
+or service which adds directives to an existing directive.
+
+**Type Validation**
+
+Directive extensions have the potential to be invalid if incorrectly defined.
+
+1. The previous directive must already be defined.
+2. Any non-repeatable directives provided must not already apply to the previous
+   directive.
+3. Any directives provided must not contain the use of a Directive which
```
yaacovCR, line 2342: Are these cyclical rules enforced yet by the graphql-js PRs? ~~(and is previous the correct word?)~~

yaacovCR: (1) previous is the correct word, it's used throughout the spec to describe the entity being extended
(2) these cyclical rules are missing, but they are not just missing with respect to directive extensions, they are missing with respect to directive definitions and so must be added to graphql-js (as far as I can tell!) but must be added irrespective of this PR.

yaacovCR: opened https://github.com/graphql/graphql-js/pull/4726 to discuss

spec/Section 3 -- Type System.md
```diff
@@ -2321,3 +2323,24 @@ input UserUniqueCondition @oneOf {
   organizationAndEmail: OrganizationAndEmailInput
 }
 ```
+
+### Directive Extensions
+
+DirectiveExtension : extend directive @ Name Directives[Const]
```
mjmahone, line 2329: This is interesting: should we be able to extend a directive with new arguments? We can extend a type with new fields that themselves have new arguments.

There is even a special Schema Coordinate for directive arguments.

BoD: Good question! 

In my opinion, I think yes it would probably make sense in principle to have that ability, even if for consistency alone. 
That being said, I would rather tackle this in a separate change later, and based on demand for it. 

Curious to know what others think!

mjmahone: I don't *think* it would be controversial, but it does make sense to stack into two PRs to make discussion specifically about argument extension more focused.

spec/Appendix D -- Specified Definitions.md
```diff
@@ -33,7 +33,7 @@ type __Schema {
   queryType: __Type!
   mutationType: __Type
   subscriptionType: __Type
-  directives: [__Directive!]!
+  directives(includeDeprecated: Boolean! = false): [__Directive!]!
```
benjie, line 36: I'm hesitant to add this, since I think the whole `includeDeprecated` dance was a mistake. I think I'd rather see the `includeDeprecated` argument be itself deprecated across the board, but it defaulting to `= false` makes this challenging.

benjie (approved):
Looks good to me, thanks for your work on this! I'd like to discuss `includeDeprecated` at a WG to determine if we might be able to address this e.g. via `__schema(includeDeprecated: Boolean! = false)` and then remove the need for this new argument, but otherwise I approve 👍

spec/Section 3 -- Type System.md
martinbonnin: @mjmahone Am I correct that https://gaps.graphql.org/GAP-33/draft/ would also solve this use case?

spec/Appendix D -- Specified Definitions.md
BoD: Yeah I certainly agree about `includeDeprecated`'s benefits being outweighed by its drawbacks. Having it here was about consistency - but maybe if we're planning on deprecating it, we can start by not having it here! Let's discuss this at the next WG.

michaelstaib (approved):

spec/Section 3 -- Type System.md
mjmahone: well, GAP-33 *assumes* that this is already available.

But I think I should iterate on that GAP to make *inner* extensions use a different syntax because as-is it's not good!

PR #1207, opened 2026-01-06
Update wording for 6.4.1 Coercing Field Arguments

magicmark:
This sentence ~doesn't quite scan as is~ is a little awkward to parse

### DIFF
--- a/spec/Section 6 -- Execution.md
+++ b/spec/Section 6 -- Execution.md
@@ -736,8 +736,8 @@ Fields may include arguments which are provided to the underlying runtime in
 order to correctly produce a value. These arguments are defined by the field in
 the type system to have a specific input type.
 
-At each argument position in an operation may be a literal {Value}, or a
-{Variable} to be provided at runtime.
+Each argument position in an operation may be a literal {Value}, or a {Variable}
+to be provided at runtime.
 
 CoerceArgumentValues(objectType, field, variableValues):
 


martinbonnin (approved):
Thanks!

benjie (changes requested):
These don’t read the same to me; the existing text reads as “there may be an entry at each argument position, and if there is then its a Value or Variable” whereas the replacement seems to say “each argument position is a Value or Variable” and doesn’t as clearly allow for omission.

I think the existing text is fine as is, though it does feel slightly awkwardly phrased.

spec/Section 6 -- Execution.md
```diff
@@ -736,8 +736,8 @@ Fields may include arguments which are provided to the underlying runtime in
 order to correctly produce a value. These arguments are defined by the field in
 the type system to have a specific input type.
 
-At each argument position in an operation may be a literal {Value}, or a
-{Variable} to be provided at runtime.
+Each argument position in an operation may be a literal {Value}, or a {Variable}
+to be provided at runtime.
```
benjie, line 740: Looking at Martin's proposed edit... what _is_ an "argument position in an executable document"? To me it's where an argument is passed to a field in a document (which is irrespective of the actual type system - the field doesn't even need to exist in the schema):

```graphql
query {
  __nonsenseDoesNotExistField__(arg: >>>THIS IS AN ARGUMENT POSITION IN AN EXECUTABLE DOCUMENT<<<<)
}
```

As such, it _can only be_ a literal or a variable... there's no other option. So I think Martin's second option can be simplified a bit, it's informational (`contains`) rather than normative (`may`):

> Each argument position in an executable document ~~may be~~ contains a literal {Value}, or a {Variable}
> to be provided at runtime. Arguments may also be missing if their definition allows it.

However, I feel like the "argument position in an executable document", whilst much more concrete, is potentially meaning something different than that which the previous text implied, so I think Martin's additional clarifying sentence might be needed - but we shouldn't rely on validation here ("if their definition allows it") since execution is allowed to be performed without validation. A "for example" might be a softer way of writing this:

```suggestion
Each argument position in an executable document contains a literal {Value}, or
a {Variable} to be provided at runtime. Some arguments may not be represented
in the document, for example if they are optional.
```

martinbonnin (approved):

BoD (approved):

spec/Section 6 -- Execution.md
```diff
@@ -736,8 +736,9 @@ Fields may include arguments which are provided to the underlying runtime in
 order to correctly produce a value. These arguments are defined by the field in
 the type system to have a specific input type.
 
-At each argument position in an operation may be a literal {Value}, or a
-{Variable} to be provided at runtime.
+Each argument position in an executable document contains a literal {Value}, or
+a {Variable} to be provided at runtime. Some arguments may not be represented in
+the document, for example if they are optional.
```
benjie, line 741: "argument position" is only used in this one place in the spec, and it's weird. We refer to "arguments" being "represented" in the latter half of this paragraph, I've incorporated that into the former for consistency.

```suggestion
Each argument represented in an executable document includes a literal {Value},
or a {Variable} to be provided at runtime. Some arguments may not be represented
in the document, for example if they are optional.
```

benjie (changes requested):
I suggest this very minor tweak so we can remove the ill-defined concept of an "argument position".

spec/Section 6 -- Execution.md
magicmark: 'includes' is perhaps overloaded in meaning, took my brain a minute to parse

is the intended usage of "include" here is in its verb form? (e.g. "include the sugar in the cake mix")

at first, my brain read "includes" it in the context of lists/object, so i was confused and thought wait how can a single argument "include" other stuff?

```diff
- Each argument represented in an executable document includes a literal {Value}
+ Each argument represented in an executable document will be a literal {Value}
```

if i'm understanding your intent correctly, perhaps "will be" should be "maps to" or something 

agreed this is very minor though, feel free to drop

PR #1212, opened 2026-02-24
fix: field selection set final values

MarkoMin:
AFAIK scalar and enum values are considered to be distinct throughout the specification so I think this would be more correct description of field selection.

I didn't change the sentence below the modified one because it doesn't contain any enum values, but I may add some if that'd be beneficial.

It really is a nitpick, but I stopped on it while reading the spec.

### DIFF
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -419,8 +419,7 @@ discrete piece of information available to request within a selection set.
 Some fields describe complex data or relationships to other data. In order to
 further explore this data, a field may itself contain a selection set, allowing
 for deeply nested requests. All GraphQL operations must specify their selections
-down to fields which return scalar or enum values to ensure an unambiguously
-shaped response.
+down to _leaf fields_ to ensure an unambiguously shaped response.
 
 For example, this operation selects fields of complex data and relationships
 down to scalar values.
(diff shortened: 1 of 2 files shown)


martinbonnin (approved):

yaacovCR (approved):

benjie (approved):
We have the concept of "leaf fields" and "leaf values" already (though it is introduced later in the spec), maybe we should reference that here instead. One concern at the moment is the current wording doesn't handle lists of scalars/enums (or lists of those lists!).

It's still better than what we currently have though; thanks for the contribution!

martinbonnin (approved):

MarkoMin:
Sorry, I forgot formatting. Now tests should pass.

martinbonnin:
Thanks! For the record, I'm not sure this requires a working group item, I'm fine to merge this async. I mostly added it as a forcing function to review this PR if we haven't acted by then.

martinbonnin:
@MarkoMin it's working group day 🎉 => https://github.com/graphql/graphql-wg/blob/main/agendas/2026/04-Apr/02-wg-primary.md 

If you want to join, please add your name to the agenda. If not, no worries, I'll present and post an update here.

leebyron (approved):

mjmahone (approved):

PR #1217, opened 2026-04-01
Rename “triple quotes“ -> “sextuple quotes“

BoD:
Since `"` are known as “double quotes”, it is only logical that `"""` should **not** be known as “triple quotes” (which would be `'''`) but as “sextuple quotes”.


martinbonnin (approved):
Finally!

BoD:
(That was my attempt as a 🐟 joke. I'll see myself out now.)

PR #1219, opened 2026-04-05
Prevent email disclosure in contributor list generation

RinZ27:
Contributor list script currently defaults to printing the raw email address when a GitHub login cannot be resolved. This behavior unintentionally exposes private email addresses in public-facing Markdown files.

The updated logic ensures that only names are displayed for unresolved contributors, leaving the GitHub link column empty instead of leaking PII. This improvement protects contributor privacy while maintaining the integrity of the generated list. Verified the script still functions correctly after this modification.

### DIFF
--- a/scripts/generate-contributor-list.mjs
+++ b/scripts/generate-contributor-list.mjs
@@ -319,7 +319,7 @@ async function main() {
     const key = normalizeName(nm) + "|" + email.toLowerCase();
     if (seenUnk.has(key)) continue;
     seenUnk.add(key);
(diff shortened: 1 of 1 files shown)


benjie:
An alternative fix would be to output the unmatched to STDERR rather than STDOUT - that way it would still be output for debugging but wouldn't be written to the file.

RinZ27:
@benjie updated to log unmatched contributors to STDERR instead. Email still stays out of the stdout output, but now shows up for debugging when running the script manually.

PR #1227, opened 2026-05-23
Allow empty selection sets

benjie:
Since 2015 we've required selection sets to contain at least one field; however it's always been possible to have an empty object returned:

```graphql
type A { a: Int }
type B { b: Int }
union U = A | B
type Query { u: U }

query {
  u {
    ... on A { __typename }
  }
}
```

Here, if `u` returns type `B` the result will be `{"data":{"u":{}}}` - i.e. an empty object.

GraphQL clients are getting smarter and smarter, and it's common to have client-side extensions such as [Apollo's @client](https://www.apollographql.com/docs/react/local-state/local-resolvers), [Relay Resolvers](https://relay.dev/docs/v15.0.0/guides/relay-resolvers/) or [`@mock`'d fields](https://gaps.graphql.org/GAP-10/draft/). In all of these cases, any selections of these fields need to be removed before the document is sent to the server. But what happens when all the fields in the selection set are to be removed?

```graphql
  query LaunchDetails($launchId: ID!) {
    launch(id: $launchId) {
      isInCart @client
      site @client
    }
  }
```

There are many options (add `__typename`, remove `launch`, throw error, ...), but the safest solution is to allow the selection set to be empty. Hence, this PR.

- Fixes #674

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -747,7 +747,7 @@ fragment scalarSelectionsNotAllowedOnInt on Dog {
 }
 ```
 
-An empty selection is still a selection, so the following is also invalid.
+An empty selection is a selection, so the following is also invalid.
 
 ```graphql counter-example
 fragment scalarEmptySelectionsNotAllowedOnInt on Dog {
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -747,6 +747,14 @@ fragment scalarSelectionsNotAllowedOnInt on Dog {
 }
 ```
 
+An empty selection is still a selection, so the following is also invalid.
+
+```graphql counter-example
+fragment scalarEmptySelectionsNotAllowedOnInt on Dog {
+  barkVolume {}
+}
+```
+
 Conversely, non-leaf fields must have a field subselection. A non-leaf field is
 any field with an object, interface, or union unwrapped type.
 
@@ -785,6 +793,14 @@ query directQueryOnObjectWithSubFields {
   human {
     name
   }
+}
+```
+
+The subselection is permitted to be empty, so the following is also valid.
+
+```graphql example
+query directQueryOnObjectWithEmptySelection {
+  human {}
 }
 ```
 
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -720,9 +720,9 @@ fragment conflictingDifferingResponses on Pet {
 - For each {selection} in the document:
   - Let {selectionType} be the unwrapped result type of {selection}.
   - If {selectionType} is a scalar or enum:
-    - The subselection set of that selection must be empty.
+    - The subselection set of that selection must not be present.
   - If {selectionType} is an interface, union, or object:
-    - The subselection set of that selection must not be empty.
+    - The subselection set of that selection must be present.
 
 **Explanatory Text**
 
--- a/spec/Appendix C -- Grammar Summary.md
+++ b/spec/Appendix C -- Grammar Summary.md
@@ -156,7 +156,7 @@ OperationDefinition :
 
 OperationType : one of `query` `mutation` `subscription`
 
-SelectionSet : { Selection+ }
+SelectionSet : { Selection\* }
 
 Selection :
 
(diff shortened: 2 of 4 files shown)


spec/Appendix B -- Notation Conventions.md
```diff
@@ -138,6 +138,25 @@ Page_list :
 - Page_list Page
 - Page
 
+A subscript suffix "{Symbol\*}" is shorthand for a list of zero or more of that
```
martinbonnin: For the record, this is how it renders: 

Image

martinbonnin (approved):
👍

spec/Appendix B -- Notation Conventions.md
benjie: Ah, we probably don’t need this then. I was just reading the markdown and noticed \* wasn’t defined - looks like it’s a sneaky spec-md feature. Thanks Martin! (Did this in the airport lounge, so not my best work ;) )

martinbonnin: Yea, probably we can remove this paragraph altogether. For the reader there's nothing new. In all cases, both work for me. It's great to see this happening!

jerelmiller (approved):
💯

spec/Section 5 -- Validation.md
```diff
@@ -720,9 +720,9 @@ fragment conflictingDifferingResponses on Pet {
 - For each {selection} in the document:
   - Let {selectionType} be the unwrapped result type of {selection}.
   - If {selectionType} is a scalar or enum:
-    - The subselection set of that selection must be empty.
+    - The subselection set of that selection must not be present.
   - If {selectionType} is an interface, union, or object:
-    - The subselection set of that selection must not be empty.
+    - The subselection set of that selection must be present.
```
magicmark, line 725: Empty selection sets are being made valid by omission to the contrary, rather than being explicitly specified. (yes technically the grammar makes this explicit, but could be easy to miss)

"Empty selection sets are a bit weird/nonobvious, but they are valid" - is something along these lines worth calling out in the explanatory text below, or a non-norm note?

magicmark (approved):
nice :)

PascalSenn (approved):

michaelstaib (approved):

mjmahone (approved):

spec/Section 5 -- Validation.md
benjie: @magicmark I've (hopefully) addressed this by adding examples. I considered adding non-normative notes to explain why this may be useful, but decided to avoid it in favour of simplicity.

benjie:
I believe this is ready to merge, pending @yaacovCR's feedback on the GraphQL.js PR:

- https://github.com/graphql/graphql-js/pull/4852

yaacovCR:
If we're going to merge this, I think we can takeoff the experimental label on the associated PR?

martinbonnin:
> If we're going to merge this, I think we can takeoff the experimental label on the associated PR?

Excellent observation @yaacovCR. We can even probably drop the parameter altogether?

martinbonnin:
I made another version of the PR if we decide to expedite this:

* [with experimental flag](https://github.com/graphql/graphql-js/pull/4852)
* [without experimental flag](https://github.com/graphql/graphql-js/pull/4854) 

My preference goes to "without experimental flag". Less complexity is better.

benjie:
I personally would keep the flag until next semver major, even if you default to enabling the feature. Reason: their security systems might not be set up to handle queries like this; consider:

```graphql
{
  user {
    a: friends {}
    b: friends {}
    c: friends {}
  }
}
```

if they're using complexity cost that multiples expected list length by selection set cost, and selection set cost is zero because there's no fields (previously not possible), it might have an impact on their security. (The IBM cost spec does not suffer from this AFAIK, but there are lots of models of query cost in the wild.)

There may be other such concerns that we're not yet aware of, so giving people a way to opt out until v18 does not seem unwise.

martinbonnin:
> giving people a way to opt out until v18 does not seem unwise.

Works for me. All in all:

* drop `experimental`: `experimentalEmptySelectionsSets` -> `allowEmptySelectionSets`
* default to true
* make it deprecated
?

I have updated the PR in that direction. Let me know if something else is preferred.

BoD (approved):

benjie:
We'll merge this during the secondary WG on Thursday 17th unless anyone has reservations (express them now!)

leebyron (approved):

robrichard (approved):

PR #1228, opened 2026-05-23
Allow empty object types

### LINKED ISSUE #568: The ability to represent empty objects

benjie:
**Merge first:**

- #1227 

---

Since 2015 we've always required an object type to define at least 1 field. As we continue to explore GraphQL's position in the new AI-world, we realise that we need schema subsets that represents part of a larger interface, in these cases it's sometimes useful to leave in type references without including full definitions of those types:

```graphql
type Query {
  topAuthors: [Author!]
}
type Author {
  name: String!
  salesVolume: Int
  averageRating: Float
  topBooks: [Book!]
  mediaCoverage: [Article!]
  recentPublishers: [Publisher!]
}
type Book {}
type Article {}
type Publisher {}
```

Here the AI can see a basic schema it can write queries against, but it can also ask to expand the parts of the schema that are relevant to its interests - e.g. books or articles or publishers - without wasting context on things it doesn't care about.

We can make a schema like this valid by including at least one field on each of these types, but it's hard for a tool to know which field(s) to include, and not including fields makes it much more obvious to the LLM that the schema is partial. Making this partial schema itself valid would be helpful for tooling.

AI aside:

- This is also very useful for mutation payloads where no obvious value exists to return yet but we may wish to expand in future. Currently `type MyMutationPayload { success: Boolean! }` is used as a workaround, but it's messy.
- This is also very useful for schema evolution - it allows you to create a type before you know what it will contain, and allows you to deprecate all fields of a type without the schema being invalid.
- This also enables mutation-only schemas whilst still supporting introspection #490 

Related:

- Fixes https://github.com/graphql/graphql-spec/issues/568

### DIFF
--- a/spec/Appendix C -- Grammar Summary.md
+++ b/spec/Appendix C -- Grammar Summary.md
@@ -156,7 +156,7 @@ OperationDefinition :
 
 OperationType : one of `query` `mutation` `subscription`
 
-SelectionSet : { Selection+ }
+SelectionSet : { Selection\* }
 
 Selection :
 
--- a/spec/Section 2 -- Language.md
+++ b/spec/Section 2 -- Language.md
@@ -383,7 +383,7 @@ Note: many examples below will use the query shorthand syntax.
 
 ## Selection Sets
 
-SelectionSet : { Selection+ }
+SelectionSet : { Selection\* }
 
 Selection :
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -716,7 +716,7 @@ Where `name` is a field that will yield a {String} value, and `age` is a field
 that will yield an {Int} value, and `picture` is a field that will yield a `Url`
 value.
 
-A query of an object value must select at least one field. This selection of
+A query of an object value must be made via a selection set. This selection of
 fields will yield an ordered map containing exactly the subset of the object
 queried, which should be represented in the order in which they were queried.
 Only fields that are declared on the object type may validly be queried on that
@@ -1933,9 +1933,8 @@ to denote a field that uses a Non-Null type like this: `name: String!`.
 **Nullable vs. Optional**
 
 Fields are _always_ optional within the context of a _selection set_, a field
-may be omitted and the selection set is still valid (so long as the selection
-set does not become empty). However fields that return Non-Null types will never
-return the value {null} if queried.
+may be omitted and the selection set is still valid. However fields that return
+Non-Null types will never return the value {null} if queried.
 
 Inputs (such as field arguments), are always optional by default. However a
 non-null input type is required. In addition to not accepting the value {null},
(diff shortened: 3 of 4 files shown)


martinbonnin (approved):
+1000

jerelmiller (approved):
💯

leebyron:
## Notes from meeting:

We should support syntax grammar for `{}` to clearly define no fields

key motivation is that "a schema minus all deprecated elements is always valid" which today could be not true! For a bunch of reasons but in this case:

```
type Example {
  onlyField @deprecated
}
```

this cannot create a valid non-deprecated schema! that's our aim.

Another example: a mutation payload (relay style) that wants to be defined up front as 

Specifically important because the LLM-focused subschema generation still has some issues we're working out, and it's unclear that providing `type Book {}` to the LLM is what we want to do or not.

martinbonnin:
Looks like we'll need the same for input objects?

andimarek:
I am not sure this is the right place to comment, but together with the issue about interfaces without fields and interfaces without objects implementing them I am wondering if we should distinguish between valid schemas vs valid and executable schemas.

For example an interface without an implementation is not something I find sensible for a real schema that should be executed but for schema during design it is ok.

martinbonnin:
> an interface without an implementation is not something I find sensible for a real schema that should be executed

@andimarek My preference here would be a lint warning. The spec is already a bit vague on SDL validation (see https://github.com/graphql/graphql-spec/issues/1233#issuecomment-4927061038) so we're not really equipped to have executable vs non-executable validation yet.

PR #1229, opened 2026-05-23
Allow interfaces with no fields, unions with no members

benjie:
Follow up to:

- #1228 

This is again to help subschemas remain valid. Empty unions is fine (we already allow interfaces with no implementations, which is kind of the mirror), but empty interfaces makes it so people may want to use interfaces where unions should currently be used... Relevant:

- #236 

Incidentally I'm okay with deprecating unions... And adding `@oneOf` output types instead.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1449,8 +1449,7 @@ Unions are never valid inputs.
 
 Union types have the potential to be invalid if incorrectly defined.
 
-1. A Union type must include one or more unique member types.
-2. The member types of a Union type must all be Object base types; Scalar,
+1. The member types of a Union type must all be Object base types; Scalar,
    Interface and Union types must not be member types of a Union. Similarly,
    wrapping types must not be member types of a Union.
 
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -1275,8 +1275,7 @@ Interfaces are never valid inputs.
 
 Interface types have the potential to be invalid if incorrectly defined.
 
-1. An Interface type must define one or more fields.
-2. For each field of an Interface type:
+1. For each field of an Interface type:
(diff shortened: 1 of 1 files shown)


benjie:
Pulled from [this great comment](https://github.com/graphql/graphql-spec/issues/236#issuecomment-257747315) (emphasis mine):

> **A Union is a way to say "I may return one of a few things at this location"** - it should be defined locally to the field. For example, a `Video` type doesn't need to describe that it might appear alongside `Photo` for some specific fields, though a specific fields needs to describe that it might return `Video | Photo`.
> 
> **An Interface is a way to say "I conform to this contract"** - it should be defined locally to the thing which conforms, the Object type. For example, a `User` type should be able to say that it fulfills the contract of `Profile` such that any type current known or unknown can be used knowing that it fulfills the contract.
> 
> This also drives the requirement that **Interfaces must define at least one field - otherwise the Interface would not actually describe a contract to be fulfilled**. This concept of contract-less interfaces is often called "marker interfaces" and in my opinion (hopefully not an uncommon opinion) are a bad practice borrowed from limited object-oriented languages.
> -- @leebyron 

---

Whilst I agree with the above, I also think that the GraphQL spec _generally_ avoids dictating best practices and instead focuses on compatibility. I think "don't use marker interfaces" is a best practice that people should generally adhere to, and I think allowing empty interfaces does cause confusion of interface vs union as raised by Caleb, but I think this problem lives in the space of "best practices" rather than "hard requirements" and there are benefits to allowing for empty interfaces - both in terms of future expansion and prototyping (these two things are both "Searchable", but I don't know what that means yet... but when I do it will be an interface contract rather than a union) and in terms of subschema extractions (both of these implement the Searchable interface... [not included]).

I think it warrants debate.

martinbonnin (approved):
+1000
For the record, I'm also very OK with marker interfaces. Java/Kotlin have had them pretty much forever

PR #1231, opened 2026-06-10
A schema with all deprecated elements removed must itself be valid

benjie:
As a precursor to deprecating other schema elements (object types, etc), I figured it would be best to simplify the rule: the schema with all deprecated elements removed should itself be valid. This is a breaking change for servers currently (could mark a previously valid schema as invalid), but solvable in a non-breaking way by removing `@deprecated` directives, or (once they are supported) adding `@deprecated` to object types and other positions. It's non-breaking for clients.

This also aligns with #1230 wherein the schema with all deprecated elements removed is represented via `{ __schema(includeDeprecated: false) { ... } }`.

### DIFF
--- a/spec/Section 3 -- Type System.md
+++ b/spec/Section 3 -- Type System.md
@@ -136,6 +136,9 @@ All types and directives defined within a schema must not have a name which
 begins with {"\_\_"} (two underscores), as this is used exclusively by GraphQL's
 introspection system.
 
+If a GraphQL schema contains any deprecated _schema element_, a copy of the
+schema that omits all deprecated _schema element_ must itself be valid.
+
 ### Root Operation Types
 
 :: A schema defines the initial _root operation type_ for each kind of operation
@@ -968,7 +971,6 @@ IsValidImplementation(type, implementedType):
        2. Let {implementedFieldType} be the return type of {implementedField}.
        3. {IsValidImplementationFieldType(fieldType, implementedFieldType)} must
           be {true}.
-    6. If {field} is deprecated then {implementedField} must also be deprecated.
 
 IsValidImplementationFieldType(fieldType, implementedFieldType):
 


benjie:
cc @fotoetienne

fotoetienne:
Should this go along with a change to allow empty objects?

benjie:
@fotoetienne such as:

- https://github.com/graphql/graphql-spec/pull/1228

fotoetienne:
Yes! Should #1228 be a prereq for [#1231](https://github.com/graphql/graphql-spec/pull/1231)?

benjie:
I don't think it needs to be, but in an ideal world they'd all be merged together.

spec/Section 3 -- Type System.md
```diff
@@ -968,7 +971,6 @@ IsValidImplementation(type, implementedType):
        2. Let {implementedFieldType} be the return type of {implementedField}.
        3. {IsValidImplementationFieldType(fieldType, implementedFieldType)} must
           be {true}.
-    6. If {field} is deprecated then {implementedField} must also be deprecated.
```
leebyron, line 971: imho we should leave this in place though as an implementation of the new definition you've added above?

I don't think the spec should leave various ways in which the subset-is-valid needs to be looked out for

separately, I think removing this actually says something else-- that an interface field being deprecated may differ from the deprecation status of an implemented field? -- that could be a separate proposal if that's what we want to explore

spec/Section 3 -- Type System.md
```diff
@@ -136,6 +136,9 @@ All types and directives defined within a schema must not have a name which
 begins with {"\_\_"} (two underscores), as this is used exclusively by GraphQL's
 introspection system.
 
+If a GraphQL schema contains any deprecated _schema element_, a copy of the
+schema that omits all deprecated _schema element_ must itself be valid.
```
mjmahone, line 140: Maybe: "A copy of a GraphQL schema that omits every deprecated _schema element_ must itself be valid."

A schema with no deprecated elements will be trivially valid. Also making sure `_schema element_` is pluralized correctly.

spec/Section 3 -- Type System.md
benjie: >  we should leave this in place though as an implementation of the new definition you've added

I'm not convinced the redundancy gives us enough value to justify it. One of the reasons that deprecating things (object types, etc) is so hard to do spec-wise is because of all the knock-on consequences (especially with the "at least one field" rules we have, until #1228/etc are merged).

Removing all of that and replacing with "schema with all deprecated elements removed must be valid" is much much simpler; and it's really easy to build into a GraphQL implementation without having to implement each and every one of the rules that govern it individually:

```ts
function assertSchemaIsValid(schema) {
  // ... all other rules ...
  
  const reducedSchema = removeDeprecatedElements(schema)
  if (reducedSchema !== schema) assertSchemaIsValid(reducedSchema);
}
```

Libraries might need to implement more logic to make sure that violations raise useful error messages, but since the error messages aren't specified currently I think this is a fair trade-off to make?

> separately, I think removing this actually says something else

I had to read it carefully, but I don't think so - `implementedType` is the interface:

```graphql
interface ImplementedType {
  id: ID!
  f: Int # implementedField
}
type Query implements ImplementedType {
  id: ID!
  f: Int # field
}
```

This rule says: if field (`Query.f`) is deprecated, then implementedField (`ImplementedType.f`) must also be deprecated; i.e. the following is invalid:

```graphql
interface ImplementedType {
  id: ID!
  f: Int
}
type Query implements ImplementedType {
  id: ID!
  f: Int @deprecated(reason: "...")
}
```

If we remove the deprecated elements (`Query.f`) then the resulting schema is invalid, matching the overarching rule - i.e. we're saying the same thing: schema with deprecated elements removed must be valid.

Removing this rule would remove this potential confusion.

spec/Section 3 -- Type System.md
benjie: I added the more verbose wording to avoid infinite loops such as `const isValid = schema => otherValidationRules(schema) && isValid(removeDeprecated(schema))`. We can't (currently) pluralize keywords because it will break spec-md lookup.

PR #1232, opened 2026-06-15
Fixes for incremenal delivery integration branch

robrichard:
In a previous discussion we had decided to validation agains any overlapping stream fields. This has already been implemented in graphql-js for some time
See https://github.com/graphql/defer-stream-wg/discussions/100 and https://github.com/graphql/graphql-js/pull/4342

Additionally there were two instances of _completed result_ that was not missed and not renamed to _incremental
completion notice_

### DIFF
--- a/spec/Section 5 -- Validation.md
+++ b/spec/Section 5 -- Validation.md
@@ -560,7 +560,7 @@ FieldsInSetCanMerge(set):
   {set} including visiting fragments and inline fragments.
 - Given each pair of distinct members {fieldA} and {fieldB} in {fieldsForName}:
   - {SameResponseShape(fieldA, fieldB)} must be true.
-  - {SameStreamDirective(fieldA, fieldB)} must be true.
+  - {HasNoOverlappingStreams(fieldA, fieldB)} must be true.
   - If the parent types of {fieldA} and {fieldB} are equal or if either is not
     an Object Type:
     - {fieldA} and {fieldB} must have identical field names.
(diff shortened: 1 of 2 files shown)


yaacovCR (approved):

spec/Section 5 -- Validation.md
```diff
@@ -596,14 +596,10 @@ SameResponseShape(fieldA, fieldB):
...
 
-SameStreamDirective(fieldA, fieldB):
+HasNoOverlappingStreams(fieldA, fieldB):
 
 - If neither {fieldA} nor {fieldB} has a directive named `stream`.
   - Return {true}.
-- If both {fieldA} and {fieldB} have a directive named `stream`.
-  - Let {streamA} be the directive named `stream` on {fieldA}.
-  - Let {streamB} be the directive named `stream` on {fieldB}.
-  - If {streamA} and {streamB} have identical sets of arguments, return {true}.
```
benjie, line 606: Hmmm... "distinct members" may be doing a lot here. For example:

```graphql
frag A on Foo {
  ...B
  ...B
}
frag B on Foo {
  blah @stream
}
```

might merge since B.blah is the same field twice, assuming that's what we mean by "distinct members", but

```graphql
frag A on Foo {
  ...B
  ...C
}
frag B on Foo {
  blah @stream
}
frag C on Foo {
  blah @stream
}
```

would not, because `B.blah` != `C.blah` even though it seems that they should have the same result as in the previous example.

This is probably okay for `@stream` because it shouldn't really be fragment-merged in the same way regular fields would be (we want `@stream` to be used reasonably scarcely?)... but it warrants acknowledgement. It's a bit of a break from the norm for GraphQL.

benjie (approved):

spec/Section 5 -- Validation.md
robrichard: Yes this is the behavior we want. See the discussion here: https://github.com/graphql/defer-stream-wg/discussions/100

It becomes an issue when the stream, even with the same arguments has different selection sets. One fragment can influence the performance characteristics of the stream on another fragment, which is why we decided to prohibit this.

benjie: Hmmm... but what about:

```graphql
frag A on Foo {
  ...B
  ...C
}
frag B on Foo {
  ...D @include(if: $var1)
}
frag C on Foo {
  ...D @defer(if: $var2)
}
frag D on Foo {
  blah @stream
}
```

Here, is `A>B>D.foo` seen as the same as `A>C>D.foo`? Are they distinct members?
... does it matter? :thinking: I suppose not...

PR #1237, opened 2026-09-10
Install wgutils to aid with spec releases

benjie:
E.g. the following will overwrite the previous changelog with a draft version

```sh
npm exec wgutils -- spec version --previous October2021 --current f29fbcd2ab5af763fce7ad62896eb62465a669b3 September2025 --force
```


benjie:
@graphql/tsc I'll merge this next week unless there are any concerns.

build.sh
```diff
@@ -1,105 +0,0 @@
-#!/bin/bash -e
-# This script publishes the GraphQL specification document to the web.
```
benjie: This script translated to TypeScript and moved to wgutils and then made generic (so we can use it with GraphQL over HTTP/etc) here: https://github.com/graphql/wgutils/blob/9f66966ab9a8c6f237e9ddf25ea95afe987f7f87/src/spec/build/index.ts#L55

PR #1238, opened 2026-09-10
Clarify definition of response position

benjie:
> :: A _response position_ is a uniquely identifiable position in the response
data produced during execution. It is either a direct entry in the {resultMap}
of a {ExecuteSelectionSet()}, or it is a position in a (potentially nested) List
value. Each response position is uniquely identifiable via a _response path_.

The intention of _response position_ is that it includes things that were elided by error propagation; e.g. if you query `{ a { b } }` and `b` is non-null and throws, the result will be `{a: null}` but the error will have path `["a", "b"]` indicating the `a.b` response position even though that position does not actually exist in the response.

(Also: `ExecuteSelectionSet()` doesn't exist any more.)

This PR clarifies the definition of response position to include these omitted values.


martinbonnin (approved):
👍

PascalSenn (approved):

robrichard (approved):

leebyron (approved):

